PR #233, opened 2015-10-15
Correct setup.py

jusabatier:
On Debian Jessie with pip install :

Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 20, in <module>
      File "/tmp/pip-o_ERWm-build/setup.py", line 16, in <module>
        with open('requirements.txt', encoding='utf-8') as reqs:
    TypeError: 'encoding' is an invalid keyword argument for this function

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 
 import codecs
+import io
 
 from setuptools import setup, find_packages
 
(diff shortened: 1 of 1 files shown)


yohanboniface:
Thanks :)
That's more about python2/3 I think. You are in python 2.x I guess?

jusabatier:
Yes : Python 2.7.9

yohanboniface:
OK, so that explain, I'm in 3.4 and forgot to test on p2 :/

Thanks again! :)

PR #242, opened 2015-11-04
Missing package django-appconf added

amenk:


yohanboniface:
Isn't that a requirement of django-compressor, that then should be pulled in when installing it?

amenk:
Possible. I am a very newbie with django and somehow was getting an error. Will try again on a clean box

amenk:
You are right. Managed to install now without specifying that explicitly.

PR #270, opened 2016-01-02
New map settings

plepe:
- set default map location for new maps in local.py
- new setting UMAP_EXCLUDE_DEFAULT_MAPS
- settings to configure count of showcased maps


umap/settings/local.py.sample
```diff
@@ -64,6 +64,16 @@ SOCIAL_AUTH_BACKEND_ERROR_URL = "/"
 # UMAP_SHOWCASE_PK = 1156
 LEAFLET_STORAGE_ALLOW_ANONYMOUS = True
 UMAP_DEMO_SITE = True
+
+# This setting will exclude empty maps (in fact, it will exclude all maps where
+# the default center has not been updated)
+UMAP_EXCLUDE_EMPTY_MAPS = True
+
+# How many maps should be showcased on the main page resp. on the user page
+UMAP_MAP_LIST_PER_PAGE = 5
```
yohanboniface: What about `UMAP_MAPS_PER_PAGE`, more simply?

umap/views.py
```diff
@@ -54,8 +54,8 @@ class Home(TemplateView, PaginatorMixin):
 
     def get_context_data(self, **kwargs):
         qs = Map.public
-        if 'spatialite' not in settings.DATABASES['default']['ENGINE']:
-            # Unsupported query type for sqlite.
+        if getattr(settings, 'UMAP_EXCLUDE_EMPTY_MAPS', False) and 'spatialite' not in settings.DATABASES['default']['ENGINE']:
```
yohanboniface: This does not seem to be 80 cols compliant ;)

umap/settings/local.py.sample
```diff
@@ -64,6 +64,16 @@ SOCIAL_AUTH_BACKEND_ERROR_URL = "/"
 # UMAP_SHOWCASE_PK = 1156
 LEAFLET_STORAGE_ALLOW_ANONYMOUS = True
 UMAP_DEMO_SITE = True
+
+# This setting will exclude empty maps (in fact, it will exclude all maps where
+# the default center has not been updated)
+UMAP_EXCLUDE_EMPTY_MAPS = True
```
yohanboniface: `UMAP_EXCLUDE_DEFAULT_MAPS`?

plepe:
Thanks for your comments. I just updated the branch accordingly.

umap/views.py
```diff
@@ -114,7 +115,7 @@ def get_context_data(self, **kwargs):
         manager = Map.objects if owner else Map.public
         maps = manager.filter(Q(owner=self.object) | Q(editors=self.object))
         maps = maps.distinct().order_by('-modified_at')[:50]
-        per_page = 10 if owner else self.per_page
+        per_page = getattr(settings, 'UMAP_MAPS_PER_PAGE_OWNER', 10) if owner else getattr(settings, 'UMAP_MAPS_PER_PAGE', self.per_page)
```
yohanboniface: This line too needs to be under 80 ;)
Actually, here is something to make the code a bit simpler: also add all the settings that uMap expects in settings/base.py, so we are sure they exist, and so you can do `settings.XXXX` instead of `getattr(settings, 'XXX')`.
Makes sense? :)

plepe:
Now I entered the default values to base.py and replaced getattr() by the shorter version. Hope, everything is fine now ... :-)

yohanboniface:
Thanks :)

plepe:
\o/

PR #271, opened 2016-01-03
Authentication: login via username / password

plepe:
This pull request should enable login via username / password, if configured in the settings to allow this kind of method.

TODO:
- [x] En-/Disable local login in settings
- [x] Link to password change page
- [x] Successful login after login failure leads to strange error message (csrf failure)

### DIFF
--- a/umap/context_processors.py
+++ b/umap/context_processors.py
@@ -11,3 +11,8 @@ def version(request):
     return {
         'UMAP_VERSION': __version__
     }
+
+def authentication(request):
+    return {
+        'ENABLE_ACCOUNT_LOGIN': settings.ENABLE_ACCOUNT_LOGIN
+    }
--- a/umap/settings/base.py
+++ b/umap/settings/base.py
@@ -110,6 +110,7 @@ TEMPLATE_CONTEXT_PROCESSORS += (
     'social.apps.django_app.context_processors.login_redirect',
     'umap.context_processors.feedback_link',
     'umap.context_processors.version',
+    'umap.context_processors.authentication',
 )
 
 TEMPLATE_LOADERS = (
@@ -134,6 +135,7 @@ MIDDLEWARE_CLASSES = (
 # Auth / security
 # =============================================================================
 
+ENABLE_ACCOUNT_LOGIN = True
 AUTHENTICATION_BACKENDS += (
 )
 
--- a/umap/settings/local.py.sample
+++ b/umap/settings/local.py.sample
@@ -37,6 +37,10 @@ COMPRESS_OFFLINE = True
 
 LANGUAGE_CODE = 'en'
 
+# Set to False if login into django account should not be possible. You can
+# administer accounts in the admin interface.
+ENABLE_ACCOUNT_LOGIN = True
+
 AUTHENTICATION_BACKENDS = (
     'social.backends.github.GithubOAuth2',
     'social.backends.bitbucket.BitbucketOAuth',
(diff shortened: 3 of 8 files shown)


yohanboniface:
Travelling now, I'll have a look into two days.

plepe:
Ok, no problem!

yohanboniface:
Still WIP or reviewable now? :)

plepe:
No, see the TODO-list in the topmost-post. I still haven't figured out how to submit a settings-value to the template (to enable/disable the login fields).

yohanboniface:
> I still haven't figured out how to submit a settings-value to the template (to enable/disable the login fields)

Humm, AFAIK we are using a django view, so the less hackish option may be a context processor: https://docs.djangoproject.com/en/1.9/ref/templates/api/#writing-your-own-context-processors

plepe:
Fixed TODO 3 in umap-project/Leaflet.Storage#107.

plepe:
Thanks for the hint about the context processors. That was what I was hoping for. The pull request should be finished now (don't forget the change in Leaflet.Storage).

umap/context_processors.py
```diff
@@ -11,3 +11,8 @@ def version(request):
     return {
         'UMAP_VERSION': __version__
     }
+
+def authentication(request):
+    return {
+        'AUTHENTICATION_ENABLE_ACCOUNT_LOGIN': settings.AUTHENTICATION_ENABLE_ACCOUNT_LOGIN
```
yohanboniface: `ENABLE_ACCOUNT_LOGIN` would be clear enough while shorter. What do you think?

yohanboniface:
I'd love some minimal unittests, I'd say at least one for username/password login and one for change password view. Can you do that?

plepe:
Sure, I'll see what I can come up with ...

plepe:
Hmm, what exactly should the tests test?

Because I just added the following things:
* Added the login form with the enable/disable mechanism
* Added the password change form for local accounts (including the necessary URLs, and a link to the password change if the password is usable)
* Hid the SocialAuth block, when no backends are enabled

I did not add/modify any views, because these is all django internal stuff ...

So, I could add a test whether the login form is rendered if account authentication is enabled?

plepe:
I've now shortened the switch to ENABLE_ACCOUNT_LOGIN.

yohanboniface:
I'd go for some integration tests, at minimum:
- one to test the login with username and password (so get the form, fill in the form, post the form, check user is logged in)
- one for update password (same GET/fill/POST/check pattern)

Even if we are using internal Django mechanisms, it's good to have integration tests covering them because we still can break many things or have conflict with our settings or usage: forgetting some middleware, conflict with some third party app, etc. In short: we want to make sure we will not break this feature in the future :)

yohanboniface:
@plepe I'd love to see this merged. Any time/interest to add some minimal unittests?

plepe:
Yeah, I too ... But I've got no experience with unit tests in django (I have to admit, I haven't done a lot of unit testing at all until now).

The best I could come up with is this:
```python
from django.contrib.auth.models import User

class TestsLocalAccount(TestCase):
    def setUp(self):
        self.user = User.objects.create_user('temporary', 'foo', 'password')

    def test_login_successful(self):
        response = self.client.post("/de/login/", {'username': 'foo', 'password': 'password'})
        self.assertEqual(response.status_code, 200)
        print(response)
```

But response just shows the (html code of the) login form. Any idea how to proceed from here?

PR #274, opened 2016-01-09
Use new django-leaflet-storage:MapManager

plepe:
see umap-project/django-leaflet-storage#107.


yohanboniface:
django-leaflet-storage is now integrated into uMap.

PR #351, opened 2016-08-24
Improved gitignore with gitignore.io

tmerlier:
Ignore unwanted files Mac OS, Linux and Python


yohanboniface:
Huhu, is there something remaining we do not ignore? :p

PR #535, opened 2017-12-09
Mark a string for translation

Ltrlg:
Just created my account on umap.osm.org and I found this untranslated.


yohanboniface:
Thanks :)

PR #544, opened 2018-01-19
Fixes in the documentation for Ubuntu from scratch

Duvel:
I just installed umap following the Ubuntu from scratch manual, but ran in to a few issues. I fixed the issues in the manual, also including some of issues already raised.

### DIFF
--- a/docs/ubuntu.md
+++ b/docs/ubuntu.md
@@ -14,15 +14,14 @@ You need sudo grants on this server, and it must be connected to Internet.
 
 ## Create deployment directories:
 
-    sudo mkdir -p /srv/umap
-    sudo mkdir -p /etc/umap
+   sudo mkdir -p /etc/umap
 
 *You can change this path, but then remember to adapt the other steps accordingly.*
 
 
 ## Create a Unix user
 
-    sudo useradd -N umap -d /srv/umap/
+    sudo useradd -N umap -m -d /srv/umap/
 
 *Here we use the name `umap`, but this name is up to you. Remember to change it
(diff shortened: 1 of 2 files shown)


docs/ubuntu.md
```diff
@@ -74,6 +73,11 @@ you will need to run again this last line.*
     wget https://raw.githubusercontent.com/umap-project/umap/master/umap/settings/local.py.sample -O /etc/umap/umap.conf
 
 
+## Make umap use the local configuration file
+
+   export UMAP_SETTINGS=/srv/umap/local.py
```
yohanboniface: Humm, the tutorial should use the default configuration path, which is `/etc/umap/umap.conf`, and should be loaded by uMap without needing an env var. Have you tested this way?

Duvel: I just tested this and uMap doesn't seem to be using umap.conf. That is probably why after searching I used local.py with the environment settings.

Duvel: Looking at the commits and the releases, the commit https://github.com/umap-project/umap/commit/4c8b6f79de7b49692ed6a2b75c25b7e5ea6d4dfb for umap.conf is after the last release 0.8.2. Maybe that would explain it?

Duvel: Just had a look at settings/__init__.py and that contains the old code. I replaced it with the new version and now it works, so it is a version mismatch.

Duvel:
With newer versions probably no longer necessary.

yohanboniface:
Thanks, I've still cherry-picked two commits! :)

PR #580, opened 2018-05-16
bump social-auth to use https:// for osm.org oauth

datendelphin:
openstreetmap.org switched to https://
login with umap does not work anymore with old
social-auth version. The new version now has the
https:// urls.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,4 @@ Pillow==4.1.0
 psycopg2==2.7.1
 requests==2.13.0
 social-auth-app-django==1.1.0
-social-auth-core==1.2.0
+social-auth-core==1.7.0


tomhughes:
Thanks. I'm not really sure why adding the redirect has caused this as umap seems to be using https anyway without following the redirect, but since the redirect was added it seems to be sending requests without any OAuth signature but @datendelphin reports that the update seems to fix it.

yohanboniface:
I've made a workaround "the papa way" until a cleaner fix tonight or tomorrow as said.

yohanboniface:
Released as 0.8.3 https://pypi.org/project/umap-project/0.8.3/#description

Next step is 1.0.0, with a full upgrade of python and js deps, a merge of django-leaflet-storage into umap (to ease development for new comers), some fixes in the air, not more.
Full upgrade means Django 2.x, so Python 3.x.

PR #599, opened 2018-07-07
Proposal of migrating step by step to ES6 (or newer)

vinyll:


davidbgk:
Thanks mate, we'll handle it through #1081 where I discovered the existence of [lebab](https://github.com/lebab/lebab) 🙃

PR #652, opened 2018-11-15
Updated ubuntu installation doc

aaj013:
Added missing dependency packages
Added missing ownership to deployment directories


yohanboniface:
Thanks!
You can also take inspiration from https://github.com/umap-project/umap-deploy/blob/master/remote/__main__.py#L22 which is what I use now and is then up to date for latest ubuntu (but don't use the scripts themselves, they are custom made).

PR #669, opened 2019-02-13
Change 'cured' to 'improved'

brylie:
From the context, I think it was implied that OpenStreetMap is *improved* by community adoption, rather than being 'cured' of some 'sickness' :-)


yohanboniface:
Thanks :)

PR #674, opened 2019-02-25
Fix original English translation strings

ward:
It seems mistakes in the original English (en) translation strings are
not fixed in the website despite them being correct for en_US in
Transifex for several years. My guess is that they are the default for
en, so unless you somehow manage to explicitly select en_US (which does
not seem possible), you are stuck with the incorrect English.


yohanboniface:
Thanks!

PR #683, opened 2019-04-09
Allow to star maps and retrieve starred maps

yohanboniface:
Not sure we really need this actually, but here is a POC.

Screenshot-20190409094456-2524x670
Screenshot-20190409094419-644x498


umap/static/umap/img/24-white.png
davidbgk, line 1: Missing white star?

umap/static/umap/img/24-white.svg
davidbgk, line 1: Missing stars?

umap/static/umap/js/umap.controls.js
```diff
@@ -966,6 +966,30 @@ L.U.AttributionControl = L.Control.Attribution.extend({
...
+            link = L.DomUtil.create('a', '', container);
+        link.href = '#';
+        link.title = L._('Star this map');
+        var fn = function () {
+            map.star()
+        };
+
+        L.DomEvent
+            .on(link, 'click', L.DomEvent.stop)
+            .on(link, 'click', fn, map)
```
davidbgk, line 987: I'm not sure you need to define a `fn` here, `.on(link, 'click', map.star, map)` should work?

umap/views.py
```diff
@@ -182,6 +182,30 @@ def get_template_names(self):
...
+        manager = Map.objects if owner else Map.public
+        stars = Star.objects.filter(by=self.object).values("map")
+        maps = manager.filter(pk__in=stars)
+        if owner:
+            per_page = settings.UMAP_MAPS_PER_PAGE_OWNER
+            limit = 100
+        else:
+            per_page = settings.UMAP_MAPS_PER_PAGE
+            limit = 50
+        maps = maps.order_by('-modified_at')[:limit]
```
davidbgk, line 199: Is the `limit` pertinent if you paginate the view anyway?

umap/views.py
```diff
@@ -634,6 +668,20 @@ def post(self, *args, **kwargs):
...
+class MapStar(View):
+
+    def post(self, *args, **kwargs):
+        map_inst = get_object_or_404(Map, pk=kwargs['map_id'])
+        qs = Star.objects.filter(map=map_inst, by=self.request.user)
+        if qs.exists():
+            qs.delete()
+            status = False
+        else:
+            Star(map=map_inst, by=self.request.user).save()
```
davidbgk, line 680:
```suggestion
            Star.objects.create(map=map_inst, by=self.request.user)
```

I think this is a more djangonic(?) way (optional).

umap/views.py
```diff
@@ -634,6 +668,20 @@ def post(self, *args, **kwargs):
         return response
 
 
+class MapStar(View):
```
davidbgk, line 671:
```suggestion
class ToggleMapStarStatus(View):
```

For a more explicit name (URLs need to be updated accordingly).

umap/views.py
yohanboniface: I'd so, otherwise we'd paginate indefinitely, but I may be missing something.

umap/views.py
yohanboniface: Done!

umap/static/umap/js/umap.controls.js
yohanboniface: Done! :)

umap/static/umap/img/24-white.png
yohanboniface: I'd say we don't need those, because there are no actions nor button in the "edit area" (which is white on black), but I can use those just in case if you think that's more consistent.

davidbgk:
See also https://github.com/umap-project/umap/pull/1084

davidbgk (approved):

PR #715, opened 2019-10-08
Update ubuntu from scratch doc

gislars:
I had to add a config for /static and /uploads folder in nginx when I set DEBUG = False


yohanboniface:
Good catch, thanks!

Issue #491, opened 2017-05-13
Changing umap account name 


aftocext:
Hello
 I associated my Umap account with an OpenStreetMap account thru OAuth, but afterwards I changed the displayname of my account in OpenStreetMap and I really wish that this change is reflected into my Umap account (which is not the case, i still see my previous name on Umap). Is there a way this can be changed at least manually  by somebody, as I didn't see any thing related to account management on the Umap side ? Thank you very much if this can be done !  (I ll provide account details when necessary).
Cheers !

pdfext:
No answer on this one ?

grandbic:
Bonjour,
If i've understood:
open account 1
permissions/editers/name of 2nd account
close 1st account /open 2nd account
edit/advanced actions /clone

tomhughes:
Can we please, please fix this! I'm fed up with people closing their OSM accounts and creating new ones just so they can change their name on uMap!

Well that and complaining that they can't contact anybody at uMap and hence come to OSM support instead with their uMap issues.

vinber:
I think it's a duplicate : https://github.com/umap-project/umap/issues/7

Binnette:
Fixed by #754

Issue #302, opened 2016-04-03
when you change your name (aka) in OSM settings, there's no change in umap


vinber:
Example for the same account :
- umap : vinber-inli
- OSM : vinber-num&lib

yohanboniface:
Yes, we need a way for a user to resync with an auth provider. See also #7

vinber:
Thanks, I haven't seen this issue (#7) and I didn't know it was "same" kind of issue :(

datendelphin:
Ah I see this is already reported. Just got a similar question from a user.
For the changing osm user names, I think from OSM the user Id should be saved, because that one is stable.

siSSSou:
This feature (sync uMap account name with OSM account name) would be valuable.
Our corporate structure has changed its name (a ministry) and our maps can't be updated !

Binnette:
Fixed by #754

PR #754, opened 2020-02-29
Update username with value from OAuth provider

Binnette:
Here is a fix for the issue #302 (changing username in OSM do not sync in uMap).

I tracked this problem deep down the following issues:

-  python-social-auth/social-core#263
-  python-social-auth/social-core#348

This issue is related to the fact that in "Python Social Auth - Core", the field "username" is protected and can not be updated. So if you change your username in OSM, GitHub, BitBucket or Twitter, it will never be synced in uMap. :disappointed:

The best solution is a fix in "Python Social Auth - Core", we should allow developers to ignore those "default protected fields", i will probably make a PR to social-core.

So i made a workaround in uMap to fix this issue and it is working nicely. To do so, iI defined my own [pipeline](https://python-social-auth.readthedocs.io/en/latest/pipeline.html) and used it instead of the default pipeline.

This PR will fix following issues: #491, #302

### DIFF
--- a/umap/utils.py
+++ b/umap/utils.py
@@ -110,8 +110,12 @@ def gzip_file(from_path, to_path):
         with gzip.open(to_path, 'wb') as f_out:
             f_out.writelines(f_in)
 
-def user_details(strategy, details, user=None, *args, **kwargs):
-    """Update user details using data from provider."""
+def oauth_user_details(strategy, details, user=None, *args, **kwargs):
+    """
+    This method is a pipeline stage only to be used with social_auth
+    Note this is a workaround that can break if social_auth is updated
+    Update user details using data from provider.
+    """
     if not user:
         return
 
--- a/umap/utils.py
+++ b/umap/utils.py
@@ -109,3 +109,33 @@ def gzip_file(from_path, to_path):
     with open(from_path, 'rb') as f_in:
         with gzip.open(to_path, 'wb') as f_out:
             f_out.writelines(f_in)
+
+def user_details(strategy, details, user=None, *args, **kwargs):
+    """Update user details using data from provider."""
+    if not user:
+        return
+
+    changed = False  # flag to track changes
+    # uMap fix: remove 'username' from protected fields
+    protected = ('id', 'pk', 'email') + \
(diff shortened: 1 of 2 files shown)


umap/utils.py
```diff
@@ -109,3 +109,33 @@ def gzip_file(from_path, to_path):
     with open(from_path, 'rb') as f_in:
         with gzip.open(to_path, 'wb') as f_out:
             f_out.writelines(f_in)
+
+def user_details(strategy, details, user=None, *args, **kwargs):
```
yohanboniface: I'd name it something like `oauth_user_details`, given it's in a generic `utils` file, thus without namespace to explicit what `user_details` are related to.
Also, I'd add a line in the docstring to define it's a pipeline stage to be used with social_auth.

umap/utils.py
```diff
@@ -109,3 +109,33 @@ def gzip_file(from_path, to_path):
...
+    """Update user details using data from provider."""
+    if not user:
+        return
+
+    changed = False  # flag to track changes
+    # uMap fix: remove 'username' from protected fields
+    protected = ('id', 'pk', 'email') + \
```
yohanboniface: Maybe those protected fields could be another setting upstream ?

Binnette: Yes, I want to make a PR to social_auth that way :+1:

Binnette:
Hi @yohanboniface, I totally agree you. Before merging this PR, I will PR social_auth and see what happens.

It seems that social_auth [releases](https://github.com/python-social-auth/social-core/releases) are less frequents lately. So my workaround can be valuable. :horse:

Binnette:
So here is my PR python-social-auth/social-core#444 to social core. If i want to remove my workaround, I will need python-social-auth/social-core#348 to be merged too.

yohanboniface:
@Binnette seems that PRs have been merged upstream. Want to update on our side before the release ?

Binnette:
I don't have time to do it now. I'm busy on an other project :sweat_smile:  https://github.com/opencovid19-fr/data/

Issue #444, opened 2017-02-07
sandstorm.io


QJKX:
I'd love to see umap as a @sandstorm-io app.

yohanboniface:
See also #50

PR #760, opened 2020-03-08
Add 40 icons with basic shapes

Binnette:
I drew 20 icons in SVG and exported them in black and white.
This PR fix issue #110

Here are my icons:
all-shapes


yohanboniface:
Usually, uMap does not come with it's own icons, but why not providing some default.

In that case, I suggest we follow the design rules used on umap.osm.fr, see "osmic.yaml" and "osmic-white.yaml" to have details, but basically:

- icons are provided both in grey and white
- grey ones are in #444 (not black) with a whiter halo
- white ones are in #ededed with a #444 halo

Other option is to provide those basic shapes to osmic project, so we can continue to consume that source.

Binnette:
In fact, I wanted to fix #110, but may be we can just close #110 instead? (and close my PR BTW)

yohanboniface:
Documentation actually exists https://umap-project.readthedocs.io/en/master/administration/#icons

PR #764, opened 2020-03-17
Set icon img max-height

### LINKED ISSUE #143: Portrait images used for marker picto are overlapping
Originally reported by: **BitBucket: [orhygine](http://bitbucket.org/orhygine), GitHub: [orhygine](http://github.com/orhygine)**

---

It is possible to use "{url}" for the source of marker picto which is a great feature.

If the image used is portrait oriented, it is not properly clipped to the marker shape.

Here is an example: http://umap.openstreetmap.fr/fr/map/photos-baigura_23901#17/43.28596/-1.29596

---
- Bitbucket: https://bitbucket.org/yohanboniface/umap/issue/143

yohanboniface:
Not sure why I didn't fix that before, sounds like an easy fix. So maybe there is a trap I forgotten since then…

Before:
Screenshot-20200317192120-224x250

After:
Screenshot-20200317192055-250x250

fix #143

### DIFF
--- a/umap/static/umap/map.css
+++ b/umap/static/umap/map.css
@@ -1000,6 +1000,7 @@ a.add-datalayer:hover,
 .umap-div-icon .icon_container img {
     vertical-align: middle;
     max-width: 24px!important; /* leaflet.css has !important, so... */
+    max-height: 24px!important;
 }
 .umap-div-icon .icon_arrow {
     border-left: 8px solid transparent;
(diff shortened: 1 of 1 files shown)


Binnette:
I will probably to this the exact same way, except i'm a bit "allergic" to `!important`. But i saw that this is a leaflet "best practice"...

yohanboniface:
> But i saw that this is a leaflet "best practice"...

Not a best practice even for Leaflet ;)
But in this case we are overriding a property set on Leaflet with `!important` on it, can't remember why this has been added sadly.

yohanboniface:
Ah, now I remember:

https://github.com/Leaflet/Leaflet/commit/61d746818b99d362108545c151a27f09d60960ee

Comes from Bootstrap doing shitty things that Leaflet needs to override and thus we…

PR #766, opened 2020-03-18
Add a blinking marker on user position on geolocate

### LINKED ISSUE #339: feature request - blue spot for current gps position
dear staff,
I see that the function "center in the map" is not showing the blue spot with current gps position like it works for Google maps. Can you make it? I think that it is a important function for every maps.

thank you for the support!

yohanboniface:
fix #339

It's a naive approach:

Screenshot-20200318104817-1572x1376

@Binnette wanna play a bit with it ?


Binnette:
It's working great! I was wondering that my position will never been remove as in any other maps.

Try it on https://www.openstreetmap.org

My position is remove only if i click again on the button. I think this is the standard behaviour. So i think we must implement it like that :wink:

yohanboniface:
> We should implement the default behavior wink

:+1: 

Will do next week-end :)

yohanboniface:
@Binnette pushed a new version, with toggle:

Screenshot-20200321181736-2164x1014

Binnette:
Great :+1: You're gonna say i am annoying. I really hate things that blink on my screen :sweat_smile: 

image

Can you please stop the blink after 2/3 blinks please :sweat_smile: 

https://www.theworldsworstwebsiteever.com/

yohanboniface:
Hehe :)

I'd say that the comparison is a bit excessive, but I take it ;)

We could remove the blinking totally also, the thing is that now that we're watching the position, the blinking serve as a reminder that there is something "alive".

So, I'd not remove the blinking after 2/3 blinks (it would make the code and the UX more complex), I'd either keep it (if we think it has a meaning) or remove it totally.

Thoughts ?

Binnette:
Let's remove it then. Other maps don't make this marker blink :wink:

yohanboniface:
Done! :)

umap/static/umap/js/umap.controls.js
```diff
@@ -826,21 +826,62 @@ L.U.LocateControl = L.Control.extend({
...
+    onFound: function (e) {
+        this._map._geolocated_circle.setRadius(e.accuracy);
+        this._map._geolocated_circle.setLatLng(e.latlng);
+        this._map._geolocated_marker.setLatLng(e.latlng);
+        this._map.addLayer(this._map._geolocated_circle);
+        this._map.addLayer(this._map._geolocated_marker);
+    },
+
+    onError: function (e) {
+        this.ui.alert({content: L._('Unable to locate you.'), 'level': 'error'});
```
Binnette: New translations :)

yohanboniface: Good catch :)

brunetton:
Thanks to both of you; I was waiting for this. It's party time :tada:

OttoRiehl:
My request 
[](https://forum.openstreetmap.org/viewtopic.php?id=66170)

Is there a parameter in the URL for the browser to draw a marker at a central position?

sojwolf:
How do I add this feature, the blue dot user position, to my uMap? Is it a built-in function?

OttoRiehl:
sojwolf - blue dot 

- Karteneigenschaften bearbeiten
- Einstellung Benutzeroberfläche
- Den Ortungsbutton anzeigen.

sojwolf:
Thanks! However, it seems not to work on my iPhone IOS 14.3.. (Location set to allowed). Any ideas?

yohanboniface:
Make sure you are using https

OttoRiehl:
Hello developers of uMap:

I want to mark a specific position on the map when opening a map.
I want to mount in the link the position as lati and longi.
[Example 1](https://umap.openstreetmap.fr/de/map/germans-from-russia-and-eastern-europe-settlement-_320599#12/46.1578/29.4274)

But how do I cause a position indicator (for example pin or crosshairs) to be set at coordinates 46.1578/29.4274 in the browser call?

Here is an example in another project:
[Example 2](https://www.historic.place/themes/GRSL/index.html?zoom=11&lat=46.1578&lon=29.4274&pid=KdSa&pins=29.4274!46.1578)

I want to show: Where is the position around the existing nodes?

This is only temporary for the call.

PR #770, opened 2020-03-22
Feature permalink

### LINKED ISSUE #294: Unique marker url
Hi
Is there a way to get unique POI (marker, trace) URL ?

I had a look in slideshow mode but did not see unique url, but only datalayer url like "/en/datalayer/{pk}/"
It should be nice to have direct link on one POI (to retrieve one POI data).

Cheers

### LINKED ISSUE #497: UMAP - External call (iframe) &datalayers parameters: is it possible to specify a marker name/id to focus the map on it
Hello
I'm calling the iframe text generated by umap to call the map from a website: that's super cool.
The map has for instance 50 markers.
But i'd like for my 50 websites pages to call for the map but the map must be "centered" to a specific marker (the one we talk about on the specific page).
I see there is &datalayers parameters, that if we choose option "show as current view", the outcome is like this: &datalayers=370676#7/35.747/35.145
It's cool but i'd like to easy  my job and i don't want to go to each marker to get their coordinates so to call the map focused on the marker from outside.
Is there a way, via parameter to do so?
Of course you might tell me to import map data to my website and build whatsoever php, but please try answer my question. Any suggestion is welcomed :)

yohanboniface:
That may need a bit more thinking, but that may also be a good starting point:

- every feature has a unique identifier, based on one of its properties (by default: name)
- this property name can be changed in the map settings
- when exporting to iframe, there is now an option to include the current feature
- on right click on a feature, there is a "permalink" action

Screenshot-20200322103920-754x1132

fix #497
fix #294

### DIFF
--- a/umap/static/umap/js/umap.core.js
+++ b/umap/static/umap/js/umap.core.js
@@ -175,6 +175,19 @@ L.Util.flattenCoordinates = function (coords) {
     return coords;
 };
 
+
+L.Util.buildQueryString = function (params) {
+    var query_string = [];
+    for (var key in params) {
+        query_string.push(encodeURIComponent(key) + '=' + encodeURIComponent(params[key]));
+    }
+    return query_string.join('&');
+};
+
+L.Util.getBaseUrl = function () {
+    return '//' + window.location.host + window.location.pathname;
+};
+
 L.DomUtil.add = function (tagName, className, container, content) {
     var el = L.DomUtil.create(tagName, className, container);
     if (content) {
@@ -428,6 +441,7 @@ L.U.Help = L.Class.extend({
     shortCredit: L._('Will be displayed in the bottom right corner of the map'),
     longCredit: L._('Will be visible in the caption of the map'),
     sortKey: L._('Property to use for sorting features'),
+    slugKey: L._('The name of the property to use as feature unique identifier.'),
     filterKey: L._('Comma separated list of properties to use when filtering features'),
     interactive: L._('If false, the polygon will act as a part of the underlying map.'),
     outlink: L._('Define link to open in a new window on polygon click.'),
--- a/umap/static/umap/js/umap.features.js
+++ b/umap/static/umap/js/umap.features.js
@@ -45,6 +45,15 @@ L.U.FeatureMixin = {
         return this.datalayer && this.datalayer.isRemoteLayer();
     },
 
+    getSlug: function () {
+        return this.properties[this.map.options.slugKey || 'name'] || '';
+    },
+
+    getPermalink: function () {
+        const slug = this.getSlug();
+        if (slug) return L.Util.getBaseUrl() + "?" + L.Util.buildQueryString({feature: slug}) + window.location.hash;
+    },
+
     view: function(e) {
         if (this.map.editEnabled) return;
         var outlink = this.properties._umap_options.outlink,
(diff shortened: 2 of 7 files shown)


Binnette:
Hi @yohanboniface, can't we have 2 features with the same name? I think the answer is yes.

Don't we have an "internal" id for each feature ?

Ok, I need to test your code to understand :)

Binnette:
I confirm there is a bug when we have 2 or more feature with the same name. uMap will not alway open the good feature (depending of wich you clicked to create the permalink)

yohanboniface:
> Hi @yohanboniface, can't we have 2 features with the same name? I think the answer is yes.

Yes, we can ;)
In this proposal, `name` is just the default key to be used. But one may use another if they have a more unique value.

> Don't we have an "internal" id for each feature ?

Nope, not at this moment. Features are just part of a geojson, which is saved on disk,  and associated to a layer, which is saved on database.

Binnette:
Ok so we have no way to fix the bug i'm talking about then. :sob:

yohanboniface:
> I confirm there is a bug when we have 2 or more features with the same name.

yes, but that's not a bug, it's a known edge case of using a simple solution ;)

I've thought about adding the layer name in the slug generation, but that would make the slug more fragile (if you change the layer name, all its feature will change slug, so any URL used before then will be obsolete.

umap/static/umap/js/umap.core.js
```diff
@@ -428,6 +441,7 @@ L.U.Help = L.Class.extend({
     shortCredit: L._('Will be displayed in the bottom right corner of the map'),
     longCredit: L._('Will be visible in the caption of the map'),
     sortKey: L._('Property to use for sorting features'),
+    slugKey: L._('The name of the property to use as feature unique identifier.'),
```
Binnette: Oh. This is a new string to translate :)
I don't kwnow yet how tu update the english "source" file on Transifex.

umap/static/umap/js/umap.features.js
```diff
@@ -45,6 +45,15 @@ L.U.FeatureMixin = {
         return this.datalayer && this.datalayer.isRemoteLayer();
     },
 
+    getSlug: function () {
+        return this.properties[this.map.options.slugKey || 'name'] || '';
```
Binnette: slug ?
image

:smile:

umap/static/umap/js/umap.core.js
yohanboniface: `make tx_push` ;)

But I usually don't do it until the feature branch is merged on master.

umap/static/umap/js/umap.features.js
yohanboniface: You mean `slug` instead of `name` ?

umap/static/umap/js/umap.core.js
yohanboniface: Ah, sorry:

```
# Collect the new strings
make messages
# Push to transifex
make tx_push
```

umap/static/umap/js/umap.features.js
yohanboniface: Ah, ok `slug` means `limace`, I didn't know ;)

Cyrille37:
Merci !!

gislars:
If I have multiple POI at one location at a clustered layer, I can open the permalink, but the popup doesn't open. Would this be a bug, should I file an issue?

For example you can have a look at http://umap.openstreetmap.fr/de/map/glutenfree_6373

PR #772, opened 2020-03-25
Optimistic conflicts resolution mecanism

yohanboniface:
original code from @Biondilbiondo (Thanks! :) )

Basically:
- when there is a conflict (i.e. someone try to save with an old ETag)
- we compare new version with its reference version and with the latest known
- if we see only simple case (features added both side, removal on one side…),
  then we merge
- otherwise we raise the conflict as before

### DIFF
--- a/umap/static/umap/js/umap.layer.js
+++ b/umap/static/umap/js/umap.layer.js
@@ -1535,6 +1535,12 @@ L.U.DataLayer = L.Evented.extend({
     this.map.post(this.getSaveUrl(), {
       data: formData,
       callback: function (data, response) {
+        // Response contains geojson only if save has conflicted and conflicts have
+        // been resolved. So we need to reload to get extra data (saved from someone else)
+        if (data.geojson) {
+          this.clear()
+          this.fromGeoJSON(data.geojson)
+        }
         this._geojson = geojson
         this._last_modified = response.getResponseHeader('Last-Modified')
         this.setUmapId(data.id)
--- a/umap/tests/test_datalayer_views.py
+++ b/umap/tests/test_datalayer_views.py
@@ -1,8 +1,11 @@
 import json
+import time
+from copy import deepcopy
 from pathlib import Path
 
 import pytest
 from django.core.files.base import ContentFile
+from django.core.files.uploadedfile import SimpleUploadedFile
 from django.urls import reverse
 
 from umap.models import DataLayer, Map
@@ -19,7 +22,10 @@ def post_data():
         "display_on_load": True,
         "settings": '{"displayOnLoad": true, "browsable": true, "name": "name"}',
         "rank": 0,
-        "geojson": '{"type":"FeatureCollection","features":[{"type":"Feature","geometry":{"type":"Polygon","coordinates":[[[-3.1640625,53.014783245859235],[-3.1640625,51.86292391360244],[-0.50537109375,51.385495069223204],[1.16455078125,52.38901106223456],[-0.41748046875,53.91728101547621],[-2.109375,53.85252660044951],[-3.1640625,53.014783245859235]]]},"properties":{"_umap_options":{},"name":"Ho god, sounds like a polygouine"}},{"type":"Feature","geometry":{"type":"LineString","coordinates":[[1.8017578124999998,51.16556659836182],[-0.48339843749999994,49.710272582105695],[-3.1640625,50.0923932109388],[-5.60302734375,51.998410382390325]]},"properties":{"_umap_options":{},"name":"Light line"}},{"type":"Feature","geometry":{"type":"Point","coordinates":[0.63720703125,51.15178610143037]},"properties":{"_umap_options":{},"name":"marker he"}}],"_umap_options":{"displayOnLoad":true,"name":"new name","id":1668,"remoteData":{},"color":"LightSeaGreen","description":"test"}}',
+        "geojson": SimpleUploadedFile(
+            "name.json",
+            b'{"type":"FeatureCollection","features":[{"type":"Feature","geometry":{"type":"Polygon","coordinates":[[[-3.1640625,53.014783245859235],[-3.1640625,51.86292391360244],[-0.50537109375,51.385495069223204],[1.16455078125,52.38901106223456],[-0.41748046875,53.91728101547621],[-2.109375,53.85252660044951],[-3.1640625,53.014783245859235]]]},"properties":{"_umap_options":{},"name":"Ho god, sounds like a polygouine"}},{"type":"Feature","geometry":{"type":"LineString","coordinates":[[1.8017578124999998,51.16556659836182],[-0.48339843749999994,49.710272582105695],[-3.1640625,50.0923932109388],[-5.60302734375,51.998410382390325]]},"properties":{"_umap_options":{},"name":"Light line"}},{"type":"Feature","geometry":{"type":"Point","coordinates":[0.63720703125,51.15178610143037]},"properties":{"_umap_options":{},"name":"marker he"}}],"_umap_options":{"displayOnLoad":true,"name":"new name","id":1668,"remoteData":{},"color":"LightSeaGreen","description":"test"}}',
+        ),
     }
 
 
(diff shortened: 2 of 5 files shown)


umap/static/umap/js/umap.layer.js
```diff
@@ -1013,6 +1013,12 @@ L.U.DataLayer = L.Evented.extend({
         this.map.post(this.getSaveUrl(), {
             data: formData,
             callback: function (data, response) {
+                // Response contains geojson only if save has conflicted and conflicts have
+                // been resolved. So we need to reload to get extra data (saved from someone else)
+                if (data.geojson) {
+                    this.clear();
+                    this.fromGeoJSON(data.geojson);
```
yohanboniface: We may also reload the whole map in this case, to lower the possible conflicts in other layers. But that would be a bit more aggressive for the user to have whole page reloaded.

umap/tests/test_datalayer_views.py
```diff
@@ -14,34 +15,37 @@
...
-    if getattr(settings, 'UMAP_XSENDFILE_HEADER', None):
-        assert response['ETag'] is not None
-    assert response['Last-Modified'] is not None
-    assert response['Cache-Control'] is not None
-    assert 'Content-Encoding' not in response
+    if getattr(settings, "UMAP_XSENDFILE_HEADER", None):
+        assert response["ETag"] is not None
+    assert response["Last-Modified"] is not None
+    assert response["Cache-Control"] is not None
+    assert "Content-Encoding" not in response
```
yohanboniface: Bah, noise because I ran black by reflex…

umap/static/umap/js/umap.layer.js
yohanboniface: Other option instead of returning the geojson: make a 302 to the "normal" GET

umap/views.py
```diff
@@ -773,13 +781,51 @@ def is_unmodified(self):
                 modified = False
         return modified
 
+    def merge(self):
+        """Try to merge conflicts."""
+
+        # Reference data source of the edition.
+        modified_at = self.request.META.get("HTTP_IF_UNMODIFIED_SINCE")
+
+        for name in self.object.get_versions():
```
davidbgk, line 790: Given that `get_versions()` is reverse ordered by date, why don't we take the first element here? I'm probably missing a use-case with multi-users edit 🤔

umap/views.py
```diff
@@ -773,13 +781,51 @@ def is_unmodified(self):
...
+
+        # New data received in the request.
+        entrant = json.loads(self.request.FILES["geojson"].read())
+
+        # Latest known version of the data.
+        with open(self.path) as f:
+            latest = json.loads(f.read())
+
+        try:
+            merge_conflicts(
```
davidbgk, line 808: I would better understand if we reuse the value returned by `merge_conflict` likewise:
```suggestion
            merged_features = merge_conflicts(…)
            latest["features"] = merged_features
```

It's a bit more explicit than the in place modification of latest. Maybe `merge_features()` better describe what it does too?

umap/views.py
yohanboniface: For merging we need to find **three** versions:

- the current HEAD version (current)
- the one send by POST (candidate)
- the base version of the changes made by the candidate version (reference)

So we can compute the diff between the candidate and its reference, and then see if we can apply this diff on the current HEAD. If yes, bingo, no conflict. Otherwise, back to 422.

umap/views.py
yohanboniface: > Maybe merge_features() better describe what it does too?

Renamed!

> I would better understand if we reuse the value returned by merge_conflict

If I've correctly understood what you mean, we'll end up with something like:

```diff
diff --git a/umap/views.py b/umap/views.py
index 432e36f..8b0fdd7 100644
--- a/umap/views.py
+++ b/umap/views.py
@@ -794,11 +794,13 @@ class DataLayerUpdate(FormLessEditMixin, GZipMixin, UpdateView):
             latest = json.loads(f.read())
 
         try:
-            merge_features(
+            merged_features = merge_features(
                 reference["features"], latest["features"], entrant["features"]
             )
         except ValueError:
             return False
+        else:
+            latest["features"] = merged_features
 
         # Update request data.
         self.request.FILES["geojson"].file = BytesIO(json.dumps(latest).encode("utf-8"))
```

But then this line is really weird, because those two values are already the exact same:

    +            latest["features"] = merged_features


So I'm not sure what is best :/

yohanboniface:
One of the keys, if we go that direction, is to keep the relevant versions while there is an editing session going, so maybe instead of keeping the n last versions, we should combine that with the versions from the last hour (but this implies some cron to delete versions, which is less KISS).

almet:
Happens also on the latest master, I've opened an issue https://github.com/umap-project/umap/issues/1404

almet:
I've pushed an updated version of the code and tests.

It's still doing some "magic" stuff with the form that I'm not super happy with, but I would gladly see a review of these changes already.

umap/utils.py
```diff
@@ -117,3 +116,32 @@ def gzip_file(from_path, to_path):
...
+
+    reference_removed = [item for item in reference if item not in incoming]
+    incoming_added = [item for item in incoming if item not in reference]
+
+    # Ensure that items changed in the reference weren't also changed in the latest.
+    for removed in reference_removed:
+        if removed not in latest:
+            raise ConflictError()
+
+    merged = latest[:]
```
almet, line 138: We might actually do a `copy.deepcopy` here to avoid editing objects elsewhere.

umap/views.py
```diff
@@ -937,34 +941,83 @@ class DataLayerUpdate(FormLessEditMixin, GZipMixin, UpdateView):
...
-        self.object = form.save()
-        # Simple response with only metadatas (client should not reload all data
-        # on save)
-        response = simple_json_response(
-            **self.object.metadata(self.request.user, self.request)
-        )
-        response["Last-Modified"] = self.last_modified
-        return response
+    def has_been_modified_since(self, if_unmodified_since):
+        if if_unmodified_since and self.last_modified != if_unmodified_since:
```
almet, line 945: Maybe we could just `return if_unmodified_since and self.last_modified != if_unmodified_since`

umap/views.py
```diff
@@ -937,34 +941,83 @@ class DataLayerUpdate(FormLessEditMixin, GZipMixin, UpdateView):
...
-        return super(DataLayerUpdate, self).post(request, *args, **kwargs)
+
+        ius_header = self.request.META.get("HTTP_IF_UNMODIFIED_SINCE")
+
+        if self.has_been_modified_since(ius_header):
+            merged = self.merge(ius_header)
+            if not merged:
+                return HttpResponse(status=412)
+
+            # Replace the uploaded file by the merged version.
```
almet, line 999: I'm not a fan of this "let's overwrite what the client sent us", do you have a better idea?

umap/tests/test_datalayer_views.py
```diff
@@ -392,62 +393,157 @@ def test_anonymous_user_can_edit_if_inherit_and_map_in_public_mode(
...
+    post_data["geojson"] = SimpleUploadedFile(
+        "foo.json",
+        json.dumps(client1_data).encode("utf-8"),
     )
+    response = client.post(
+        url, post_data, follow=True, HTTP_IF_UNMODIFIED_SINCE=reference_timestamp
+    )
+    assert response.status_code == 200
+
+    # The first client changes the first feature.
```
yohanboniface, line 533: second ?

umap/utils.py
```diff
@@ -118,33 +118,30 @@ def is_ajax(request):
...
-            # element, so let's raise a conflict by caution.
-            raise ValueError
-
-    # We can merge.
-    for feature in reference:
-        latest.remove(feature)
-    for feature in entrant:
-        latest.append(feature)
-    return latest
+    reference_removed = [item for item in reference if item not in incoming]
```
yohanboniface, line 130: removed_from_reference ?

yohanboniface: Or just `removed` ? And then `added` ? Seems clear enough to me.

umap/views.py
```diff
@@ -1008,10 +1000,25 @@ def post(self, request, *args, **kwargs):
...
+
+            # Mark the data to ne reloaded by form_valid
             self.request.session["needs_reload"] = True
         return super().post(request, *args, **kwargs)
 
+    def form_valid(self, form):
+        self.object = form.save()
+        # Simple response with only metadata (client should not reload all data
+        # on save)
+
```
yohanboniface, line 1012: Not sure about this empty line ?

umap/views.py
```diff
@@ -1008,10 +1000,25 @@ def post(self, request, *args, **kwargs):
             self.request.FILES["geojson"].file = BytesIO(
                 json.dumps(merged).encode("utf-8")
             )
-            # Mark the
+
+            # Mark the data to ne reloaded by form_valid
```
yohanboniface, line 1004: ne => be ? :)

yohanboniface (commented):
Sounds good to me!
Didn't you said you wanted to use sets instead ?

almet:
>  Didn't you said you wanted to use sets instead ?

Actually, sets needs to have hashable elements inside, which is not our case at the moment, because we're storying dicts. and dicts of lists, so it's not possible :-)

umap/tests/test_datalayer_views.py
almet: Nope, that's the first. We don't actually care who sent the first base doc, so it's out of the picture.

yohanboniface: Humm, but how is the var name `client2_data` then ? :upside_down_face:

almet: Ah, that's the second one 😅 (sorry was missing context)

umap/utils.py
almet: Might be simpler yes 🙌

almet:
There is some magic done with the form handling here. It's useful in order to circumvent the way we use Django model forms.

Changing this would probably mean using a separated "update datalayer form", not tied to the model, so we can validate the form without altering the model (which seem to be how django does).

Here are some notes for later usage if needed:

---

While trying to change how the Form is used in a DataLayer update, I've understood how things work, so here are some notes.

I wanted the submitted form to be processed and validated, and then merged with a document already present in the database.

However, since the form is a `ModelForm`, it must be linked to something in the database which doesn't seem to be the case here. On data validation the FileField is considered a new field, meaning that the geojson object is overwritten by the new one, and `self.path`, being derived from `self.object.geojson` is updated.

PR #839, opened 2020-11-07
Update requirements.txt for Pillow with python3.9

mmeinert:
Bump Pillow version to 8.0.1
Pillow version >= 8.0 is mandatory for python3.9 (default python version on current Ubuntu LTS 20.04


yohanboniface:
Thanks! :)

Issue #820, opened 2020-08-26
Add coalesce function in Popup content template


sfkeller:
Often attributes can be NULL but for display purposes one wants to show "0" or another value (like "-").

Proposal: Add a coalesce function in "Popup content template" which returns the first non-NULL value from the parameter list, similar to QGIS or PostgreSQL.

sfkeller:
Here's a use case of a template, where the attribute/value of ele isn't always set:

    # {name} ({ele} m ü. M.)

=> shows: Piz Signina ( m ü. M.)

This is the proposal:

    # {name} (coalesce({ele},'-') m ü. M.)

=> shows: Piz Signina (- m ü. M.)

das-g:
I think the syntax should be more like

    # {name} ({coalesce(ele,'-')} m ü. M.)

or maybe something more compact like

    # {name} ({ele|'-'} m ü. M.)

Binnette:
Sorry, i don't get it. Can you add a screenshot of the issue?

yohanboniface:
As I understand it, the need is to display a fallback value when a template variable points to an empty value.

The uMap templating system (to say a big word) is very minimal (because uMap is targetted as users without tech skills), and I'm not sure it's feasable without adding a lot of complexity, but I'll have a look.

das-g:
> As I understand it, the need is to display a fallback value when a template variable points to an empty value.

Exactly. For a quick grasp of SQL's `COALESCE` feature that @sfkeller referred to, see [this tweet](https://twitter.com/b0rk/status/1181201631557357568) by @jvns. To quote from that tweet:

> `COALESCE` is a SQL function that returns the first argument you give it that isn't `NULL`
> 
> `COALESCE(NULL, 1, 2)` &rarr; `1`
> `COALESCE(NULL, 'hi')` &rarr; `'hi'`
> `COALESCE(4, NULL, 2)` &rarr; `4`

Importantly, it allows specifying not only hard-coded fallback values, but also fallback columns (or more generally, fallback expressions). This is something we might want for a corresponding feature in the uMap "templating system", too, e.g.

* if `ele` is `NULL`, use `0` instead
* if `ele` is `NULL`, use `'-'` instead

(hard-coded fallback value)

but also

* if `addr_street` is `NULL`, use `addr_place` instead

(lookup in an alternative fallback column)

yohanboniface:
I've made a proposal in this direction:

`{key|fallbackkey|"fallbackstring"}`

[...]

PR #862, opened 2020-12-13
Fix and enhance property fallbacks

mstock:
The new 'coalesce feature' from issue #820 was mentioned at our latest (virtual) [Swiss/Zurich OSM meeting](https://wiki.openstreetmap.org/wiki/DE:Switzerland:Z%C3%BCrich/OSM-Treffen#124._OSM-Stammtisch_Fr._11._Dezember_2020). However, it did not seem to work at all on [our](https://umap.osm.ch/) uMap instance. Although this turned out to have been a simple deployment issue, when looking at the code, I also noticed a small implementation error which made it always use the last fallback if there were any, even if the first property lookup or any of the previous fallbacks returned a defined value. I also noticed that the characters that could be used in the static fallback were quite limited such that the original example from issue #820 would not have worked (even when adapted to the final syntax). The changes in this pull request should fix these two issues.

I didn't look into fixing the issue with failing fallback if the property is called `description` though - I assume this is caused by the line below `// Resolve properties inside description` in the `renderBody` function of `L.U.PopupTemplate.Default` where its value gets set to the empty string even if it didn't exist in the original properties, so there will always be a defined `description` property when evaluating the template in the next but one line. Might be worth fixing if somebody gets around to it.

### DIFF
--- a/umap/static/umap/js/umap.core.js
+++ b/umap/static/umap/js/umap.core.js
@@ -134,8 +134,11 @@ L.Util.greedyTemplate = function (str, data, ignore) {
         return value;
     }
 
-    return str.replace(/\{ *([\w_\:\."\|]+) *\}/g, function (str, key) {
+    return str.replace(/\{ *([\w_\:\.\|]+)(?:\|("[^"]*"))? *\}/g, function (str, key, staticFallback) {
         var vars = key.split('|'), value, path;
+        if (staticFallback !== undefined) {
+            vars.push(staticFallback);
+        }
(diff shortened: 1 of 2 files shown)


sfkeller:
@yohanboniface Pouvons-nous ou quelqu'un peut-il aider à examiner la demande de retrait (pull request) et aider à la fusionner (merge)?

yohanboniface:
Sorry, can't remember why I did not merge in the first place; I had something to check, but it's out of my mind now, so let's proceed.

Thanks for the work, and sorry again for the delay!

PR #1031, opened 2022-11-24
Use Leaflet.Locate instead of a custom locate control

### LINKED ISSUE #923: Localisation is locked when using it
Using the localisation feature could be more comfortable.

Now:
Enabling localisation locks the zoom level. I cannot zoom in or out while using localisation.

How it should be:
Enabling localisation shows the blue dot where my current location is and zooms close to it.
Zooming in and out is still possible. Moving into the map to another location is possible but after a certain while (e.g. 10 sec) the map will automatically center again to the current location (without changing the zoom level!).

aleksejspopovs:
The current geolocation behavior is somewhat limited: it centers the map on the user's location whenever a new location comes in, which makes it impossible to pan and zoom the map when location updates come in regularly (e.g. on a phone with GPS).

This PR replaces the control with a Leaflet plugin, Leaflet.Locate, that implements a more sophisticated behavior: location updates only move the map if the user has not panned or zoomed it since enabling geolocation. When the location is outside of the current viewport, clicking the control again will center the location. On devices with a compass, we also get a direction arrow.

Fixes #923.


umap/static/umap/js/umap.js
```diff
@@ -250,7 +250,18 @@ L.U.Map.include({
...
-        this._controls.locate = new L.U.LocateControl();
+        this._controls.locate = L.control.locate({
+            strings: {
+                title: L._('Center map on your location'),
+            },
+            showPopup: false,
+            // We style this control in our own CSS for consistency with other controls,
+            // but the control breaks if we don't specify a class here, so a fake class
+            // will do.
+            icon: 'umap-empty',
```
yohanboniface, line 261: Watch out, there is a `umap-empty` css [class](https://github.com/umap-project/umap/blob/7893ff1c7d943b7fb71b74dbd8a3b47b7b104ca3/umap/static/umap/base.css#L471) ;)

yohanboniface (changes requested):
Thanks a lot!

umap/static/umap/js/umap.js
aleksejspopovs: Oh! I really should have checked :sweat_smile:

yohanboniface:
Very cool, thanks a lot! :)

PR #1041, opened 2023-01-14
New asked features: checkbox filters, permanent credits, and caption menus options

k-3st3ban:
This pull request is divided into 3 separated commits:
1. Filters checkboxes using features' properties (the user sets which properties should be used for filtering in the map options, then checkboxes appear in the filtering tab)
2. Permanent credits feature in the bottom left corner (the user writes those credits from the map settings, like other credits, and has the ability to hide the permanent credit's background, for example, to only show an image/logo)
3. Interface option to hide caption menus (interface option to hide menus in the caption bar: "Browse data", "Filter data", "About" and "Home")

### DIFF
--- a/umap/static/umap/base.css
+++ b/umap/static/umap/base.css
@@ -508,6 +508,7 @@ i.info {
 .umap-layer-properties-container,
 .umap-footer-container,
 .umap-browse-data,
+.umap-filter-data,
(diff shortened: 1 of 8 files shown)


yohanboniface:
:+1:

PR #1058, opened 2023-03-27
🐛 — Allow to use SHA1-signed anonymous edit URL

ldidry:
The default django.core.signing Signer uses SHA256 algorithm since Django 3. Umap used Django 2 in the paste, so people had SHA1 signed anonymous edit URLs, which became unusable when umap switch to Django 3. This commit makes them usable again (the new SHA256-signed anonymous edit URLs still works, obviously).

### DIFF
--- a/umap/tests/test_map_views.py
+++ b/umap/tests/test_map_views.py
@@ -4,6 +4,7 @@ import pytest
 from django.contrib.auth import get_user_model
 from django.urls import reverse
 
+from django.core.signing import Signer
 from umap.models import DataLayer, Map
 
 from .base import login_required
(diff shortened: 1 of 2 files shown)


yohanboniface:
Thanks Luc!

Do you think it's worth adding a test to cover this ? Here is a test that can serve as example https://github.com/umap-project/umap/blob/master/umap/tests/test_map_views.py#L392-L401

Also, uMap master is now on Django 4 (and a new release will come soon), would you mind checking that this upgrade will not break again the URLs ? Thanks! :)

ldidry:
SHA256 is still the default for Django 4 (and 4.1, and the dev version) according to the [documentation](https://docs.djangoproject.com/en/dev/topics/signing/).

ldidry:
Could you please try the test I added to the commit?

yohanboniface:
All good! :)

PR #1067, opened 2023-04-23
Optimize SVG icons sizes

davidbgk:
Using Jake Archibald’s tool: https://jakearchibald.github.io/svgomg/ with default settings except keeping width/height attributes because some old browsers require default sizes for rendering (it also avoid huge/100% icons in case the CSS is not applied for whatever reason).


yohanboniface:
:+1:

Issue #834, opened 2020-10-31
Display distance while drawing lines


Cyrille37:
Hello
Thanks for that so great tool !

It should be nice to display length (aka distance) of lines while drawing them. Like there is a "distance" tool and a "line" drawing tool it should be not so hard to implements ?

Regards & Cheers

yohanboniface:
Nice suggestion. I'll keep open, but I've no time to work on this now.

oli44:
Would be much appreciated: how can I go for a daily walk inside a 1km radius around my house?

woodpeck:
@yohanboniface I would like to build this feature into umap and make a pull request for it. Is umap.openstreetmap.fr running the current Github version of umap so that the change would then end up there at some point, or are you running a fork of some kind?

yohanboniface:
Hi @woodpeck :)

It's running the latest release, not the current master, but new release is in process. 

The new release in this case depends on a server upgrade to have a newer python, so it's not as simple as "pip install", so I'm unable to give an idea of when it will occur, but let's say "soon", cc @cquest :)

yohanboniface:
Sounds like a good step forward to me, thanks!

About the translation, one other option would have been to keep the variable out of the translated string: `L._("blaba bla") + "(" + distance + ")"` (that does not allow for translators to move the variable in the sentence, but may be enough as a first step).

Anyway, given you have already make the effort to update each string, we could also proceed as is :)

So, yeah, let's put that in a PR! :+1:

PR #1068, opened 2023-04-24
Show distance when drawing Polylines.

knowname:
Implements feature request #834.

### DIFF
--- a/umap/static/umap/js/umap.controls.js
+++ b/umap/static/umap/js/umap.controls.js
@@ -1198,7 +1198,7 @@ L.U.Editable = L.Editable.extend({
 
     initialize: function (map, options) {
         L.Editable.prototype.initialize.call(this, map, options);
-        this.on('editable:drawing:start editable:drawing:click', this.drawingTooltip);
+        this.on('editable:drawing:start editable:drawing:click editable:drawing:move', this.drawingTooltip);
         this.on('editable:drawing:end', this.closeTooltip);
         // Layer for items added by users
         this.on('editable:drawing:cancel', function (e) {
@@ -1250,18 +1250,26 @@ L.U.Editable = L.Editable.extend({
 
     drawingTooltip: function (e) {
         var content;
+        var readableDistance;
         if (e.layer instanceof L.Marker) content = L._('Click to add a marker');
         else if (e.layer instanceof L.Polyline) {
             if (!e.layer.editor._drawnLatLngs.length) {
-                if (e.layer instanceof L.Polygon) content = L._('Click to start drawing a polygon');
-                else if (e.layer instanceof L.Polyline) content = L._('Click to start drawing a line');
-            } else if (e.layer.editor._drawnLatLngs.length < e.layer.editor.MIN_VERTEX) {
-                content = L._('Click to continue drawing');
+                if (e.layer instanceof L.Polygon){
+                    content = L._('Click to start drawing a polygon');
+                } else if (e.layer instanceof L.Polyline) {
+                    content = L._('Click to start drawing a line');
+                }
             } else {
-                content = L._('Click last point to finish shape');
+                var tempLatLngs = e.layer.editor._drawnLatLngs.slice();
+                tempLatLngs.push(e.latlng);
+                var length = L.GeoUtil.lineLength(this.map, tempLatLngs);
+                readableDistance = L.GeoUtil.readableDistance(length, this.map.measureTools.getMeasureUnit());
+                content = L._('Click last point to finish shape ({distance})', {distance: readableDistance});
             }
         }
-        if (content) this.map.ui.tooltip({content: content});
+        if (content) {
+            this.map.ui.tooltip({content: content});
+        }
     },
 
     closeTooltip: function () {
--- a/umap/static/umap/locale/am_ET.js
+++ b/umap/static/umap/locale/am_ET.js
@@ -115,7 +115,7 @@ var locale = {
     "Choose a preset": "ፕሪሴት ምረጥ",
     "Choose the format of the data to import": "ሊያመጡ የፈለጉትን የመረጃ ፎርማት ይምረጡ",
     "Choose the layer to import in": "የሚያስገቡበት ሌየር ይምረጡ",
-    "Click last point to finish shape": "ቅርፁን ለማስጨረስ የመጨረሻውን ነጥብ ጠቅ ያድርጉ",
+    "Click last point to finish shape ({distance})": "ቅርፁን ለማስጨረስ የመጨረሻውን ነጥብ ጠቅ ያድርጉ ({distance})",
     "Click to add a marker": "መለያ ለመጨመር ጠቅ አድርግ",
     "Click to continue drawing": "መሳል ለመቀጠል ጠቅ አድርግ",
     "Click to edit": "ለማረም ጠቅ አድርግ",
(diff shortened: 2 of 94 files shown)


yohanboniface:
First, I want to be sure I fully understand the feature we want: do we want the total length/area of the line/polygon being drawn or something else ? :)

If yes, this is a good starting point, but there are a few things left that need a bit more work:

- we also need to deal with a polygon being drawn (so compute an area instead of a line length)
- we also need to deal with an existing line or polygon being modified (eg. moving an edge point)
- I suggest to do a bit of refactor, so the `getMeasure` ([here](https://github.com/umap-project/umap/blob/master/umap/static/umap/js/umap.features.js#L856-L859) and [here](https://github.com/umap-project/umap/blob/master/umap/static/umap/js/umap.features.js#L1007-L1010)) are being used

How do you feel with those changes ? Do you need help ? Should we do part of it ? Or do you feel comfortable with doing this by yourself ?

Thanks any way pushing this already! :)

knowname:
I am showing the polyline which has been drawn plus the dotted line which is about to be drawn.
Its constantly updating while moving the cursor.
I think this has more use cases than only showing how long the Polyline currently is.

Thats also why I didn't use the Polyline.getMeasure() function. I clone the latlng-array, add the mouse position as coordinate and recalculate the length constantly when moving the cursor.

What I could do is creating a new temporary Polyline instead and call `getMeasure()` on it. But i am not sure if this brings other difficulties.

Lets finish the polyline first, then do the polygons?

knowname:
I extended the getMeasure() function by `extraPoint`. It works for polygon + polyline. Also fixed the bug where it broke when editing an existing point.

Do you have a suggestion how to deduplicate https://github.com/umap-project/umap/blob/3f12b69c25392bff5e086212e181e59e6af3d45f/umap/static/umap/js/umap.features.js#L1016-L1023 and https://github.com/umap-project/umap/blob/3f12b69c25392bff5e086212e181e59e6af3d45f/umap/static/umap/js/umap.features.js#L857-L864 ?

yohanboniface:
Ah, it seems that now the measure is not update *while* drawing the last point (i.e. it does not includes the currently moving point), is that on purpose ?

umap/static/umap/js/umap.controls.js
```diff
@@ -1252,16 +1252,37 @@ L.U.Editable = L.Editable.extend({
...
+                    }  else {
+                        // when drawing third point (or more)
+                        content = L._('Click last point to finish shape ({distance})', { distance: readableDistance });
+                    }
+                }
             } else {
-                content = L._('Click last point to finish shape');
+                // when moving an existing point
+                if (e.layer instanceof L.Polygon){
+                    content = L._('Polygon area: {area}', { area: e.layer.getMeasure() });
```
yohanboniface, line 1277: Maybe we could always keep this phrase, and when drawing append it after the "Click blah blah", what do you think ?

umap/static/umap/js/umap.controls.js
```diff
@@ -1252,16 +1252,37 @@ L.U.Editable = L.Editable.extend({
...
+                } else if (e.layer instanceof L.Polyline) {
+                    content = L._('Line distance: {distance}', { distance: e.layer.getMeasure() });
```
yohanboniface, line 1279: I'd name the parameter `measure` instead of `distance`, what do you think ?

umap/static/umap/js/umap.controls.js
yohanboniface: (And in this case you don't have to deal with translation)

knowname: > Ah, it seems that now the measure is not update _while_ drawing the last point (i.e. it does not includes the currently moving point), is that on purpose ?

It does _not_? When i create a new polyline, on my end it looks like this (at least in Firefox and Safari):
Bildschirmfoto 2023-04-27 um 11 01 48

yohanboniface:
> It does not? When i create a new polyline, on my end it looks like this (at least in Firefox and Safari):

It does display a measure, but this measure does not include the "currently drawn" part of the line (the dashed):

image

So here 371 km is only the total of the "finished" part of the line. I suppose that feature is to have also the drawn part (so we can adapt what we do), but I'm not sure. :)

knowname:
> > It does not? When i create a new polyline, on my end it looks like this (at least in Firefox and Safari):
> 
> It does display a measure, but this measure does not include the "currently drawn" part of the line (the dashed):
> 
> 
> So here 371 km is only the total of the "finished" part of the line. I suppose that feature is to have also the drawn part (so we can adapt what we do), but I'm not sure. :)

Could you try again? This is how it should look like:

https://user-images.githubusercontent.com/1724290/234837537-eec69cd8-5212-4928-9def-87d94558e19f.mov

https://user-images.githubusercontent.com/1724290/234837490-0f2f63b9-3912-4624-bdd2-85cc4255d002.mov

umap/static/umap/js/umap.controls.js
```diff
@@ -1252,16 +1252,40 @@ L.U.Editable = L.Editable.extend({
...
+                    }
+                    content += " ("+readableDistance+")";
+                }
             } else {
-                content = L._('Click last point to finish shape');
+                // when moving an existing point
+                if (e.layer instanceof L.Polygon){
+                    content = L._('Polygon area: {measure}', { measure: e.layer.getMeasure() });
+                } else if (e.layer instanceof L.Polyline) {
+                    content = L._('Line distance: {measure}', { measure: e.layer.getMeasure() });
```
yohanboniface, line 1282: No sure about my English, but I feal like this should be `Line length` instead of `Line distance`, what do you think ?

Also, at this stage, I think you can leave out the translations, this will be done automatically by Transifex later in the process.

yohanboniface (commented):
Hey! A few more small comments, but we are very close to the finish line, thank you very much! :)

Also, if you feel you'll miss time to finish this up, just tell me I can merge and do the polishing myself :)

umap/static/umap/js/umap.controls.js
```diff
@@ -1252,16 +1252,40 @@ L.U.Editable = L.Editable.extend({
...
+                    tmpLatLngs.push(e.latlng);
+                    var readableDistance = e.layer.getMeasure(tmpLatLngs);
+                    if (e.layer.editor._drawnLatLngs.length < e.layer.editor.MIN_VERTEX) {
+                        // when drawing second point
+                        content = L._('Click to continue drawing');
+                    }  else {
+                        // when drawing third point (or more)
+                        content = L._('Click last point to finish shape');
+                    }
+                    content += " ("+readableDistance+")";
```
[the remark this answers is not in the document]
knowname: i am now appending the `Polyon area` / `Line length` part if necessary. So its always being used.

knowname:
i refactored a litte more.

If we have reached the point where its easier for you to just do it yourself, do it! :)
Thanks for your comments!

One more thing: I found a bug when drawing multiple polylines. `getMeasure()` always uses `this._defaultShape()` which seems to only return the first drawn Geometry (Polyline/Polygon). What we probably want is the last drawn geometry instead or when moving an existing point the selected geometry.
Not sure at which part we should tackle this problem.

yohanboniface:
Thanks for you work! Now we need a release.

Which instance are you using ?

See https://github.com/osm-fr/infrastructure/issues/444 to follow the upgrade of the OSM FR instance, in case you use that one.

PR #1069, opened 2023-04-28
Setup to create integration tests with Playwright

davidbgk:
More as a working base for a discussion on the topic.


Makefile
```diff
@@ -1,7 +1,13 @@
 test:
 	py.test -xv umap/tests/
+
+test-integration:
+	DJANGO_ALLOW_ASYNC_UNSAFE=1 py.test -xv umap/tests/integration/
```
davidbgk, line 5: There is probably a more clever way to avoid that `DJANGO_ALLOW_ASYNC_UNSAFE` 🤷

Issue #890, opened 2021-03-12
Feature request: pop-up on mouse-over / hover


mary-loulou:
It would be so great if the viewer didn't have to click on a marker to show the pop-up.
Can't you make it available on mouse-over?

Somebody made a marker popup in Leaflet (1.3.1) that appears on mouse-over, so it could work:
https://gis.stackexchange.com/questions/271602/show-popup-on-marker-hover-mouseover-hide-on-mouseout-and-keep-open-on-click
https://i.stack.imgur.com/16TEH.gif

PR #1074, opened 2023-05-01
Make visible titles from the control bar on hover

davidbgk:
This is an attempt at making labels of buttons from the control bar more explicit when hovering (it reuses the `title` attribute of these links).

umap-titles-hover


davidbgk:
For the border radius, we can fake the junction with a few less px of the negative margin, for the arrow though it will get harder because we are already dealing with an extra element from `::before` and no extra markup.

Anyway, that was more a proof of concept to verify it it's useful or not to go down that way.

davidbgk:
I guess this would fix https://github.com/umap-project/umap/issues/890

Issue #88, opened 2014-07-15
Add a Print function


yohanboniface:
Originally reported by: **BitBucket: [leftshifted](http://bitbucket.org/leftshifted), GitHub: [leftshifted](http://github.com/leftshifted)**

---

It would be really nice to have an Option to print the map.
I know printing from HTML can be very anoying because browsers behave unexpectable.
Still, something like a "Print-mode" which lets users arrange the Map and a list of all points (and other objects), adds a copyright notice (and maybe a "Generated with umap") and then spits it out in a reasonable size and resolution (maybe a few options like custom size, A4, letter, whatever) would be nice.
If no one else wants to try to tackle this, I might also fork and try to get something done, just know that I have somewhat limited time and skills.

Really enjoying all other aspects of umap, this should be so much more public, amazingly simple project!

---
- Bitbucket: https://bitbucket.org/yohanboniface/umap/issue/88

thewholelifetolearn:
:+1: 

I would enlarge this proposal to be more generic. It would be great if it could be possible to export in different images format (SVG, PNG, JPEG, PDF, ...). It would ease the printing and could be also reused in documents.

jphnavarro:
C'est vraiment une fonctionnalité qui manque dans uMap... je plussoie !

trofima:
So 3.5 years passed. Is there any progress on that?

hannesw:
+1 for that - PDF export would be great!

Dubidubiduu:
+1
Or at least a way to export the map. Including style!

Guddi11:
+1 
Definitely would love an export functionality, may it be pdf or reasonably sized image!!
Is there any attempt in creating that?

clembeb:
+1
It would be very nice to have an export to pdf functionality.

sylvain-m:
+1 :+1: 
A mon tour de plussoyer : on vient de s'en sortir avec des captures d'écran, mais une fonction d'impression (PDF) serait beaucoup plus pratique...

On my turn to add :+1:  we just got out with screenshots, but a print function (PDF) would be much more convenient ...

crti:
[...]

PR #1075, opened 2023-05-03
Allow the user to print the current viewed map

davidbgk:
Refs #88

### DIFF
--- a/umap/static/umap/js/umap.controls.js
+++ b/umap/static/umap/js/umap.controls.js
@@ -86,6 +86,39 @@ L.U.UpdatePermsAction = L.U.BaseAction.extend({
   },
 })
 
+L.U.PrintAction = L.U.BaseAction.extend({
+
+    options: {
+        className: 'print-map dark',
+        tooltip: L._('Print the current view')
+    },
+
+    addHooks: function () {
+        // dom-to-image creates the image from the map which is then inserted
+        // and opened within a new tab to be able to print it, we close the
+        // window in case the body regain the focus (printing cancelled).
+        domtoimage
+            .toPng(this.map._container)
+            .then(function (dataUrl) {
+                var win = window.open('about:blank', "_new");
+                win.document.open();
+                win.document.write(`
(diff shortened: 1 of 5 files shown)


davidbgk:
Refs https://github.com/1904labs/dom-to-image-more/issues/50 too.

yohanboniface:
Exciting!

Two first comments:
- I'd put the print button on the left/white bar, which the one for **reading** the map, while the right/black one is dedicated to editing it
- I'd hide all the map controls

yohanboniface:
Ah, bug, while using a map with a lot of objects, not sure if it's related:

image

yohanboniface:
Same with "a lot" of points:

image

davidbgk:
> I'd put the print button on the left/white bar, which the one for reading the map, while the right/black one is dedicated to editing it

I tried at first but it was a bit too complicated for a first time, let's adapt that together.

> I'd hide all the map controls

There is a mechanism to [filter out elements](https://github.com/1904labs/dom-to-image-more#filter) but I didn't try it yet!

> Ah, bug, while using a map with a lot of objects, not sure if it's related:

The two errors are different, we might hit some limitations at some point, the dom-to-image tool is actually re-rendering the whole page to be able to convert it.

Also https://github.com/1904labs/dom-to-image-more has probably more bugs fixed but we have to retrieve the fix for iframes (see comments at the top of the copied file).

yohanboniface:
> There is a mechanism to [filter out elements](https://github.com/1904labs/dom-to-image-more#filter) but I didn't try it yet!

Can't we use a print css ?

davidbgk:
We tried https://html2canvas.hertzen.com/ but:

> If you wish to load images that reside outside of your pages origin, you can use a [proxy](https://html2canvas.hertzen.com/proxy) to load the images. — https://html2canvas.hertzen.com/faq

So we do not have background tiles in the preview of the map.

yohanboniface:
Superseded by #2866

Issue #905, opened 2021-05-05
please make the Link to a map more unique


rocky-III:
i really like umap - Thank you!

Here is a feature request: 
It would be nice if the link to a map would bit more complex if the option "Who can view" is set to "anyone with link" .

Right now the link looks like this:

https://domain-name.de/en/map/NAME_12345

"https://domain-name.de/en/map/" is always the same
"NAME" does not really matter because if you know the number _12345 and put any NAME in front of it  - the link still works - so:
 
https://domain-name.de/en/map/NAME_12345

will link me to the same map as 

https://domain-name.de/en/map/NAMExx_12345

Since the number at the end is an ongoing upcount,, raised my 1 - it is easy to view all created maps and find the link for the map your are looking for. - The link is not really a secret 

If the NAME would matter or the number an the end would be a bit more complex like cryptpad https://github.com/xwiki-labs/cryptpad is doing it  - the share  feature by "anyone with link" would be much more usable. 

THANKS

ilyamikheevcoder:
i really think this should be talked about more. potential danger of link sharing while this issue is present is not documented anywhere, and i'm pretty that it _will_ eventually catch an unaware user off-guard!

yohanboniface:
This is indeed a concern, but it's not easy to fix with respect for history: if we change the URL form (eg. uuid instead of id), what should happen for previous URL, which are embed in HTML code else where ?

We could anyway fix if in opt in way, so new instances of uMap could use uuid from the beginning.

PR welcome :)

almet:
> This is indeed a concern, but it's not easy to fix with respect for history: if we change the URL form (eg. uuid instead of id), what should happen for previous URL, which are embed in HTML code else where ?

I believe we could catch this on the server side? Issue new IDs (UUIDs) for new maps (with a different URL scheme), and still support legacy IDs for old maps ? :thinking:

PR #1080, opened 2023-05-06
Check both matching pk and slug for map urls

davidbgk:
Note: this is some kind of breaking change if previous URLs have incorrect (but still working) slugs.

Also, the URL change when changing the title and now the resulting slug is significant during the matching.

PS: a clean commit is available to filter out black noise.

### DIFF
--- a/umap/tests/test_map_views.py
+++ b/umap/tests/test_map_views.py
@@ -16,50 +16,47 @@ User = get_user_model()
 @pytest.fixture
 def post_data():
     return {
-        'name': 'name',
-        'center': '{"type":"Point","coordinates":[13.447265624999998,48.94415123418794]}',  # noqa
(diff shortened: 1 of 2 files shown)


yohanboniface:
I'm not sure to see how this answers the #905 : if we still support the pk, then we still have the issue, or I'm missing something ?

davidbgk:
> I'm not sure to see how this answers the #905 : if we still support the pk, then we still have the issue, or I'm missing something ?

With the `query_pk_and_slug` option, the check will be performed on both the `pk` _and_ the `slug` so you cannot guess the URL from the incremental `pk` only.

yohanboniface:
:+1:

davidbgk:
I'm closing it as we will not solve that issue with that strategy. We plan to define a proper UUID field on the model and use it for future maps because we have a lot of maps out there and we do not want to break the web. It's a trade-off.

Amongst the potential implementations for future devs, there is a view that would check if a UUID is available for a given map and deal with it from the URL (instead of current id) or return a 404. But if there is no UUID for that legacy map, in that case fallback to the ancient access of the map, with id, to maintain maps' URLs working.

PR #1081, opened 2023-05-09
Basic configuration for JS code consistency + modern JS syntax

davidbgk:
Using PrettierJS and an opinionated configuration (mainly: no semi, two-spaces indent) + lebab for modern JS syntax.

### DIFF
--- a/.prettierrc.yaml
+++ b/.prettierrc.yaml
@@ -1,0 +1,6 @@
+trailingComma: "es5"
+tabWidth: 2
+semi: false
+singleQuote: true
+printWidth: 88
+quoteProps: "consistent"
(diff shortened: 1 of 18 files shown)


yohanboniface:
Very cool! Let's release and then embrace the future ? :)

davidbgk:
@yohanboniface do I run the scripts on all JS files of the repository, I just checked and there are no open PR that should be problematic to merge? 🤔

davidbgk:
It's time to test that branch as much as we can!

davidbgk:
Plan after discussion with @yohanboniface : we will prettier-ify the whole JS codebase (new incoming PR) and then apply lebab's transformations one-by-one (each with a dedicated PR).

PR #1083, opened 2023-05-11
Use flyTo option in locate if map settings aks for it

yohanboniface:
We removed it by mistake in 25ab9f943c59c15eee0eb3551008dfbd692a3479

cf https://github.com/domoritz/leaflet-locatecontrol#possible-options


davidbgk (approved):
Oopsie

PR #1086, opened 2023-05-12
Install and apply Lebab for JS arrows’ conversions

davidbgk:
[Arrow functions](https://caniuse.com/arrow-functions) will exclude IE11 but global usage is at  96.96%

### DIFF
--- a/umap/static/umap/js/umap.autocomplete.js
+++ b/umap/static/umap/js/umap.autocomplete.js
@@ -130,7 +130,7 @@ L.U.AutoComplete = L.Class.extend({
 
   onBlur: function () {
     var self = this
-    setTimeout(function () {
+    setTimeout(() => {
       self.hide()
     }, 100)
   },
@@ -207,7 +207,7 @@ L.U.AutoComplete = L.Class.extend({
 
   resultToIndex: function (result) {
     var out = null
-    this.forEach(this.RESULTS, function (item, index) {
+    this.forEach(this.RESULTS, (item, index) => {
       if (item.item.value == result.item.value) {
         out = index
         return
@@ -221,7 +221,7 @@ L.U.AutoComplete = L.Class.extend({
     this.clear()
     this.container.style.display = 'block'
     this.resizeContainer()
-    this.forEach(data, function (item) {
+    this.forEach(data, (item) => {
       self.RESULTS.push(self.createResult(item))
     })
     this.CURRENT = 0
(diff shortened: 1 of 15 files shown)


umap/static/umap/js/umap.autocomplete.js
```diff
@@ -130,7 +130,7 @@ L.U.AutoComplete = L.Class.extend({
 
   onBlur: function () {
     var self = this
-    setTimeout(function () {
+    setTimeout(() => {
```
yohanboniface, line 133: Couldn't we leave out the anonymous function totally at the end ?

umap/static/umap/js/umap.features.js
```diff
@@ -622,7 +622,7 @@ L.U.Marker = L.Marker.extend({
   zoomTo: function (e) {
     if (this.datalayer.isClustered() && !this._icon) {
       // callback is mandatory for zoomToShowLayer
-      this.datalayer.layer.zoomToShowLayer(this, e.callback || function () {})
+      this.datalayer.layer.zoomToShowLayer(this, e.callback || (() => {}))
```
yohanboniface, line 625: Also unrelated, but maybe we should allow an undefined or null callback in zoomToShowLayer instead of this ?

umap/static/umap/js/umap.layer.js
```diff
@@ -375,7 +375,7 @@ L.U.DataLayer = L.Evented.extend({
       verb: 'GET',
       callback: function (raw) {
         self.clear()
-        self.rawToGeoJSON(raw, self.options.remoteData.format, function (geojson) {
+        self.rawToGeoJSON(raw, self.options.remoteData.format, (geojson) => {
```
yohanboniface, line 378: We could remove the anonymous, nope ?

yohanboniface (approved):
Let's go! :+1:

umap/static/umap/js/umap.controls.js
```diff
@@ -752,11 +752,11 @@ L.U.Map.include({
       }
       build()
       datalayer.on('datachanged', build)
-      datalayer.map.ui.once('panel:closed', function () {
+      datalayer.map.ui.once('panel:closed', () => {
         datalayer.off('datachanged', build)
       })
-      datalayer.map.ui.once('panel:ready', function () {
-        datalayer.map.ui.once('panel:ready', function () {
+      datalayer.map.ui.once('panel:ready', () => {
```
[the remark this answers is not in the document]
davidbgk: Neither do I 🫣

umap/static/umap/js/umap.autocomplete.js
davidbgk: Yup, let's do it together, I prefer that we check that the feature still works once we rewrite these parts.

PR #1089, opened 2023-05-18
Lazy load tile layers thumbnails

davidbgk:
The list can be pretty big (like on OSMfr):
Capture d’écran, le 2023-05-18 à 12 55 49


davidbgk:
By the way, some thumbnails return 404 and we cannot load tiles when selected, I guess it's an admin configuration that cannot be fixed via code.

yohanboniface:
> By the way, some thumbnails return 404 and we cannot load tiles when selected, I guess it's an admin configuration that cannot be fixed via code.

Yes, a bit of cleaning needs to be done

PR #1090, opened 2023-05-18
Display the steps for inputs of type range

### LINKED ISSUE #877: Scale on thickness slider
Would it be feasible to implement a scale to the thickness/strength slider for lines? When trying to make several lines the same thickness, it is more a guessing game of where exactly to position the slider to get the same thickness. A scale, maybe even with lock-in on even values could fix that.

davidbgk:
Fix #877

This is not as slick as I hoped for but better than nothing 🤷 

Firefox:
Capture d’écran, le 2023-05-18 à 14 15 12

Chromium:
Capture d’écran, le 2023-05-18 à 14 19 40

### DIFF
--- a/umap/static/umap/js/umap.forms.js
+++ b/umap/static/umap/js/umap.forms.js
@@ -801,6 +801,23 @@ L.FormBuilder.Range = L.FormBuilder.Input.extend({
   value: function () {
     return L.DomUtil.hasClass(this.wrapper, 'undefined') ? undefined : this.input.value
   },
+
+  buildHelpText: function () {
+    var datalist = L.DomUtil.create(
+      'datalist',
+      'umap-field-datalist',
+      this.getHelpTextParent()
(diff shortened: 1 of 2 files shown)


umap/static/umap/js/umap.forms.js
```diff
@@ -801,6 +801,23 @@ L.FormBuilder.Range = L.FormBuilder.Input.extend({
   value: function () {
     return L.DomUtil.hasClass(this.wrapper, 'undefined') ? undefined : this.input.value
   },
+
+  buildHelpText: function () {
```
davidbgk, line 805: I hacked the help text function, I missed some extra helpers to be able to inject some HTML where I want to.

davidbgk:
Heavily inspired by https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range#adding_labels

Issue #713, opened 2019-10-02
better explain impact of anonymous editing


TZornG:
Every other week or so some poor soul ends up at help.openstreetmap.org and asks how they can recover their map after losing the secret editing link ([example 1](https://help.openstreetmap.org/questions/69710/umap-openstreetmap-change-map), [example 2](https://help.openstreetmap.org/questions/70860/edit-changer-osm-map), [example 3](https://help.openstreetmap.org/questions/70832/map-recovery-from-another-account), [example 4](https://help.openstreetmap.org/questions/71001/how-to-regain-access-to-the-umap-i-created)). Apparently, users create a private map not being logged in but are unaware that they then only can edit the map with help of the secret link. 

Could you please add some more visible warning explaining the different options they have creating the map and making it clear that the secret link needs to be saved.

Alternatively, provide some means to recover a map.

It would also be a good idea to add some FAQ dealing with this topic on the umap web sites.

Binnette:
Hi @TZornG, here is a possible way to solve this kind of issue.
https://wiki.openstreetmap.org/wiki/UMap/FAQ#Why_can.27t_I_edit_my_map.3F

I agree you that we should advice users to connect to uMap before creating a uMap project. (I get in trhis trap too).

Binnette:
May be we can add a modal with buttons and the following text:
____
**You are not logged in uMap. Do you want to continue?** 

If you choose to continue be sure to save the administration link of your map in a safe place. Without this link you will lost the ownership of your map.

Cancel - Continue 
____

I think we should put this modal on the main screen (not the map screen). This way, if the user click Cancel, he can directly click on the login button.

@yohanboniface: i can implement it if you want. But i saw that uMap do not have any modal yet. Should I implement an homemade modal, or should i use a CSS framework? If yes, we have to choose a responsive framework for the modal to display correctly on small devices.

gislars:
How can an administrator access the "admin link" for a anonymous map.  I only found the option to set a owner and/or editor to a previously created map.

Until now I only  can offer to export all map data or to set a owner.

davidbgk:
I find it confusing that you get a first alert `The zoom and center have been set.` and then it's updated with the response from the server `Your map has been created! …`. I'll start with a PR for that particular case.

scaidermern:
🤦‍♂️🙈

Issue #420, opened 2016-12-13
Saving current center/zoom creates mistakable success message


ltog:
When saving the current center and zoom level using the icon

![umap_icon_save_center_and_zoom](https://cloud.githubusercontent.com/assets/4726277/21129867/03b91412-c105-11e6-8aae-47c6ea95a2d5.png)

, uMap reports with a mistakable error message:

In english:

> The zoom and center have been setted.

In french:

> Le zoom et le centre ont été enregistrés.

For some readers this might imply that the changed settings are already in effect, but my tests show that this is only the case after the settings have been saved using the save button:
 
![umap_buttons_abbrechen_speichern](https://cloud.githubusercontent.com/assets/4726277/21129905/42160a30-c105-11e6-9c2e-83234a7cdcc6.png)

I propose to update the text to say something like:

> The zoom and center have been registered. Use the "Save" button to apply this change to your map.

yohanboniface:
What about just `The zoom and center as been modified` ?

PR #1092, opened 2023-05-19
Do not display an alert if the map is not created

davidbgk:
Refs #713, otherwise that first message is displayed and then override by the map creation one.


davidbgk:
See also https://github.com/umap-project/umap/issues/420 maybe we should remove that message completely 🤔

yohanboniface:
Merged! I think it's still better than what we have now. Let's discuss #420 there.

PR #1093, opened 2023-05-19
Set `font-display: swap` and reduce fonts’ sizes

davidbgk:
Using https://www.fontsquirrel.com/tools/webfont-generator with subsets: Basic Latin, Latin-1 Sup and Latin Extended-A

Capture d’écran, le 2023-05-19 à 13 42 05


yohanboniface:
Humm, but what about other alphabets (arabic, russian…) ?

davidbgk:
If I keep all the glyphs, we are at 400Ko of fonts for everybody which is suboptimal.

Maybe we should generate a font file for western languages and then a new one for other languages using the [unicode-range](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range) property?

There is a [list of unicode blocks](https://en.wikipedia.org/wiki/Unicode_block#List_of_blocks) on Wikipedia that might help.

yohanboniface:
> If I keep all the glyphs, we are at 400Ko of fonts for everybody which is suboptimal.

Isn't this file supposed to be cached ?

I mean, how big is the issue ? And how much time do you want to spend on fixing it ?

davidbgk:
Closing for now in favor of https://github.com/umap-project/umap/pull/1155

Issue #869, opened 2021-01-18
make link target configurable


dieterdreist:
Currently, all links you set in popups get the `target="_blank"` attribute, which will open a new tab/page and "break" the browser history experience. It would be desirable if one could choose to set the target to "_self" / omit it (optionally).

yohanboniface:
cf #323

PR #1094, opened 2023-05-19
Use DOMPurify to escape malicious input from user

davidbgk:
There are a lot of [complex](https://cheatsheetseries.owasp.org/cheatsheets/XSS_Filter_Evasion_Cheat_Sheet.html) [potential](https://portswigger.net/web-security/cross-site-scripting/cheat-sheet) use cases that were not covered.

It adds an extra 7.8Kb once minified+gzipped.


yohanboniface:
It strips the `target` attribute : 

image

But fix does not seem that simple: https://github.com/cure53/DOMPurify/issues/317

(Keeping in mind that we'll have other value than `_blank` when we fix #869 )

davidbgk:
The case for `target` is fixed, see [this commit](https://github.com/umap-project/umap/pull/1094/commits/16c7e0b239a06ed69d9e174bc6da0b5768c97512) for details (I forgot to apply PrettierJS on tests, will do once this PR is merged!).

yohanboniface (approved):

PR #1100, opened 2023-05-22
Add a very basic `/stats/` JSON view

davidbgk:
Will be useful to feed munin for instance.


davidbgk:
We might want to be inspired by how Mastodon displays stats per instance: https://mastodon.social/api/v1/instance

yohanboniface (approved):

PR #1101, opened 2023-05-23
Improving contrasts on the homepage #a11y

davidbgk:
While working on another issue I realized links are lacking some contrasts and affordances. I tried to improve that with more vibrant colors and arrows.


davidbgk:
Closed in favour of https://github.com/umap-project/umap/pull/1406

PR #1102, opened 2023-05-23
Add a button to send edit link by email in anonymous mode

yohanboniface:
This is a work I started last year, and not yet finished.

The goal is to add a button "Send me the link" when creating a map in anonymous mode.

### DIFF
--- a/umap/settings/base.py
+++ b/umap/settings/base.py
@@ -114,6 +114,8 @@ INSTALLED_APPS = (
 )
 DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
 
+EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
+
 # =============================================================================
 # Calculation of directories relative to the project module location
 # =============================================================================
--- a/umap/settings/base.py
+++ b/umap/settings/base.py
@@ -115,6 +115,7 @@ INSTALLED_APPS = (
 DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
 
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
+FROM_EMAIL = None
 
 # =============================================================================
 # Calculation of directories relative to the project module location
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -1336,11 +1336,23 @@ L.U.Map.include({
       data: formData,
       context: this,
       callback: function (data) {
-        let duration = 3000
+        let duration = 3000,
+          alert = { content: L._('Map has been saved!'), level: 'info' }
         if (!this.options.umap_id) {
-          duration = 100000 // we want a longer message at map creation (TODO UGLY)
+          alert.duration = 100000 // we want a longer message at map creation (TODO UGLY)
           this.options.umap_id = data.id
           this.permissions.setOptions(data.permissions)
+          if (data.permissions && data.permissions.anonymous_edit_url) {
+            alert.actions = [
+              {
+                label: L._('Send me edit link by email'),
+                callback: function () {
+                  this.sendEditLink()
+                },
+                callbackContext: this,
+              },
+            ]
+          }
         } else if (!this.permissions.isDirty) {
           // Do not override local changes to permissions,
           // but update in case some other editors changed them in the meantime.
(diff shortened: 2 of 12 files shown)


umap/static/umap/js/umap.js
```diff
@@ -1336,11 +1336,23 @@ L.U.Map.include({
...
         if (!this.options.umap_id) {
-          duration = 100000 // we want a longer message at map creation (TODO UGLY)
+          alert.duration = 100000 // we want a longer message at map creation (TODO UGLY)
           this.options.umap_id = data.id
           this.permissions.setOptions(data.permissions)
+          if (data.permissions && data.permissions.anonymous_edit_url) {
+            alert.actions = [
+              {
+                label: L._('Send me edit link by email'),
+                callback: function () {
```
davidbgk, line 1349:
```suggestion
                callback: this.sendEditLink
```

Might work?

umap/static/umap/js/umap.js
```diff
@@ -1350,18 +1362,23 @@ L.U.Map.include({
...
+          this.ui.alert(alert)
         })
         this.ui.closePanel()
         this.permissions.save()
       },
     })
   },
 
+  sendEditLink: function () {
+    var url = L.Util.template(this.options.urls.map_send_edit_link, {
```
davidbgk, line 1377: `const` 😛

umap/views.py
```diff
@@ -622,6 +622,28 @@ def post(self, *args, **kwargs):
...
+    model = Map
+    pk_url_kwarg = 'map_id'
+
+    def form_valid(self, form):
+        if (self.object.owner
+           or not self.object.is_anonymous_owner(self.request)
+           or not self.object.can_edit(self.request.user, self.request)):
+            return HttpResponseForbidden()
+        email = form.cleaned_data["email"]
+        from django.core.mail import send_mail
```
davidbgk, line 635: Does `send_mail` import requires external dependencies/settings? If so, maybe add a comment?

yohanboniface:
We should have a working version, now let's make it clean :p

image

umap/views.py
yohanboniface: Yes, settings. I'll add an entry in the documentation.

umap/static/umap/js/umap.js
yohanboniface: Yes! Done :)

davidbgk:
Is that possible to make the link clickable? Last time I tried I failed (because it was not on the top layer?)

yohanboniface:
If you click it, you'll lose it :p

It will open a page that will add a cookie (which is already present in this case, given the map has just been created), then redirect to the map with the normal URL.

davidbgk:
Maybe a `Copy link` button close to it then?

yohanboniface:
image

Simpler than expected!

But it will work only in https, right ?

davidbgk:
> But it will work only in https, right ?

I don't think so, did you use the [Navigator.clipboard](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard) API? (e.g. https://code.larlet.fr/javascript/#copier-dans-le-presse-papier)

yohanboniface:
> I don't think so, did you use the [Navigator.clipboard](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/clipboard) API

Yep, see last commit :)

umap/static/umap/js/umap.ui.js
```diff
@@ -108,26 +107,33 @@ L.U.UI = L.Evented.extend({
...
         if (action.callback)
-          L.DomEvent.on(
-            el,
-            'click',
-            action.callback,
-            action.callbackContext || this.map
-          )
+          L.DomEvent.on(el, 'click', () => {
+            action.callback.bind(action.callbackContext || this.map)()
+            close()
```
yohanboniface, line 127: This `close` call does not work :/

umap/settings/base.py
```diff
@@ -114,6 +114,9 @@
 )
 DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
 
+EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
+FROM_EMAIL = None
```
davidbgk, line 118: I still wonder if we should add `FROM_EMAIL = "test@example.com"` to `local.py.sample` to see the feature in action when developing or if that's really an on-demand one 🤔

umap/static/umap/js/umap.js
```diff
@@ -1336,11 +1336,40 @@ L.U.Map.include({
...
+              {
+                label: L._('Send me the link'),
+                input: L._('Email'),
+                callback: this.sendEditLink,
+                callbackContext: this,
+              },
+              {
+                label: L._('Copy link'),
+                callback: () => {
+                  navigator.clipboard.writeText(data.permissions.anonymous_edit_url)
```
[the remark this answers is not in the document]
yohanboniface: Or is that the sign that we should force https ?

umap/views.py
```diff
@@ -622,6 +611,36 @@ def post(self, *args, **kwargs):
...
+            return HttpResponseForbidden()
+        form = self.get_form()
+        if form.is_valid():
+            email = form.cleaned_data["email"]
+        else:
+            return HttpResponseBadRequest("Invalid")
+        link = self.object.get_anonymous_edit_url()
+
+        send_mail(
+            _("Your secret edit link"),
```
davidbgk, line 633:
```suggestion
            _("The uMap edit link for your map: %(map_name)s" % {"map_name": self.object.name}),
```

This original subject looks spam-y to me :)

davidbgk (approved):

PR #1103, opened 2023-05-25
📝 — Fix X-Accel-Redirect doc

ldidry:
Since 1.2.5, you don’t need `gzip` instructions for `location /internal` anymore. In fact, it breaks the calls for datalayers!

I think it comes from 8294f896c4ccd6efff642e6f615f07ec19e49ed4, but I can’t be sure.

Anyway, removing `gzip` instructions for `location /internal` fixes our instance.

### DIFF
--- a/docs/ubuntu.md
+++ b/docs/ubuntu.md
@@ -328,8 +328,6 @@ And then add this new location in your nginx config (before the `/` location):
 
     location /internal/ {
         internal;
-        gzip_vary on;
-        gzip_static on;
         alias /path/to/umap/var/data/;
     }
 


yohanboniface:
Hey Luc :)

That's weird! Here is what we have in OSM Fr servers (which also run the 1.2.x):

```
# X-Accel-Redirect
location /internal/ {
    internal;
    gzip_vary on;
    gzip_static on;
    alias /srv/umap/media_root/;
}
```

It's true that we changed the gzip management since 1.2.5 version, but that documentation should be up to date with this version.

Do you have `UMAP_XSENDFILE_HEADER = 'X-Accel-Redirect'` in your local settings ?

ldidry:
Yes, I have it.

ldidry:
I found the problem: I had `add_header Content-Encoding gzip;` in `location /internal/`

ldidry:
My fault, sorry.

yohanboniface:
:+1: 

@ldidry We are doing a bunch of work in uMap, would you be available for a short visio to present Framacarte needs ?

PR #1108, opened 2023-05-26
Fix JS tests

davidbgk:
Since this commit https://github.com/umap-project/umap/commit/97e2df0a8d698d1a9a531447db360187062a4843 tests were broken because the URL has an extra date parameter to update the cache for datalayers. With the new regexp, we match these new URLs with the fakeServer from sinon.

Also update sinon to v15.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -42,7 +42,7 @@
         "optimist": "~0.4.0",
         "phantomjs": "^1.9.18",
         "prettier": "^2.8.8",
-        "sinon": "^1.10.3",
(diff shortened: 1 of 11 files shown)


davidbgk:
At the moment, if you reach the test page the first time, you'll have 3 tests failures but if you hard refresh that same page, you'll be down to 0 😅

yohanboniface:
> At the moment, if you reach the test page the first time, you'll have 3 tests failures but if you hard refresh that same page, you'll be down to 0

Javascript…

PR #1121, opened 2023-06-05
Docker part 3

yohanboniface:
Following #801 

cc @Duvel

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -1,0 +1,33 @@
+# Docker
+
+There is now an official [uMap](https://hub.docker.com/r/umap/umap) image.
+
+To run it with docker compose, use a `docker-compose.yml` like this:
+
+```yaml
+version: '3'
+services:
+  db:
+    image: postgis/postgis:14-3.3-alpine
+    environment:
+      - POSTGRES_HOST_AUTH_METHOD=trust
+    volumes:
+      - db:/var/lib/postgresql/data
+
+  app:
+    image: umap/umap:x.x.x
+    ports:
+      - "8001:8000"
+    environment:
+      - DATABASE_URL=postgis://postgres@db/postgres
+      - SECRET_KEY=some-long-and-weirdly-unrandom-secret-key
+      - SITE_URL=https://umap.local/
+      - UMAP_ALLOW_ANONYMOUS=True
+      - DEBUG=1
+    volumes:
+      - data:/srv/umap/uploads
+
+volumes:
+  data:
+  db:
+```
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,65 @@
+FROM node:alpine AS vendors
+
+COPY . /srv/umap
```
jpetazzo, line 3: Suggestion: 
```
WORKDIR /srv/umap
COPY package.json . # or whatever is used to defined dependencies
RUN npm install
COPY . .
RUN npm run vendors
```

Dockerfile
```diff
@@ -0,0 +1,65 @@
...
+
+RUN mkdir -p /srv/umap/uploads
+
+COPY . /srv/umap
+
+COPY --from=vendors /srv/umap/umap/static/umap/vendors /srv/umap/umap/static/umap/vendors
+
+WORKDIR /srv/umap
+
+RUN apt-get update && \
```
jpetazzo, line 24: Suggestion: move that whole `RUN apt-get...` block so that it's just after the `FROM` line; that way it won't be executed again every single time a file is changed

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
+version: '3'
+services:
+  db:
+    image: postgis/postgis:14-3.3-alpine
```
jpetazzo, line 4: I'd suggest to check if there is a looser pinning here; maybe postgis/postgis:14-alpine - unless you specifically depend on 14-3.3 ?

docker/entrypoint.sh
```diff
@@ -0,0 +1,29 @@
...
+    tries=$((tries + 1))
+  done
+)}
+
+# first wait for the database
+wait_for_database
+# then migrate the database
+umap migrate
+# then collect static files
+umap collectstatic --noinput
```
jpetazzo, line 25: nit: if static files have already been collected, is this a no-op?

docker-compose.yml
yohanboniface: Does not seem to be https://hub.docker.com/r/postgis/postgis :(

Dockerfile
yohanboniface: Should be good, thanks :)

docker/entrypoint.sh
yohanboniface: Yes, Django will compare collected files with the source and just display a message if everything is up to date.

```
$ umap collectstatic --no-input
Loaded local config from /home/ybon/Code/py/umap/umap/settings/local.py

114 static files copied to '/var/local/umap/static', 241 unmodified.
$ umap collectstatic --no-input
Loaded local config from /home/ybon/Code/py/umap/umap/settings/local.py

0 static files copied to '/var/local/umap/static', 355 unmodified.
```

PR #1124, opened 2023-06-05
Add Liberapay funding

yohanboniface:
Hopefully, that's the way to go :p


davidbgk (approved):
Maybe put your own Patreon too?

yohanboniface:
We have one yet ?

davidbgk:
I thought it was personal but we have to/can create a uMap account, let's do that tomorrow 👍

PR #1136, opened 2023-06-10
Add experimental choropleth datalayer type

yohanboniface:
image

TODO:

- [ ] tests (at least the color computation)
- [ ] automatic legend with range colors and steps values
- [ ] control the lower color (currently white)
- [ ] redraw all features when a feature is modified/added (as each feature color depends on the whole set)

Note: I've not use Leaflet.Choropleth plugin for two reasons: it does not implement `L.Layer` (which we rely on), it will only set `fillColor`, while I'd like our choropleth mode to also affect point or lines, just to keep things generic.


yohanboniface:
First naive version of an automatic legend:

image

yohanboniface:
Works with points too:

image

davidbgk:
`[choropleth_region_chomage.geojson](https://github.com/umap-project/umap/pull/1136/files#diff-a50888c5d9951944ee56acbe7ba18aedc228f686e824e0b776c7f8c2b2301d7a)` is quite big, don't we have a more minimalist example?

yohanboniface:
> [choropleth_region_chomage.geojson](https://github.com/umap-project/umap/pull/1136/files#diff-a50888c5d9951944ee56acbe7ba18aedc228f686e824e0b776c7f8c2b2301d7a) is quite big, don't we have a more minimalist example?

Good point! I've simplified it! It can be even simpler if you think it should.

davidbgk:
> Good point! I've simplified it! It can be even simpler if you think it should.

Way better, we lost 50kLOC in the process 😅

PR #1141, opened 2023-06-14
Allow `geo:` scheme in (description) links

### LINKED ISSUE #1140: [[geo:{lat},{lon}|Directions]] no longer works
**Describe the bug**
[[geo:{lat},{lon}|Directions]] when used in a Popup content template used to return a link to: `<a href="geo:<lat>,<lon>">Directions</a>`.
Now only `<a>Directions</a>` is returned.

**To Reproduce**
1. Create Popup content template
2. Add [[geo:{lat},{lon}|Directions]]
3. View item popup

**Expected behavior**
Link using html geo schema 

 - OS: Ubuntu
 - Browser: Firefox/Chrome
 - Version: 113.0.2

davidbgk:
Fix #1140

### DIFF
--- a/umap/static/umap/js/umap.core.js
+++ b/umap/static/umap/js/umap.core.js
@@ -64,6 +64,12 @@ L.Util.escapeHTML = (s) => {
     ],
     ADD_ATTR: ['target', 'allow', 'allowfullscreen', 'frameborder', 'scrolling'],
     ALLOWED_ATTR: ['href', 'src', 'width', 'height'],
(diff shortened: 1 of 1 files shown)


yohanboniface:
Maybe we should also create a PR to dompurify ?

davidbgk:
There are so many existing schemes, I understand that it's complicated to support all and defaulting to popular ones is a decent compromise. Allowing to set your own regexp is already enough.

Issue #1135, opened 2023-06-10
Enhance "datalayers switcher"


yohanboniface:
Here is how it looks currently:

![image](https://github.com/umap-project/umap/assets/146023/600d888d-0121-4bf5-8565-b9c4ce921998)

A few feedback:

- it should include colors and icons of each datalayer
- it should be in the bottom of the screen, to be less intrusive (maybe merge with new "permanent credit")
-

yohanboniface:
A few more questions/remarks:

- should the users be able to chose the position (top/bottom left/right) ?
- should we include caption text when provided ? Otherwise, we may need to work on a "small caption box", because the current one in the full panel is quite intrusive
- I'm in the process to allow choosing which layers would go in this box: some layer be only be for esthetics reasons, or be unfinished…
- should we allow to display some features in the legend ? either because they represent some sort of features one may want to showcase ("those features are the hospitals"), either because as a feature it should be show cased ("this feature is the finish of the race")

almet:
See #1469

yohanboniface:
Done in #1687 and following PRs

entrider:
> should the users be able to chose the position (top/bottom left/right) ?

It would be great! I still can't get used to it being on the left.

PR #1143, opened 2023-06-15
Ability to resize the list of layers displayed on the map

pfbreton-perso:
This is a proposed change to resize the list of layers to fit the content of the list by default and give the ability to resize the window.

I felt that I was constrained by the fixed size that wasn't displaying enough layers to be comfortable to use.

layer_resize


yohanboniface:
Thanks!

Related #1135

PR #1144, opened 2023-06-16
Show messages also in the content part

yohanboniface:
Messages are already shown in the map, if any.

Useful in our case when an error is raised by the auth flow.

Screenshot from 2023-06-15 08-30-06


davidbgk (approved):
At some point, we might want to use CSS variables for recurring colours 👍

yohanboniface:
> At some point, we might want to use CSS variables for recurring colours +1

Yep! :)

PR #1148, opened 2023-06-16
Overflow popup

yohanboniface:
Before:

image

After:

image

### DIFF
--- a/umap/static/umap/js/umap.popup.js
+++ b/umap/static/umap/js/umap.popup.js
@@ -116,12 +116,7 @@ L.U.PopupTemplate.Default = L.Class.extend({
 
   renderFooter: function () {
     if (this.feature.hasPopupFooter()) {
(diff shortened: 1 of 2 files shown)


umap/static/umap/js/umap.popup.js
```diff
@@ -116,12 +116,7 @@ L.U.PopupTemplate.Default = L.Class.extend({
 
   renderFooter: function () {
     if (this.feature.hasPopupFooter()) {
-      const footerContainer = L.DomUtil.create(
-          'div',
-          'umap-footer-container',
```
davidbgk, line 121: You need to remove that class from base.css too

davidbgk (approved):

umap/static/umap/js/umap.popup.js
yohanboniface: :+1:

Issue #1137, opened 2023-06-14
Basic Public API for stats


virgile-dev:
**Is your feature request related to a problem? Please describe.**
Was wondering if umap had some kind of public api that can expose some stats like number of maps, number of users etc.

**Describe the solution you'd like**
Here is an example of what we had with decidim, more advanced than what I'm describing but allowed a member of the community to make a dashboard with all known live instances and their basic stats : https://decidim-census.digidemlab.org/
<img width="951" alt="Capture d’écran 2023-06-14 à 09 37 16" src="https://github.com/umap-project/umap/assets/11473995/cb27a942-204e-4768-ba63-96564fe00bc7">

**Describe alternatives you've considered**
Every instances send regularly their data to be agregated by the mainteners.

**Additional context**
Would help measuring the impact of the project.

yohanboniface:
Very lite, but it exists since recently!

https://umap.openstreetmap.fr/stats/

Could be improved if you have other needs :)

virgile-dev:
I do have one, could you expose the version of umap that is deployed on the server. This is will give us an idea of how much up to date an instance is with the last version.

PR #1153, opened 2023-06-20
Expose version in /stats/ endpoint

yohanboniface:
cf #1137

### DIFF
--- a/umap/tests/test_views.py
+++ b/umap/tests/test_views.py
@@ -8,6 +8,7 @@ from django.contrib.auth import get_user, get_user_model
 from django.urls import reverse
 from django.test import RequestFactory
 
+from umap import VERSION
 from umap.views import validate_url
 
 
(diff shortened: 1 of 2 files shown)


umap/tests/test_views.py
```diff
@@ -157,6 +158,7 @@ def test_stats_empty(client):
         "maps_count": 0,
         "users_active_last_week_count": 0,
         "users_count": 0,
+        "version": VERSION,
```
davidbgk, line 161: At least we are testing the existence of the key 😅

davidbgk (approved):

umap/tests/test_views.py
yohanboniface: How would you test that without needing to update the value at each version otherwise ? :)

davidbgk: You're right, monkeypatching would be a lie anyway 👍

PR #1154, opened 2023-06-20
Linting and prettifying templates with djLint for consistency

davidbgk:


Makefile
```diff
@@ -9,6 +9,14 @@ install: ## Install the dependencies
 develop: ## Install the test and dev dependencies
 	python3 -m pip install -e .[test,dev]
 
+.PHONY: pretty-templates
+pretty-templates: ## Prettify template files
+	djlint umap/templates --reformat
+
+.PHONY: lint-templates
```
yohanboniface, line 16: AFAIK, you don't need to add .PHONY if the command name does not match with a local path (like `build/`)

umap/templates/500.html
```diff
@@ -1,8 +1,11 @@
 <!DOCTYPE html>
-<html>
-<head>
+<html lang="en">
+  <head>
     <title>Ooops</title>
     <meta charset="utf-8">
+    <meta name="description"
+          content="uMap lets you create maps with OpenStreetMap layers in a minute and embed them in your site.">
```
yohanboniface, line 7: Shouldn't this be translated ? I think we already have a string like this in transifex.

umap/templates/500.html
```diff
@@ -1,8 +1,11 @@
 <!DOCTYPE html>
-<html>
-<head>
+<html lang="en">
```
yohanboniface, line 2: Shouldn't this be set according to current LANGUAGE_CODE ?

yohanboniface (approved):
:+1:

Issue #930, opened 2021-07-22
feature request: add option to hide more from user interface


maphouse:
Under "update permissions and editors", map creators are given the ability to choose how they want their map to be modified by collaborators or the public. However, the options available to any anonymous public user are too powerful for most use-cases: most map creators who make their maps public may want the public's contributions (i.e. creating new map markers), but don't want the public to be able to remove existing markers, modify data, edit a map's user interface or even to update user permissions. To resolve this, it would be ideal if within the "user interface options", a map creator was able to hide all tabs on the right except for the geometry creation tabs (see highlighted in img below).
![image](https://user-images.githubusercontent.com/4475249/126718733-13c5df9d-8653-4c3f-b120-647b3d446e29.png)

Looking forward to feedback on this!

yohanboniface:
Makes sense!

See also #584

yohanboniface:
This should be a side effet of new permissions at datalayer level: if one user has write permissions on a datalayer but not on the map itself, they will not see those advanced buttons (in yellow in your screenshot).

So I'm closing as I think it answers the needs. Please reopen or comment if you think I'm missing something! :)

PR #1159, opened 2023-06-20
Allow to hide advanced edit buttons

yohanboniface:
This only deals with showing or hidding those buttons, but does not deal with real persmission behind (which are handled by the back-end).

There is at least one workaround possible for users (apart from using the JS console): clicking the "pencil" near to the map name in the edit bar will open the map settings. Not sure how much we want to block this.

Also we may want to be more granular: at this stage, if unclicked, it will hide the buttons for everyone but the owner, we may for example show buttons for editors but not anonymous…

cf #930

image

### DIFF
--- a/umap/static/umap/js/umap.controls.js
+++ b/umap/static/umap/js/umap.controls.js
@@ -272,8 +272,13 @@ L.U.ContinueLineAction = L.U.BaseVertexAction.extend({
   },
 })
 
-// Leaflet.Toolbar doesn't allow twice same toolbar class…
-L.U.SettingsToolbar = L.Toolbar.Control.extend({})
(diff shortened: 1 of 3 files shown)


davidbgk:
I wonder if this simple/advanced interface switch should be the responsibility of the owner or the user 🤔

yohanboniface:
Maybe both ?

image

And then we could imagine the same "More" button when the setting is set to `hidden` ?

davidbgk:
> And then we could imagine the same "More" button when the setting is set to `hidden` ?

Yes, it would be consistent with the left side panel.

yohanboniface:
Replaced by #1307

Issue #858, opened 2020-12-08
Assing pin color through table parameter (column)


guaspito:
When using tables and remote csv sources, in umap it is possible to define name, description and pin icon selecting the table column/field: {name}, {description}, {icon_url}. 

It would be nice if were possibile to select the color of pins of a layer defining an appropriate table field (containing color names or codes) ad recalling it in the map configuration as we do for name, description and pin_icon (see image)

![umap_color_field](https://user-images.githubusercontent.com/5469846/101434536-023e6880-390b-11eb-9315-a6fac1d762e4.png)

This feature allows to have pins of different colors belonging to the same layer, as requested here #818

yohanboniface:
Will be in next release.

PR #1160, opened 2023-06-21
Allow to use properties as color value

yohanboniface:
cf #858

I've created a new function `getSmartOption`, that extends `getOption`, so to use it explicitly for now. We may at some point make that `getOption` is always smart, but let's be a bit conservative here…

### DIFF
--- a/umap/static/umap/js/umap.features.js
+++ b/umap/static/umap/js/umap.features.js
@@ -109,7 +109,7 @@ L.U.FeatureMixin = {
     properties.unshift('properties.name')
     builder = new L.U.FormBuilder(this, properties, {
       id: 'umap-feature-properties',
-      callback: this.resetTooltip,
+      callback: this._redraw, // In case we have smart options…
     })
     container.appendChild(builder.build())
     this.appendEditFieldsets(container)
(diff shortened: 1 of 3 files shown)


umap/static/umap/js/umap.features.js
```diff
@@ -271,6 +271,18 @@ L.U.FeatureMixin = {
     return value
   },
 
+  getSmartOption: function (option, fallback) {
```
davidbgk, line 274: Not really fond of the `Smart` name, maybe `Dynamic`?

yohanboniface: Yep, smart is not smart choice :p

I've made the change :)

umap/static/umap/js/umap.features.js
```diff
@@ -109,7 +109,7 @@ L.U.FeatureMixin = {
     properties.unshift('properties.name')
     builder = new L.U.FormBuilder(this, properties, {
       id: 'umap-feature-properties',
-      callback: this.resetTooltip,
+      callback: this._redraw, // In case we have smart options…
```
davidbgk, line 112:
```suggestion
      callback: this._redraw, // In case we have dynamic options…
```

😉

davidbgk (approved):

Issue #323, opened 2016-05-02
Option to open links in "_parent" target instead of "_blank" for easier iframe integration


nouser2013:
Dear dev's,

for easier iFrame integration I'd like to be able to set the target property of outgoing URLs. On my webpage, I currently have an interactive map with several areas. Upon clicking, a link such as `www.mydomain.com/deep-link#area_1` is opened; the browser scrolls down to the respective anchor.

Unfortunately, it always opens in a new window. I'd like to open in "_parent" to hint to the browser to simply scroll instead of loading the whole page again, even over cross-origin.

Thanks.

yohanboniface:
Would allow anchor values only in the link be an option? In other words:
- if the link starts with http:// => open in a new window
- if the link starts with '#' => open in the same window

It would be a bit implicit, but we already have a bunch of options in uMap, and I'm not sure this needs a new one.

yohanboniface:
Of maybe `link domain == current domain` => same window?

nouser2013:
> Would allow anchor values only in the link be an option? In other words: [...]

Well, functionality-wise, yes, of course. You could also directly parse the `target=""` value from the textbox like so:

* _parent:http(s)://...
* _blank:http(s)://
* _self:http(s)://

> Or maybe link domain == current domain => same window?

Nope, due to x-origin restriction you have no way of determining where the map is viewed: you cannot say if it's stand-alone in a separate tab/window or within an iframe on a parent page. Your condition would likely never match since you probably don't want to link to uMap in most cases.

yohanboniface:
> Nope, due to x-origin restriction you have no way of determining where the map is viewed

Good point!

> You could also directly parse the target="" value from the textbox

I'll think about that option.

nouser2013:
Alright, thanks. I'd be very nice to have it, here's the (German) web page I'd like to use it in: http://www.pro-fuerstenried.de/2016/05/oekologie-im-viertel-ein-gastbeitrag/

Click on a green area on the map and jump down to the appropriate detailed description...

tourenfex:
Is there an solution for this problem?

nouser2013:
@yohanboniface with regard to your other comment in #349 (bottom):

> There is not way to change that behaviour for now sadly. I've looked at it, and it's not that easy given the action is made in javascript (it's not a real HTML link), and thus there are restriction for an iframe to communicate with parent.

[...]

PR #1162, opened 2023-06-21
Control links target

### LINKED ISSUE #869: make link target configurable

yohanboniface:
This PR does two things:

- make the outlink available for any feature, not only polygon: which is, one can define a link that will be opened when clicking on the feature (instead of opening a popup) cf #323 
- use the `outlinkTarget` option to also control where popup links will be opened (self/blank/parent) cf #869 

fix #869

### DIFF
--- a/umap/static/umap/js/umap.core.js
+++ b/umap/static/umap/js/umap.core.js
@@ -73,8 +73,9 @@ L.Util.escapeHTML = (s) => {
   })
   return s
 }
-L.Util.toHTML = (r) => {
+L.Util.toHTML = (r, options) => {
   if (!r) return ''
+  const target = options && options.target || 'blank'
   let ii
 
   // detect newline format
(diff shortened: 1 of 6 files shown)


umap/static/umap/js/umap.core.js
```diff
@@ -73,8 +73,9 @@ L.Util.escapeHTML = (s) => {
   })
   return s
 }
-L.Util.toHTML = (r) => {
+L.Util.toHTML = (r, options) => {
   if (!r) return ''
+  const target = options && options.target || 'blank'
```
davidbgk, line 78: Should we restrict the value to the [3 possible options](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#target) only?

yohanboniface: Is that needed ? Given that user does not manually enter those values (they come from a radio input) ?

davidbgk: Alright

davidbgk (approved):

PR #1163, opened 2023-06-21
Pass options at datalayer creation when importing from umap file

### LINKED ISSUE #1042: Last layer of umap import file doesn't show name
My browser is MS Edge.

When I import a umap file with multiple layers and the map has set datalayersControl to expanded, the last layer name is not the name from the file in the datalayersControl. It is still the default layer name, in my case "Ebene 7".
when i edit the layer the correct name is shown in the right sidebar. As soon as I edit the name in the sidebar (e.g. adding a character) the name gets updated immediatly in the datalayersControl.

My current workaround is to add an dummy layer as the last layer to the file. At least then my data layers have the correct names.

Expected behaviour: after import all layers show the name from the umap file.

yohanboniface:
fix #1042

Until now, we create an empty datalayer (which has some defaults), then we overwrite the options while calling fromGeoJSON method

### DIFF
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -1038,7 +1038,7 @@ L.U.Map.include({
     if (importedData.geometry) this.options.center = this.latLng(importedData.geometry)
     const self = this
     importedData.layers.forEach((geojson) => {
-      const dataLayer = self.createDataLayer()
+      const dataLayer = self.createDataLayer(geojson._umap_options)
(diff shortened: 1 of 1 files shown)


davidbgk (approved):
I'm not sure I get all the intricacies of this but it's OK.

yohanboniface:
Me neither :p

But I suppose it's safe because the `_umap_options` will be passed anyway at the next step.

PR #1168, opened 2023-06-23
Be more strict when coordinates are set manually

yohanboniface:
cf https://forum.openstreetmap.fr/t/impossible-de-charger-la-carte-carte-acteurs-naturels-batiment/15607/7

### DIFF
--- a/umap/static/umap/js/umap.features.js
+++ b/umap/static/umap/js/umap.features.js
@@ -606,17 +606,22 @@ L.U.Marker = L.Marker.extend({
 
   appendEditFieldsets: function (container) {
     L.U.FeatureMixin.appendEditFieldsets.call(this, container)
+    const latbk = this._latlng.lat,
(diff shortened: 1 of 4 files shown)


umap/static/umap/js/umap.features.js
```diff
@@ -606,17 +606,22 @@ L.U.Marker = L.Marker.extend({
 
   appendEditFieldsets: function (container) {
     L.U.FeatureMixin.appendEditFieldsets.call(this, container)
+    const latbk = this._latlng.lat,
```
davidbgk, line 609: I can't see where you reuse these vars 🤔

yohanboniface: Nowhere, that's coming from a first attempt :p

Thanks for spotting! :)

davidbgk (approved):

PR #1170, opened 2023-06-23
More natural sort of features

### LINKED ISSUE #353: [wishlist] better sorting
Data browser and Caption are sorting elements in a poor way. For instance, elements named :

1
9
10
11
100

are actually sorted :

100
10
11
1
9

I know it seems easy to overcome by adding zeroes. But it can be a pain when dealing with large numbers or remote data you may not have full control on. This would be a nice improvement of the default behavior.

yohanboniface:
fix #353

+ a bit of prettier noise, sorry


davidbgk (approved):
See, batteries are included sometimes 🔋

Issue #698, opened 2019-05-26
Introduce feature ordering


unpack5:
It would be useful if the visual order of features (markers, lines, polylines, etc.) could be manually modified. For example, right now I have a polyline that eclipses another and that causes issues in its visibility and how easy it is to edit it.

Could we have drag-and-drop functionality in "Browse data" to change the order of features within a layer?

Binnette:
I don't think it is possible yet. Please consider using multiple layers.

yohanboniface:
There is no such feature, but in the meantime, one can use the "sort key" to control ordering.

![image](https://github.com/umap-project/umap/assets/146023/c75c5733-ff80-4576-84cf-83b90a99acdf)

picomiles:
Hi @yohanboniface :)
it is possible add more than one key in the Sort key field? and an order (ascending, descending)?
if it is possible, can you explain how. 
Thanks

picomiles:
Thank you so much @yohanboniface :)

PR #1173, opened 2023-06-26
Allow to sort reverse

yohanboniface:
cf #698


davidbgk (approved):
Maybe we should add something in the help (?) message too in order to document the feature?

yohanboniface:
> Maybe we should add something in the help (?) message too in order to document the feature?

Indeed! Pushed something :)

PR #1174, opened 2023-06-27
Svg icons

yohanboniface:
Before:
image

After:
image


davidbgk:
The new SVG icons are so crisp on my screen 😱 💛 

The little arrow at the bottom of the list is still a bit too narrow compared to the icons, I don't know if it's intentional?

Capture d’écran, le 2023-06-27 à 09 23 06

These icons are not yet SVG and it's noticeable :)

yohanboniface:
Yes, I only made the 24px black on white, so to test the system, and when we're good I'll do the other (24px white on black, and both 16px).

PR #1190, opened 2023-07-03
Use LANGUAGE_CODE instead of locale name for browser APIs

### LINKED ISSUE #1187: Locale "pt-pt" stoped working, need to use just "pt" for map to work
**Describe the bug**

Using URL's like https://umap.openstreetmap.fr/pt-pt/map/XXXXX stoped working correcty, to get the correct results the "pt-pt" needs to be replaced with just "pt"

**To Reproduce**
1. Go to https://umap.openstreetmap.fr/pt-pt/map/antenas-mobile-v21_724717
2. Only the pink layer will show data
3. Go to https://umap.openstreetmap.fr/pt/map/antenas-mobile-v21_724717
4. All layers will show data

**Expected behavior**
The URL with pt-pt should work as it always did

**Desktop:**
 - OS: Windows 10 or Ubuntu 20
 - Browser: Same issue with Firefox, Chrome or Brave
 - Version: any

**Smartphone (please complete the following information):**
 - Device: Any android device
 - OS: 12 or 13
 - Browser: stock or chrome
 - Version: any

**Additional context**

yohanboniface:
We use locale name (pt_PT) instead of the language code to use in localeCompare, which is wrong.

Given that both need to exist, let's also expose it in extendedProperties.

fix #1187

### DIFF
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -65,8 +65,15 @@ L.U.Map.include({
   ],
 
   initialize: function (el, geojson) {
-    if (geojson.properties && geojson.properties.locale)
+    // Locale name (pt_PT, en_US…)
+    // To be used for Django localization
+    if (geojson.properties.locale)
       L.setLocale(geojson.properties.locale)
+
+    // Language code (pt-pt, en-us…)
+    // To be used in javascript APIs
+    if (geojson.properties.lang)
+      L.lang = geojson.properties.lang
 
     // Don't let default autocreation of controls
     const zoomControl =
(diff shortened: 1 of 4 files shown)


umap/static/umap/js/umap.js
```diff
@@ -65,9 +65,16 @@ L.U.Map.include({
   ],
 
   initialize: function (el, geojson) {
-    if (geojson.properties && geojson.properties.locale)
```
davidbgk, line 68: Not sure how critical it is but we are not testing anymore if `geojson` has `properties` first.

[Optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) might be a way to handle the case without overcharging the syntax but it's for relatively recent browsers (2020).

davidbgk (approved):

umap/static/umap/js/umap.js
yohanboniface: Yep, I removed it be cause all the following lines were consuming properties without this check, so at this point…

Issue #1048, opened 2023-02-21
Option for automatic backups of map data


forteller:
Hi. Thank you for this amazingly useful tool! And for your hard work to get everything back after the recent trouble, #1045.

What happened did highlight, though, the need for better tools for backing up our maps individually. The next time this happens, maybe you won't be able to restore some of it, or maybe someone needs the data while you work on restoring it.

There already is a way to download all the data from each map, of course. But you have to remember to do this manually from each map individually.

My feature request is an option to have the data sent to your e-mail automatically after edits. Not after every single time you press Save on any map, but for example on the hour after edits, or something like that. This could be enabled on a pr map basis, or on a pr account basis.

Thank you again!

PR #1196, opened 2023-07-04
Dashboard

### LINKED ISSUE #391: [wishlist] Better distinction between own maps and others in "My maps"
"My maps" barely separates : 
- maps you own (created or cloned by you)
- maps you're allow to edit (because you were invited to)

As for now, you only see the difference when you look at the "by User" mention. 

It would be clearer if there was some sort of visual way to distinguish between the 2 cases. 

Like a different tab for maps that ain't 100% "yours", or a colorful tag to help the distinction.

### LINKED ISSUE #1062: Getting list of one's own maps?
Hello,

Unless I missed it, Umap doesn't seem to provide a way to list someone's maps.

Is there a reliable way, besides this type of (Windows) script?

```@ECHO OFF
REM Know of no other way than trial and error to find how many times to loop
FOR /L %%i IN (1,1,15) DO (
  wget -qcO - http://umap.openstreetmap.fr/en/user/joe/?p=%%i | grep -Poha "/en/map/.+?_\d+" | sed -r "s@^(.+)$@http://umap.openstreetmap.fr/\1@"
)
```

Thank  you.

--

Edit: The script above seems to work. I get the same list when scrolling manually in a browser.

--

Edit: Scrolling seems to only show the last fifty maps: Some of my maps aren't in the list :-/

yohanboniface:
Very basic version. Must land here so we can theme it in DSFR.

image

fix #1062
fix #391 
cf #1048

### DIFF
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -254,6 +254,7 @@ L.U.Map.include({
       if (this.options.onLoadPanel === 'databrowser') this.openBrowser()
       else if (this.options.onLoadPanel === 'caption') this.displayCaption()
       else if (this.options.onLoadPanel === 'datafilters') this.openFilter()
+      else if (this.options.onLoadPanel === 'share') this.renderShareBox()
     })
     this.onceDataLoaded(function () {
       const slug = L.Util.queryString('feature')
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -91,12 +91,8 @@ L.U.Map.include({
 
     this.ui = new L.U.UI(this._container)
     this.xhr = new L.U.Xhr(this.ui)
-    this.xhr.on('dataloding', function (e) {
-      this.fire('dataloding', e)
-    })
-    this.xhr.on('datalaod', function (e) {
-      this.fire('datalaod', e)
-    })
+    this.xhr.on('dataloading',  (e) => this.fire('dataloading', e))
+    this.xhr.on('dataload', (e) => this.fire('dataload', e))
 
     this.initLoader()
     this.name = this.options.name
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -260,6 +260,7 @@ L.U.Map.include({
       const slug = L.Util.queryString('feature')
       if (slug && this.features_index[slug]) this.features_index[slug].view()
       if (L.Util.queryString('edit')) this.enableEdit()
+      if (L.Util.queryString('download')) this.download()
     })
 
     window.onbeforeunload = (e) => {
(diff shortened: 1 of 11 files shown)


umap/static/umap/js/umap.js
```diff
@@ -258,10 +254,13 @@ L.U.Map.include({
       if (this.options.onLoadPanel === 'databrowser') this.openBrowser()
       else if (this.options.onLoadPanel === 'caption') this.displayCaption()
       else if (this.options.onLoadPanel === 'datafilters') this.openFilter()
+      else if (this.options.onLoadPanel === 'share') this.renderShareBox()
```
yohanboniface, line 257: Do we want to use `?share` instead, to be consistent with `?edit` and `?download` ?

umap/templates/umap/map_table.html
```diff
@@ -0,0 +1,29 @@
+{% load umap_tags umap_tags i18n %}
+<table class="maps">
+  <tr>
```
davidbgk, line 3: That `tr` should be in a `thead`

umap/templates/umap/map_table.html
```diff
@@ -0,0 +1,29 @@
...
+    <th></th>
+    <th></th>
+    <th>{% blocktrans %}Status{% endblocktrans %}</th>
+    <th>{% blocktrans %}Last save{% endblocktrans %}</th>
+    <th>{% blocktrans %}Owner{% endblocktrans %}</th>
+    <th>{% blocktrans %}Actions{% endblocktrans %}</th>
+  </tr>
+  {% for map_inst in maps %}
```
davidbgk, line 11: That part should be in a `<tbody>`

davidbgk (commented):
I wonder if we want search results to match that new table style 🤔

umap/templates/umap/map_table.html
yohanboniface: Can you elaborate on why we should add those tags ? I thought that HTML5 thing was to only add needed tags, to KISS :)

yohanboniface:
> I wonder if we want search results to match that new table style thinking

I'd say no, given those pages do not a address the same users/needs (we don't need status nor actions for example in the search results)

umap/templates/umap/map_table.html
yohanboniface: Added `<thead>` and `<tbody>`, but not sure why we'd have a `<caption>`.

Note that all seems optional according to (my understanding of) MDN: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table :)

yohanboniface:
Merging, so I can put it in today's release, and start working on it in DSRF version. Let's iterate to make it better :)

Pilaiv:
thank you for this new dashboard interface. I am new to github. Is here that I can suggest an addition in the "status" column, can you also add the detail "Who can see" in addition to "Who can modify"

yohanboniface:
Makes sense!

Something like this ?

image

PR #1199, opened 2023-07-10
We do not want horizontal overflow/scroll bar in popups

### LINKED ISSUE #1198: CSS .umap-popup-content Scroll Bars
After the latest update there is an issue with the scoll bars. In the former CSS .umap-popup-content there was:
  overflow-y: auto;
  overflow-x: hidden;

Now there is only:
  overflow-y: auto;

I link to html-files and there remains a horizontal scoll bar now which I cannot change (in the pop-up small).

See: https://umap.openstreetmap.fr/nl/map/nederland-routemaps-nl_795244#7/52.510/5.570

Less visible on a smartphone, good visible on the PC in any browser.

image

yohanboniface:
:upside_down_face: 

fix #1198

### DIFF
--- a/umap/static/umap/map.css
+++ b/umap/static/umap/map.css
@@ -1352,6 +1352,7 @@ a.add-datalayer:hover,
     display: flex;
     flex-direction: column;
     overflow-y: auto;
+    overflow-x: hidden;
 }
 .umap-popup-content iframe {
     min-width: 310px;


davidbgk (approved):
🤷 we need a test map for these edge cases

yohanboniface:
>  we need a test map for these edge cases

Let's start it!

https://umap.openstreetmap.fr/fr/map/popopedup_939587

Xonnext:
Are 'Popup' and 'Popup (large)' the same size now as a result of the not showing the horizontal scroll bar? Selecting the 'Popup' shows the same result as selecting the 'Popup (large)'.

PR #1200, opened 2023-07-11
Fix permissions panel empty at first save

yohanboniface:
cf https://forum.openstreetmap.fr/t/apres-enregistrement-dune-carte-les-permissions-ne-sont-pas-presentes/15995


davidbgk (approved):
Not the first time I see that kind of permutations 🤔

yohanboniface:
> Not the first time I see that kind of permutations thinking

You mean in uMap or in general ? :)

PR #1201, opened 2023-07-11
Bump Leaflet to 1.9.4

yohanboniface:
Changelog:

https://github.com/Leaflet/Leaflet/blob/main/CHANGELOG.md

Sounds safe for us. Tested localy. Tests pass. Inch Allah.


davidbgk (approved):
🤞

PR #1228, opened 2023-07-24
Add target URL in CORS error message

### LINKED ISSUE #1226: Display URL in CORS message
When one has many layers, this message may be hard to interpret:

image

Maybe we should just add the target URL in the message.

yohanboniface:
fix #1226

### DIFF
--- a/umap/static/umap/js/umap.xhr.js
+++ b/umap/static/umap/js/umap.xhr.js
@@ -95,7 +95,7 @@ L.U.Xhr = L.Evented.extend({
             // 0 === request cut by user or CORS
             self.ui.alert({
               content: L._(
-                'Issue reaching that URL (network problem or CORS protection)'
+                'Issue reaching that URL (network problem or CORS protection)' + ` ${settings.uri}`
               ),
               level: 'error',
             })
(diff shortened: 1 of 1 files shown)


umap/static/umap/js/umap.xhr.js
```diff
@@ -95,7 +95,7 @@ L.U.Xhr = L.Evented.extend({
             // 0 === request cut by user or CORS
             self.ui.alert({
               content: L._(
-                'Issue reaching that URL (network problem or CORS protection)'
+                'Issue reaching that URL (network problem or CORS protection)' + ` ${settings.uri}`
```
davidbgk, line 98:
```suggestion
                `Issue reaching that URL (network problem or CORS protection) ${settings.uri}`
```

davidbgk (approved):

umap/static/umap/js/umap.xhr.js
yohanboniface: Why not, even if this means that every languages would need an update, but in that case I'd better add a variable, to keep the string "simple" for translators. I'll do that :)

Issue #1156, opened 2023-06-20
Have a sense of deepness affordance for the side panel (+navigation!)


davidbgk:
**Is your feature request related to a problem? Please describe.**

It might be confusing to see if you are currently editing a map, a layer or a feature.

**Describe the solution you'd like**

I imagine something with a sense of visual deepness, like multiple sheets that might serve as upward navigation too (like an Ariadne's thread).

Maybe unfoldable upper layers with accordions?

**Describe alternatives you've considered**

Mental load 😛 

**Additional context**

For instance in Basecamp's Message Board you can visualize the current layer you're in:
https://basecamp.com/features/message-board

yohanboniface:
Big fan of this suggestion!

One first step may be to add a simple breadcrumb, something around this:

![image](https://github.com/umap-project/umap/assets/146023/8909e142-9451-4e27-9b5f-dc2a085f2fd4)

yohanboniface:
With the work in fields (cf #2868), we think that this gain priority: it would be very helpful at least to have a clear idea (color? icon?) of which object is being edited (map? datalayer? feature?).

Eg. on this dialog, it would be helpful to quickly understand which field are edited, the map ones, or some datalayer specific ones:

<img width="1082" height="781" alt="Image" src="https://github.com/user-attachments/assets/9c68e8a5-408d-4c3c-8f43-898994f5ed2e" />

Same when this panel is opened on the field form:

<img width="772" height="1616" alt="Image" src="https://github.com/user-attachments/assets/c0a2108b-e958-4247-812d-daa771fa1bae" />

cc @Aurelie-Jallut :)

yohanboniface:
This will become even more important when we'll have nested datalayer (cf #939)

davidbgk:
@Aurelie-Jallut we tried something 😅 

<img width="766" height="437" alt="Image" src="https://github.com/user-attachments/assets/93c6ee17-df68-4115-8313-16737bf325ff" />

<img width="414" height="370" alt="Image" src="https://github.com/user-attachments/assets/e48aaab5-7f7d-428a-af80-201437e587b6" />

What do you think of these tests? Let's iterate together!

davidbgk:
## Pairing with @Aurelie-Jallut 

To add a new filter, something like that:

<img width="416" height="350" alt="Image" src="https://github.com/user-attachments/assets/a9e1a9c4-a9c7-4cc3-bb3a-884da1ec35f5" />

To select the target of a filter:

[...]

PR #1232, opened 2023-07-24
Add icons to better distinguish edit panels

yohanboniface:
cf #1156

image

image

image

image

image


yohanboniface:
Merging as I need to touch that same SVG file and I don't want to deal with conflicts :p 

We can still iterate later!

Issue #556, opened 2018-02-21
Don't say "Disable editing". Say instead "View map"


jidanni:
In the upper right corner (of e.g,. https://umap.openstreetmap.fr/en/map/testmapa_200038#6/25.265/120.630 which I am editing)
please don't say "Disable editing".
That looks like a permission change command.

I would say one of
View / Preview / View mode / View map or eyeball icon
And toggle 
Edit / Edit mode / pencil icon

With mouseover: Switch back to Edit / View mode.

VladimirSlavik:
This is actually a pair of strings "Enable editing" + "Disable editing".

Curiously enough, translation contains "Start editing" and "Stop editing" which sounds like a better fit.

PR #1233, opened 2023-07-25
Edit header revamp

yohanboniface:
cf #609 

image

image


yohanboniface:
"Mobile" mode:

image

image

There is still the "Disable editing" that does not fit when screen get smaller, but this is already the case, so I'd say we can deal with it later (need an icon, or at least a new wording, cf #556 )

davidbgk (approved):

PR #1238, opened 2023-07-31
adapt iframe min-width for popup large

### LINKED ISSUE #1236: 'Popup' and 'Popup (large)' the same width now?
Are 'Popup' and 'Popup (large)' the same width now as a result of the not showing the horizontal scroll bar? Selecting the 'Popup' shows the same result as selecting the 'Popup (large)'.

_Originally posted by @Xonnext in https://github.com/umap-project/umap/issues/1199#issuecomment-1652304833_

yohanboniface:
fix #1236

Before:

image

After:

image

### DIFF
--- a/umap/static/umap/map.css
+++ b/umap/static/umap/map.css
@@ -1456,6 +1456,9 @@ a.add-datalayer:hover,
     .umap-permanent-credits-container {
         max-width: 100%;
     }
+    .umap-popup-large iframe {
+        min-width: 300px;
+    }
 }
 
 /* ****** */
--- a/umap/static/umap/map.css
+++ b/umap/static/umap/map.css
@@ -1415,10 +1415,10 @@ a.add-datalayer:hover,
 .leaflet-contextmenu-icon {
     display: none;
 }
-.umap-popup-large iframe,
-.umap-popup-large img {
+.umap-popup-large iframe {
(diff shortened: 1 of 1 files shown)


umap/static/umap/map.css
```diff
@@ -1415,10 +1415,10 @@ a.add-datalayer:hover,
 .leaflet-contextmenu-icon {
     display: none;
 }
-.umap-popup-large iframe,
-.umap-popup-large img {
+.umap-popup-large iframe {
     /* See https://github.com/Leaflet/Leaflet/commit/61d746818b99d362108545c151a27f09d60960ee#commitcomment-6061847 */
-    max-width: 500px !important;
+    max-width: 100% !important;
+    min-width: 500px;
```
davidbgk, line 1421: I guess it will break on small devices, did you try?

yohanboniface: "Fixed"!
Sadly, I think popup sizing still needs a refacto, specifically for small screens but not only, mainly because it's partly done in JS and partly in CSS, and mainly because of [this](https://github.com/Leaflet/Leaflet/commit/61d746818b99d362108545c151a27f09d60960ee#commitcomment-6061847)

davidbgk (approved):

PR #1255, opened 2023-08-12
Better control of default view

### LINKED ISSUE #40: Centrage automatique sur le dernier marqueur renseigné
Originally reported by: **Anonymous**

---

Bonjour,
j'ai une suggestion de développement. 
Est-ce qu'il serait possible de créer une fonctionnalité qui permette de centrer par défaut une carte sur le dernier marqueur ajouté? Ce serait dans le but de permettre une meilleure consultation de celle-ci, notamment lorsqu'il s'agit de suivre un itinéraire qui évolue de jour en jour. 
Il me semble que Google Map propose une fonctionnalité similaire 

---
- Bitbucket: https://bitbucket.org/yohanboniface/umap/issue/40

yohanboniface:
image

fix #40

Let the user choose the map behaviour to define default view:

- saved center and zoom: current behaviour, and default
- bounds: will compute the view to fit all the map data, useful for having the same view in different sized devices (mobile…)
- latest feature: useful in some situation, where the map tracks some progress (travel…), this will certainly needs a bit more iterations, mainly to have an explicit default datalayer
- user location: this option used to exist but was then removed, I can remember when and why

Note: when there is a URL hash, the hash will be used

### DIFF
--- a/umap/static/umap/js/umap.forms.js
+++ b/umap/static/umap/js/umap.forms.js
@@ -417,7 +417,16 @@ L.FormBuilder.DataLayerSwitcher = L.FormBuilder.Select.extend({
   },
 })
 
-L.FormBuilder.onLoadPanel = L.FormBuilder.Select.extend({
+L.FormBuilder.DefaultView = L.FormBuilder.Select.extend({
+  selectOptions: [
+    ['center', L._('Saved center and zoom')],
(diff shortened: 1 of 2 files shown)


davidbgk (approved):
Not easy to automate testing 😬

PR #1278, opened 2023-08-23
Use magiclink extension for changelog with links

davidbgk:
/cc @eMerzh 😉

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,2 +1,3 @@
 # Force rtfd to use a recent version of mkdocs
 mkdocs==1.4.2
+pymdown-extensions==10.1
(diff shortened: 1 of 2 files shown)


yohanboniface (approved):

docs/requirements.txt
```diff
@@ -1,2 +1,3 @@
 # Force rtfd to use a recent version of mkdocs
 mkdocs==1.4.2
+pymdown-extensions==10.1
```
yohanboniface, line 3: I think this can be put in pyproject.toml actually now, as RTFD is now using a modern version of mkdocs, and is it now, it is not installed when doing a `make develop` ;)

yohanboniface: Ah, right, it will be in october, my bad!

PR #1280, opened 2023-08-24
Use X-Accel-Redirect for serving ajax-proxy request

yohanboniface:
uMap allows to use remote URL as data sources, but those URLs are not always CORS open, so this is why there is this "ajax-proxy" feature, where the URL is passed to the backend.

Additionally, there is a caching feature, which duration is configurable through frontend settings. Valid values are: disabled, 5 min, 1 hour, 1 day.

Initially, I wanted this to be totally handled by Nginx, but I never found a wayt to set the proxy_cache_valid value from a query string.

Since then, at least in OSM France servers, the ajax-proxy is still handled by a Django view, which then opens the remote URL and transfert the data. This is not optimal. And I suppose this is what is causing hicups on the OSM France servers lately.

This PR provides a mix option, where python deals with validating the URL and parsing the TTL parameter, and then it passes the hand to nginx which will serve the remote content.

So, roughtly:

- the client calls /ajax-proxy/?url=xxx&ttl=300
- python will validate the URL (not internal calls…)
- if UMAP_SENDFILE_HEADER is set, then the python returns an empty response with the path /proxy/http://url plus it will set the cache ttl through the header X-Accel-Expires
- this /proxy/ location is then handled by nginx

cc @cquest @JocelynJ could you have a look ? :)

### DIFF
--- a/docs/ubuntu.md
+++ b/docs/ubuntu.md
@@ -361,7 +361,6 @@ In Nginx:
               set $target_url $1/$2;
             }
             add_header X-Proxy-Target $target_url; # For debugging
-            resolver 8.8.8.8;
             proxy_read_timeout 10s;
             proxy_connect_timeout 5s;
             proxy_pass $target_url;
@@ -369,7 +368,6 @@ In Nginx:
             error_page 301 302 307 = @handle_proxy_redirect;
         }
         location @handle_proxy_redirect {
-            resolver 8.8.8.8;
             set $saved_redirect_location '$upstream_http_location';
             proxy_pass $saved_redirect_location;
         }
--- a/docs/ubuntu.md
+++ b/docs/ubuntu.md
@@ -332,6 +332,7 @@ And then add this new location in your nginx config (before the `/` location):
         alias /path/to/umap/var/data/;
     }
 
+
 ### Configure ajax proxy cache
 
 In Nginx:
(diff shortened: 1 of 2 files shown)


yohanboniface:
cc @ldidry @knowname @woodpeck too if you have the opportunity to have a look :)

docs/ubuntu.md
```diff
@@ -341,31 +342,41 @@ In Nginx:
...
+              set $target_url $1://$2;
+            }
+            if ( $target_url ~ (.+?)%3A(.*) ){ # fix : between destination and port
+              set $target_url $1:$2;
+            }
+            if ( $target_url ~ (.+?)%2F(.*) ){ # fix / after port, the rest will be decoded by proxy_pass
+              set $target_url $1/$2;
+            }
+            add_header X-Proxy-Target $target_url; # For debugging
+            resolver 8.8.8.8;
```
davidbgk, line 364: Do we want to use Google's services? Not sure if [Cloudflare](https://www.cloudflare.com/learning/dns/what-is-1.1.1.1/) is better though… maybe the ones from [FDN](https://www.fdn.fr/actions/dns/) but I'm not sure about the reliability?

yohanboniface: It seems to be working without specifying a `resolver`, so maybe I should just drop it ?

ldidry: I think you should. I think Nginx will use the nameserver used by the OS.

jocelynj: +1 to remove this resolver line (and line 372 below too)

jocelynj:
It looks fine from my side, but I'm not a specialist :) 

On OSM-Fr infrastructure, will you need any change? Or do you only need to modify nginx's configuration on umap container?

yohanboniface:
> On OSM-Fr infrastructure, will you need any change? Or do you only need to modify nginx's configuration on umap container?

Yep, it's just about nginx config on uMap container, so I'm autonomous :)

davidbgk (approved):

docs/ubuntu.md
yohanboniface: Nope

PR #1288, opened 2023-08-29
Remove custom message from onbeforeunload event

yohanboniface:
It's not supported anymore by browsers.


davidbgk (approved):
Code isLessDirty 🙃

PR #1290, opened 2023-08-29
A11Y: switch from links to buttons when pertinent

davidbgk:
CSS part is missing to be able to actually toggle to the `button` element. But some choices had to be discussed/made prior to that.

### DIFF
--- a/umap/static/umap/js/umap.core.js
+++ b/umap/static/umap/js/umap.core.js
@@ -315,10 +315,10 @@ L.DomUtil.createFieldset = (container, legend, options) => {
 
 L.DomUtil.createButton = (className, container, content, callback, context) => {
   // TODO: actually switch to buttons’ elements.
-  const el = L.DomUtil.add('a', className, container, content)
-  el.href = '#'
-  // const el = L.DomUtil.add('button', className, container, content)
-  // el.type = 'button'
+  // const el = L.DomUtil.add('a', className, container, content)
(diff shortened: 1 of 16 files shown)


umap/static/umap/js/umap.core.js
```diff
@@ -286,14 +286,25 @@ L.DomUtil.createFieldset = (container, legend, options) => {
 }
 
 L.DomUtil.createButton = (className, container, content, callback, context) => {
+  // TODO: actually switch to buttons’ elements.
   const el = L.DomUtil.add('a', className, container, content)
   el.href = '#'
+  // const el = L.DomUtil.add('button', className, container, content)
```
davidbgk, line 292: This is the line to uncomment to get the full button experience 😅

yohanboniface:
Our buttons seems grey instead of white, like buttons comings from Leaflet/plugins:

image

I'd say we should keep them white, or make them all grey.

yohanboniface:
This button text seems to long to me:

image

Save/Enregistrer already says enough, no need to add "current changes" imho. :)

yohanboniface:
If we touch the labels of those buttons, a bit of context to take into account: https://github.com/umap-project/umap/issues/556

yohanboniface:
Broken picto:

image

yohanboniface (approved):
Let's go!

PR #1291, opened 2023-08-29
Avoid using a tuple as a content type when guessed

davidbgk:
Only use the content type with mimetypes.guess_type


yohanboniface:
Can we add a quick unit test ?

PR #1293, opened 2023-08-30
Put focus on name when opening feature edit panel

### LINKED ISSUE #243: [wishlist] The marker's name 'field' should get the focus after drawing a new marker
Naming a marker is (in my opinion) the next thing one does after drawing it.
By giving the focus to the name field, this would avoid an additional left click.

yohanboniface:
fix #243


AbdullahWins (commented):
well done!

davidbgk (approved):

Issue #584, opened 2018-06-01
More granularity for editor permissions


akien-mga:
I'd like to have more options to define what editors can do, especially when using the "anyone can edit" option, which seems to make everyone *admin* (everyone has access to the full map settings and can even delete the map).

It would be nice if the editor role (or a new role if need be) would be able to edit the map (add/edit markers in existing layers, etc.) but not have access to the map settings, and maybe also be limited when it comes to deleting content (should not be able to delete layers for example). Ideally there should be granularity for all actions so that you can configure such a role yourself, but a preset with a good default configuration would be good already.

---

**My use case:** I want to make a map for all users and contributors of the @godotengine project to pin their location, to help organize local meetups and events. We're talking hundreds (possibly thousands) of people who need to be able to place a marker on the map, so I can't add them all to the editors whitelist manually, and thus need to use the "everyone can edit" option.

But as far as I can see, it means that everyone can decide to delete the layer with the data carefully input by hundreds of users, or even delete the map itself altogether, which is a big concern to me.

yohanboniface:
>  So if we'd go on the intermediate open permission level, we'd need the frontend to know which buttons/actions to show or hide.

Note: this will be easier once https://github.com/umap-project/umap/pull/583 is done.

akien-mga:
> This should not be the case. If so, it's a bug. In theory, only the owner can delete the map (not even the editors).

I'll admit that I did not test it as I don't want to delete my map, but when browsing a map while unlogged (even in Private Browsing mode), I still see those buttons:
![screenshot_20180604_153621](https://user-images.githubusercontent.com/4701338/40920511-10383ad2-680d-11e8-91c1-33b7e5226c4b.png)

Still as anonymous/private browsing user, I also have access to most other settings AFAICT, including some permissions management (but only limited to apparently adding editors?):
![screenshot_20180604_153739](https://user-images.githubusercontent.com/4701338/40920555-380ad9ac-680d-11e8-8d43-c353ada0e958.png)

> Basically, as of today, the right management rely on resource + verb, where resource is Map or DataLayer, and verb is one of GET, POST/PUT/PATCH, DELETE.
The first thing you can see is that there is no resource for a feature element, the smallest element is the DataLayer.

Thanks for the details, that makes sense. It does indeed limit the granularity of permissions that could be offered, though what is doable with the current design might already be a nice improvement. Basically it would mean having per-map (and optionally per-layer) permissions of the type read/write/delete.

[...]

PR #1307, opened 2023-09-07
Allow to define permissions for each datalayer instead of for the whole map

yohanboniface:
Work in progress!

cf #584 

Todo list:

- [x] Create DataLayer.edit_status and expose in the API
- [x] adapt `can_edit` logic
- [x] allow to edit `edit_status` in the frontend
- [x] take it into account to allow or not editing feature
- [x] Expose only editable layers for new feature / editing feature's layer
- [x] make it work with anonymous maps
- [x] populate DataLayer.edit_status values from Map.edit_status
- [x] hide frontend map forms to users that do not have right to save
- [x] only show buttons create/delete datalayers to users with "map" access level
- [x] remove Map.edit_status from frontend/API
- [ ] ~delete Map.edit_status (certainly not in this PR just in case)~
- [x] backend tests
- [x] adapt frontend tests
- [x] integration tests (playwright)
- [x] decide if we keep both OWNER and EDITORS modes at datalayer level (this distinction does not exist anymore at map level)
- [x] naming (allowEdit / canEdit / allowMapEdit, hasDirty / isDIrty)
- [x] check delete map button for editors
- [x] check context menu
- [x] set default value for `Datalayer.options.editMode`
- [ ] ~add unittests for `editMode` value returned by back-office~ (it's json dumped in a `<script>` in a HTML page, so it's better to use integration tests, which test the behaviour instead of the key itself, and are already done)


umap/forms.py
```diff
@@ -8,8 +8,12 @@
 
 from .models import Map, DataLayer
 
-DEFAULT_LATITUDE = settings.LEAFLET_LATITUDE if hasattr(settings, "LEAFLET_LATITUDE") else 51
-DEFAULT_LONGITUDE = settings.LEAFLET_LONGITUDE if hasattr(settings, "LEAFLET_LONGITUDE") else 2
+DEFAULT_LATITUDE = (
```
yohanboniface, line 11: Black noise :(

davidbgk:
```suggestion
DEFAULT_LATITUDE = getattr(settings, "LEAFLET_LATITUDE", 51)
```

Even shorter 😎

umap/static/umap/js/umap.js
```diff
@@ -192,16 +192,15 @@ L.U.Map.include({
       this
     )
 
-    let isDirty = false // global status
+    let isDirty = false // self status
     try {
       Object.defineProperty(this, 'isDirty', {
         get: function () {
-          return isDirty || this.dirty_datalayers.length
+          return isDirty
```
yohanboniface, line 199: Not sure why this check on `dirty_datalayers` was here before. As I can see, the `isDirty` status should always be set when a datalayer is made dirty…

umap/static/umap/js/umap.js
```diff
@@ -192,16 +192,15 @@ L.U.Map.include({
...
         },
         set: function (status) {
-          if (!isDirty && status) self.fire('isdirty')
           isDirty = status
-          self.checkDirty()
+          this.checkDirty()
```
yohanboniface, line 203: Not sure this change is harmless. It works in our browser, but may it break on some older versions ?

PR #1311, opened 2023-09-13
Merge OWNER and EDITORS into RESTRICTED level

yohanboniface:
At Map level, there is no more distinction between owner and editors, they all share the same rights (but the right to delete the map, but there is no level to control it).

So having this distinction at DataLayer level does not make sense anymore, in my opinion.

This is not retrocompatible, as the change at Map level, but it goes in the sense of permissions management more targeted to real uMap usage while keeping the system simple.

If we agree on this change, I'll add a data migration.

### DIFF
--- a/umap/forms.py
+++ b/umap/forms.py
@@ -53,8 +53,8 @@ class DataLayerPermissionsForm(forms.ModelForm):
 
 class AnonymousDataLayerPermissionsForm(forms.ModelForm):
     STATUS = (
-        (Map.ANONYMOUS, _("Everyone can edit")),
-        (Map.OWNER, _("Only editable with secret edit link")),
+        (DataLayer.OPEN, _("Everyone can edit")),
+        (DataLayer.RESTRICTED, _("Only editable with secret edit link")),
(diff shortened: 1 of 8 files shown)


davidbgk (commented):
Looks good to me but I thought we would do that once the changes from #1307 are deployed and there are not too much complaints about that merge of permissions?

yohanboniface:
I thought about that scenario, but part of this merge/simplification is already done in #1307 (given there is no more distinction owner/editors on the map object itself). So I arrived to the conclusion that it was more consistent to also do this change. But in a separate PR because I know it must be discussed :)
Basically, I find very strange that an editor could change anything on the map (name, style, editors…), but would not have right to manage some datalayers (if those would be in OWNER mode).
Also, keep in mind that until we delete for real Map.edit_status (which we don't use anymore but keep in the db), it's possible to rollback this distinction, but I guess that would also mean adding back edit_status on the map itself.

yohanboniface:
We instead keep those levels also in Map.edit_status.

PR #1313, opened 2023-09-14
Add outlinkTarget to default interactionProperties

### LINKED ISSUE #1312: Provide "Link target"-chooser in Default interaction options
The selector "open link in..." is only available inside layer properties. So for a map with multiple layers I have to select the link target on each layer again and again.

"Default interaction options" allow choosing all other options from that menu as per-map defaults. The layer-properties can then override them.

**Additional context**
issue-open-links

jschleic:
Fix #1312

Is there something else, that needs to be added? Seems to work locally.


yohanboniface:
Sounds good! Thanks :+1:

PR #1320, opened 2023-09-19
A more newbie-friendly Docker HOWTO

jvmatl:
I really like uMap and want to self-host. I think it's great that it is available in docker form, but the documentation for how to get started with the docker version was not very newbie-friendly. It took me several hours of experimentation to work out exactly what was needed, where to configure it, etc. 

So I put together what is, hopefully a more noob-friendly introduction to running uMap voa docker. The config files and procedures have been run through multiple times on Ubuntu 22.04LTS and should work without modification, to allow users to at least get *something* up and running quickly.

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -2,43 +2,132 @@
 
 There is now an official [uMap](https://hub.docker.com/r/umap/umap) image.
 
-To run it with docker compose, use a `docker-compose.yml` like this:
+To run it with docker compose, create a `docker-compose.yml` like this:
 
+_Sample docker-compose.yml_ 
 ```yaml
 version: '3'
 services:
   db:
-    image: postgis/postgis:14-3.3-alpine
+    # check https://hub.docker.com/r/postgis/postgis to see available versions
+    image: postgis/postgis:14-3.4-alpine
     environment:
       - POSTGRES_HOST_AUTH_METHOD=trust
     volumes:
-      - db:/var/lib/postgresql/data
+      - umap_db:/var/lib/postgresql/data
 
   app:
-    image: umap/umap:x.x.x
+    # Check https://hub.docker.com/r/umap/umap/tags to find the latest version
+    image: umap/umap:1.3.7
     ports:
+      # modify the external port (8001, on the left) if desired, but make sure it matches SITE_URL, below
       - "8001:8000"
     environment:
       - DATABASE_URL=postgis://postgres@db/postgres
-      - SECRET_KEY=some-long-and-weirdly-unrandom-secret-key
-      - SITE_URL=https://umap.local/
-      - UMAP_ALLOW_ANONYMOUS=True
-      - DEBUG=1
+      - SITE_URL=https://localhost:8001/
+      - STATIC_ROOT=/srv/umap/static
+      - MEDIA_ROOT=/srv/umap/uploads
     volumes:
-      - data:/srv/umap/uploads
-
+      - umap_userdata:/srv/umap/uploads
+      # FIX the path on the left, below, to your location 
+      - /home/ubuntu/umap.conf:/etc/umap/umap.conf
+    restart: always
+    depends_on:
+      - db
+    
 volumes:
-  data:
-  db:
+  umap_userdata:
+  umap_db:
+
 ```
+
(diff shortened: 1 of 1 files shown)


davidbgk:
@jvmatl this is invaluable feedback, thank you so much!

docs/docker.md
```diff
@@ -2,43 +2,132 @@
...
+docker compose up
+```
+... and let it run some initial setup until the output quiesces with a message about spawning uWSGI workers. Because there is a race between the time the app tries to connect to the DB and when the DB is actually ready, you might see a few exceptions/erros about 'psychopg' being unable to connect. This should sort itself out as the app retries. 
+
+### Create superuser
+
+Now you need to create your site superuser. Stop the server (Ctrl-C) and then type:
+```bash
+docker-compose run app /venv/bin/umap createsuperuser` 
+``````
```
davidbgk, line 121: A bit too much backticks (and one extra on the line above this one).

docs/docker.md
```diff
@@ -2,43 +2,132 @@
...
+
+Once that's done, you can relaunch your server with `docker compose up`
+
+### Try It!
+
+You should not ne able to brose to your uMap instance form a browser on your local system, by pointing your browser to ${SITE_URL} (see the docker-compose file, above.)
```
davidbgk, line 127:
```suggestion
You should now be able to browse to your uMap instance form a browser on your local system, by pointing your browser to `https://localhost:8001/` (equivalent to `${SITE_URL}` in the docker-compose file, above).
```

docs/docker.md
jvmatl: oops, thanks! fixing

jvmatl:
Please note, I am a COMPLETE newbie to running this, so it's possible I left out important configuration pieces. All I can say is, it seems to work now, and I can create a map, and save it, and then view it with a link.

docs/docker.md
```diff
@@ -2,43 +2,132 @@
...
 volumes:
-  data:
-  db:
+  umap_userdata:
+  umap_db:
+
 ```
 
-### Create superuser
+next, create a basic settings file, named `umap.conf`` in the same directory.
```
davidbgk, line 44:
```suggestion
Next, create a basic settings file, named `umap.conf` in the same directory. 
```

docs/docker.md
```diff
@@ -2,43 +2,132 @@
...
+
+---
+
+### Initial launch
+
+Start the server with 
+```bash
+docker compose up
+```
+... and let it run some initial setup until the output quiesces with a message about spawning uWSGI workers. Because there is a race between the time the app tries to connect to the DB and when the DB is actually ready, you might see a few exceptions/erros about 'psychopg' being unable to connect. This should sort itself out as the app retries.
```
davidbgk, line 114:
```suggestion
... and let it run some initial setup until the output quiesces with a message about spawning uWSGI workers. Because there is a race between the time the app tries to connect to the DB and when the DB is actually ready, you might see a few exceptions/errors about 'psycopg' being unable to connect. This should sort itself out as the app retries. 
```

👍 Hopefully it will be mitigated by https://github.com/umap-project/umap/pull/1314

docs/docker.md
```diff
@@ -2,43 +2,132 @@
...
+  umap_db:
+
 ```
 
-### Create superuser
+Next, create a basic settings file, named `umap.conf` in the same directory. 
+
+You can use this example below and it will run, but you may want to look at the project sample config, using `wget https://raw.githubusercontent.com/umap-project/umap/master/umap/settings/local.py.sample -O /etc/umap/umap.conf` and modify as needed. 
+
+Make sure the settings in the docker-compose don't conflict with the sample config and vice-versa. In particular, remove the DATABASES section from the config file if useing the docker-compose file, or it will override the DATABASE_URL setting and things won't work.
```
yohanboniface, line 48:
```suggestion
Make sure the settings in the docker-compose don't conflict with the sample config and vice-versa. In particular, remove the DATABASES section from the config file if using the docker-compose file, or it will override the DATABASE_URL setting and things won't work. 
```

yohanboniface:
Thanks a lot! :+1: 

I've spotted a small typo while reviewing. Once it's fixed, I'll merge!

PR #1321, opened 2023-09-19
Restrict possible min/max zoom values

### LINKED ISSUE #1318: Don't let people enter insane zoom in overlay
Cf https://forum.openstreetmap.fr/t/impossible-dacceder-a-mes-cartes-sur-notre-compte-umap/6776/6

davidbgk:
Fix #1318

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "leaflet-contextmenu": "^1.4.0",
         "leaflet-editable": "^1.2.0",
         "leaflet-editinosm": "0.2.3",
-        "leaflet-formbuilder": "0.2.6",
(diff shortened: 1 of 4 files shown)


yohanboniface:
The absolute max zoom value is a bit tricky to define once and for all. The tilelayer is at the end what will usually define this value for a given map.

`20` seems legit, but some custom use cases may want a bit more. Also, we need to also limit the `options.overlay.min/maxZoom`. So I'd suggest having a `map.options.absoluteMaxZoom` hard coded, that could be then used where needed and will prepare in case we need to expose that option to users at some point.

yohanboniface (approved):

PR #1322, opened 2023-09-19
Ran through the 'Ubuntu from Scratch' procedure using a fresh Ubuntu 22.04.3 LTS VM and took notes on what worked and what didn't.

jvmatl:
Here's a summary of the changes and why

* added 'apt update' before the install -- only _needed_ on a completely fresh VM, but never hurts

* added an install of the virtualenv package, because it is used later but: do you still need it AND python3-venv? I'm assuming it was done that way for a reason.

* changed indentation on the mkdir so it will format correctly

* added `-D ~postgres` to the commands executed as postgres user because the postgres user does not, by default, have permission to read into real user's home directories (at least in 22.04) and so the commands would error out

* change `source <activate>` to `. activate` because the line `sudo -u umap -i` launches `sh`, not `bash`, and `sh` does not recognize the source command. `.` works in either.

* remove the instructions to change STATIC_ROOT and MEDIA_ROOT because they are already defaulted to those values in the current version sample of local.py

* added a notice to change the SECREY_KEY and ADMINS list just because it seemed like a good idea.

Side note: I have not seen any docs anywhere that explains exactly what the secret Key does, or where it is used. It's used a magic variable in the config file.. ?

I am not currently messing with nginx or wsgi, so I can't comment on those parts.

### DIFF
--- a/docs/ubuntu.md
+++ b/docs/ubuntu.md
@@ -76,7 +76,7 @@ you will need to run again this last line.*
 
     nano /etc/umap/umap.conf
 
-* update the SECRET_KEY
+* update the [SECRET_KEY](settings.md#secret_key)
 * update the ADMINS list
 
 ## Create the tables
--- a/docs/ubuntu.md
+++ b/docs/ubuntu.md
@@ -6,13 +6,14 @@ You need sudo grants on this server, and it must be connected to Internet.
 
 ## Install system dependencies
 
-    sudo apt install python3 python3-dev python3-venv wget nginx uwsgi uwsgi-plugin-python3 postgresql gcc postgis libpq-dev
+    sudo apt update
+    sudo apt install python3 python3-dev python3-venv virtualenv wget nginx uwsgi uwsgi-plugin-python3 postgresql gcc postgis libpq-dev
 
 *Note: nginx and uwsgi are not required for local development environment*
 
 ## Create deployment directories:
 
-   sudo mkdir -p /etc/umap
+    sudo mkdir -p /etc/umap
 
 *You can change this path, but then remember to adapt the other steps accordingly.*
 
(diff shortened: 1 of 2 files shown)


davidbgk:
> Side note: I have not seen any docs anywhere that explains exactly what the secret Key does, or where it is used. It's used a magic variable in the config file.. ?

Maybe we should link to that page? Would that help better understand the consequences? https://docs.djangoproject.com/en/4.2/ref/settings/#std:setting-SECRET_KEY

jvmatl:
re: SECRET_KEY, I think it would be good to include the link to the django docs on this page: https://umap-project.readthedocs.io/en/master/settings/

davidbgk:
@jvmatl alright, I created a PR about that issue #1323

jvmatl:
lol. I didn't realize you were going to make another issue - I updated the settings page here. Do you want me to back it out?

docs/ubuntu.md
```diff
@@ -75,12 +76,8 @@ you will need to run again this last line.*
 
     nano /etc/umap/umap.conf
 
-Change the following properties:
-
-```
-STATIC_ROOT = '/srv/umap/var/static'
-MEDIA_ROOT = '/srv/umap/var/data'
-```
+* update the SECRET_KEY
```
davidbgk, line 79:
```suggestion
* update the [SECRET_KEY](settings.md#secret_key)
```

Considering we'll have a better version soon, let's link to it! 😉

jvmatl:
Ugh. I'm sorry, I don't know how to remove that change from the branch (this is my first time trying to fix something directly in github's codespaces, and I don't see a way to revert my prior change to settings.md

davidbgk:
> Ugh. I'm sorry, I don't know how to remove that change from the branch (this is my first time trying to fix something directly in github's codespaces, and I don't see a way to revert my prior change to settings.md

No problem, I took care of it 👍

davidbgk (approved):
Newcomers fresh eyes are the best, thanks!

yohanboniface:
:+1:

Issue #430, opened 2017-01-19
How to add username to feature?


SABR-13:
I would like to add the username of the logged in user to the feature that they have created / edited. What would be the easiest way to do this automatically? I don't want them to type it in themselves. Is there some way to get the details of the user that has logged in while they are editing the map?

The .json file should be something like this:

`{"type":"Feature","properties":{"name":"test","description":"something", "user":"user10"}`

Thanks!

Binnette:
Hi @SABR-13, i don't know if you use Overspass queries to build your json. But if so you can output "meta" information like this:
```
node["amenity"="public_bookcase"]({{bbox}});
out body meta;
```

And you will get the name of the use that modify the osm object last:
```xml
  <node id="4989704906" lat="45.1646296" lon="5.7680993" timestamp="2019-12-06T11:04:23Z"
        version="3" changeset="78044506" uid="918586" user="Binnette">
    <tag k="amenity" v="public_bookcase"/>
    <tag k="capacity" v="60"/>
    <tag k="lit" v="no"/>
    <tag k="mapillary" v="JYGqUOMkrKzkN8fJZMtmeQ"/>
    <tag k="opening_hours" v="24/7"/>
    <tag k="public_bookcase:type" v="wooden_cabinet"/>
  </node>
````

In the resulting xml, you will get the user uid and name:
`uid="918586" user="Binnette"`

Have a nice day.

openbrian:
@yohanboniface  working on this feature.

yohanboniface:
Great! :+1: 

Don't hesitate to ask questions and submit here choices that need discussion, we'll be available for help. :)

One first thing : keep in mind that some users may contribute to a map without being logged in.

openbrian:
@yohanboniface ahhh, while I was editing my comment, you posted your comment.  Please go back and see my update.  I do have lots of questions.

openbrian:
Forgot to mention I want this owner field hidden.  Do I use BlurInput for that?

openbrian:
See #584

yohanboniface:
>  In the geojson of every feature created is the owner and the id of the user logged in.

Then do you mean there are two fields added: `owner` and `id` ?

> Some people may not want that data. It should be harmless for most people. Let me know if adding it should be an application option.

I'd say so. In a first move, at least a Django setting to opt in this feature. And maybe we may want to have a setting map by map, so users can decide for their maps.

[...]

PR #1324, opened 2023-09-19
Add username to features

openbrian:
For https://github.com/umap-project/umap/issues/430

Will look like this.

```
{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {
        "name": "K",
        "description": "K",
        "owner": 2
      },
      "geometry": {
        "type": "Point",
        "coordinates": [
          -1.911621,
          54.72462
        ]
      }
    }
  "_umap_options": {
    "displayOnLoad": true,
    "browsable": true,
    "name": "Layer 1"
  }
}
```


openbrian:
Goofed up this PR.  Working on cleaning it up or creating a new one.

openbrian:
ok removed the junk

umap/static/umap/js/umap.forms.js
```diff
@@ -939,6 +939,10 @@ L.U.FormBuilder = L.FormBuilder.extend({
 
   defaultOptions: {
     name: { label: L._('name') },
+    owner: {
+      label: 'owner',
```
davidbgk, line 943:
```suggestion
      label: L._('owner'),
```

Useful to be able to translate user-facing strings.

umap/static/umap/js/umap.forms.js
```diff
@@ -939,6 +939,10 @@ L.U.FormBuilder = L.FormBuilder.extend({
...
+//      handler: 'BlurInput',  // this field should be hidden, not sure if blur does it.
```
davidbgk, line 944: I don't think the [BlurInput](https://github.com/yohanboniface/Leaflet.FormBuilder/blob/master/Leaflet.FormBuilder.js#L290-L307) would achieve this. You probably need to extend `Input` with a [custom `type`](https://github.com/yohanboniface/Leaflet.FormBuilder/blob/master/Leaflet.FormBuilder.js#L271-L273) which returns `'hidden'`.

yohanboniface:
Given this property is not going to be editable by users, I'm not why we'd use the form to set it. Why not just setting it in the FeatureMixin.initialize method ? :)

yohanboniface:
Sounds good!

Let's add a global map option to opt in this, controlled by a django settings for now (we'll see to expose it to all users later). Something like `options.trackUsers` and `settings.UMAP_TRACK_USERS`, wdyt ?

openbrian:
First, I feel "trackUsers" is a lightning rod that should be avoided.  Let's call this featuresHaveOwners.  I think it's fine to add an option... But I'm getting dizzy with finding where to initialize it!  

Currently I have this `UMAP_DEFAULT_FEATURES_HAVE_OWNERS` in base.py.  The MapDetailMixin version of get_geojson() will set this

```
{
  "properties": {
    "features_have_owners": settings.UMAP_DEFAULT_FEATURES_HAVE_OWNERS
  }
}
```

I added a field in "Edit map settings" / "Default properties".  But I don't know how to get my map option to set that widget.

yohanboniface:
> featuresHaveOwners

`setFeaturesOwners` ?

> The MapDetailMixin version of get_geojson() will set this

I'd add it in the `MapDetailMixin.get_context_data`

> I added a field in "Edit map settings" / "Default properties". But I don't know how to get my map option to set that widget.

No need for that yet, I'd say. A Django setting is enough for this step.

openbrian:
Got the map default properties form field working.  Issue was just a typo.  If you really want it removed let me know.

I *think* this is a working version.  I'm able to toggle the base.py setting and see a difference in the feature properties.  Please review code and/or test.

yohanboniface:
Looks good to me!

My only concern is about the naming. @davidbgk any alternative suggestion, or should we go as is ? :)

davidbgk:
I wonder if adding that very specific property, we hide the fact that we need a way to let instance's owners enrich the data with custom properties. Maybe some kind of basic plugin architecture to achieve that would be more useful to the whole community.

I'm afraid that we have to repeat the current process and end up with legacy properties that we have to deal with for the rest of uMap's life 😅

davidbgk (approved):
After discussion with @yohanboniface, let's make that first step, thanks a lot for your contribution @openbrian

PR #1326, opened 2023-09-20
Fix map displayed more than once in user dashboard when multiple editors

### LINKED ISSUE #1325: Same map listed many times in my catalog
**Describe the bug**
when viewing my map catalog (Mon Espace) I can see the same map several times. The number of times seems to relate to the number of editors of the map.
- account : cartocite
- map : http://umap.openstreetmap.fr/fr/map/gares-sncf-2022_702105

**To Reproduce**
1. Login with the account cartocite on the umap.openstreetmap.fr instance
2. Open the catalog (http://umap.openstreetmap.fr/fr/me)
3. The map "Gares SNCF 2022" appears 5 times – note only the first one shows some map content. I checked they all point to the same URL.

**Expected behavior**
Each map should be listed only once.

**Screenshots**
image

 - OS: Ubuntu 22.04
 - Browser : Firefox 117.0.1 (64 bits)

yohanboniface:
fix #1325

### DIFF
--- a/umap/views.py
+++ b/umap/views.py
@@ -275,7 +275,7 @@ class UserDashboard(PaginatorMixin, DetailView, SearchMixin):
 
     def get_maps(self):
         qs = self.get_search_queryset() or Map.objects.all()
-        qs = qs.filter(Q(owner=self.object) | Q(editors=self.object))
+        qs = qs.filter(owner=self.object).union(qs.filter(editors=self.object))
         return qs.order_by("-modified_at")
 
     def get_context_data(self, **kwargs):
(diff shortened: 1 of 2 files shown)


umap/views.py
```diff
@@ -275,7 +275,7 @@ def get_object(self):
 
     def get_maps(self):
         qs = self.get_search_queryset() or Map.objects.all()
-        qs = qs.filter(Q(owner=self.object) | Q(editors=self.object))
+        qs = qs.filter(owner=self.object).union(qs.filter(editors=self.object))
```
davidbgk, line 278:
```suggestion
        qs = qs.union(qs.filter(owner=self.object), qs.filter(editors=self.object))
```

This version is more readable to me, not a strong opinion though.

davidbgk (approved):

PR #1328, opened 2023-09-20
Fix error when trying to change map owner

yohanboniface:
cf https://forum.openstreetmap.fr/t/bouton-valider-pour-changement-proprietaire-n-apparait-pas/17658/4

This bug has been introduced with this change:

https://github.com/umap-project/umap/commit/8b4842ff212541278957d74f86e886fb5c4d082f

That was not the correct fix, and this one should be the proper one.

We don't want to edit the permissions reference until we do save, otherwise user cannot save as it is already no more the owner.

So:
- change permissions.options
- save
- commit those changes to map.options.permissions
- use only those values to check for isOwner and isAnonymousMap


davidbgk (approved):
a.k.a. scier la branche sur laquelle on est assis 🫣

yohanboniface:
Cherry-picked in #1307

PR #1336, opened 2023-09-23
Redirect to canonical when URL contains the ?edit parameter

yohanboniface:
We don't want those link to be shared.


davidbgk (approved):
Please add `{}` 🙏

PR #1343, opened 2023-09-25
Set an explicit width to popup large container

yohanboniface:
Until now, the width was computed on the fly by Leaflet, based on the popup content, but this leads to a bunch of issue on image/iframe width.

Before:

image

After:

image

cf https://forum.openstreetmap.fr/t/bugs-divers-relevees-sur-une-umap-absent-sur-une-autre/17254/8


davidbgk (approved):

Elefant-aus-Wuppertal:
Hi,
Could this possibly be done not only for the large popup, but also for the small ("normal") popup?

PR #1346, opened 2023-09-26
Make sure SHORT_SITE_URL is defined to expose Map.short_url

yohanboniface:
image


umap/views.py
```diff
@@ -585,11 +585,11 @@ def get_umap_id(self):
         return self.object.pk
 
     def get_short_url(self):
-        shortUrl = None
-        if hasattr(settings, "SHORT_SITE_URL"):
+        short_url = None
```
davidbgk, line 588: Do we really want to default to `None` instead of `''`?

davidbgk (approved):

PR #1310, opened 2023-09-08
Give more room to the title of the map

davidbgk:
Fix #1308

PR #1348, opened 2023-09-26
Fix map's title when “dirty star” is present

### LINKED ISSUE #1308: Enlarge map title display area
It's a shame not to be able to see the map title, whereas on desktop there's plenty of space to display it.

image

jschleic:
The star indicating a dirty map uses more than 5px (depending on font-size). So the additional star switched from showing the map name to ellipses-overflow in px-implementation.
1ch is the with of the digit '0'.

So the issue in #1308 is not about max-width but also reproducible for very short map names.

Nevertheless max-width of 200px for a.map-name seems a bit restrictive for desktop browsers and could be increased?!

Replaces #1310

### DIFF
--- a/umap/static/umap/map.css
+++ b/umap/static/umap/map.css
@@ -516,7 +516,7 @@ ul.photon-autocomplete {
 a.map-name:after {
     content:  '\00a0';
     padding-left: 3px;
-    width: 5px;
+    width: 1ch;
     display: inline-block;
 }
 .umap-is-dirty a.map-name:after {


yohanboniface:
Thanks a lot! Sounds good to me! @davidbgk ?

davidbgk (approved):

davidbgk:
Alright, I don't think it fully fixes #1308 but it's a good step in that direction, at least with the “dirty” scenario.

The max width will be fixed in #1290 as a flexbox rendering of the header is in progress and should resolve the current width issue.

Thanks for your proposal @jschleic, really appreciated!

PR #1349, opened 2023-09-27
Hide attribution on small screen and add a small ? to display it

### LINKED ISSUE #844: Adjust the Attribution leaflet
On an iPhone7 in Safari, the little box in the bottom right corner "Data © OpenStreetMap Contributors, map style provided by FOSSGIS e.V..." is way too large and obtrudes the view. (It's even worse on Opera)

I'd propose shortening the text and maybe downsize the font a bit.

yohanboniface:
fix #844

[Screencast from 2023-09-27 10-28-28.webm](https://github.com/umap-project/umap/assets/146023/54a927b0-97c9-4c62-82fd-f83a89df4f17)

@davidbgk can you test with a real touch device ? I don't have one at hand, so I've just checked with Chrome device emulation…

### DIFF
--- a/umap/static/umap/map.css
+++ b/umap/static/umap/map.css
@@ -147,7 +147,14 @@ a.umap-control-less {
     margin-bottom: 5px!important;
     padding: 0.5rem;
 }
-
+.attribution-toggle {
+    display: none;
+    width: 24px;
+    height: 24px;
+    vertical-align: middle;
+    background-image: url('./img/16.svg');
+    background-position: 0px 0px;
+}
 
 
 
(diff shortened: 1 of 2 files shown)


davidbgk:
It does work with my iPhoneX (quite small) with both Safari and Firefox Focus. Not sure about the opacity though:

A1F24A66-6AD8-4862-A649-FAC44333C33B

26EFC1F2-DA92-410B-BE59-8FFEED2CDBDE

umap/static/umap/map.css
```diff
@@ -1542,6 +1549,22 @@ a.add-datalayer:hover,
...
+
+    .attribution-toggle {
+        display: inline-block;
+    }
+    .attribution-container {
+        display: none;
+    }
+    .leaflet-control-attribution:active .attribution-container,
+    .leaflet-control-attribution:hover .attribution-container {
+        display: inline;
```
davidbgk, line 1561:
```suggestion
        display: inline;
        background-color: rgba(255, 255, 255);
```

yohanboniface: `#fff` at this point ?

davidbgk: Yup, I kept the declaration from Leaflet but `#fff` looks good to me too.

PR #1351, opened 2023-09-28
First naive version of datalayer grouping

yohanboniface:
image

A few questions/remarks:

- we are using datalist so groups name have to be written only once and will then appear in autocomplete, but if one wants to reword a group name, they will need to reword it once, but then go to all datalayers of this group to select the new value
- the "toggle all" button does not reflect the current state of the visibility of the group (tricky: all visible, all hidden, or mixed)
- at this stage, datalayers without a group are not displayed any more in the browser, letting users choose either to have a group "Others" or let them hidden, but maybe we want something more automatic
- we do not reflect groups yet in the caption panel, but we should do if this is the direction we want to take
- going this way, we touch the idea of also changing colors and settings for a whole group at once, so that would mean a sort of datalayer hierarchy, but this would open a lot of questions (perms being maybe the tricky one) 
- no integration tests yet as I'm not sure yet it's the way to go

### DIFF
--- a/umap/static/umap/js/umap.controls.js
+++ b/umap/static/umap/js/umap.controls.js
@@ -514,10 +514,43 @@ L.U.DataLayersControl = L.Control.extend({
   update: function () {
     if (this._datalayers_container && this._map) {
       this._datalayers_container.innerHTML = ''
-      this._map.eachDataLayerReverse(function (datalayer) {
-        this.addDataLayer(this._datalayers_container, datalayer)
-      }, this)
+      const hasGroups = !!Object.keys(this._map.groups_index).length
+      L.DomUtil.classIf(this._datalayers_container, 'grouped', hasGroups)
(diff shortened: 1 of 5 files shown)


rprovilij:
Hoping to keep the conversation going: 
1. Could a "Rename group" feature automatically iterate over all layers linked with that group? When a group name is changed, layer associations are updated. 
2. Simplifying it to two states, all visible or all hidden, would already be valuable. 
3. When creating a new map, the initial layer could be placed within an undefined group (as standard) that would be renamed later. Also the layers of existing maps could be placed in this undefined group. 
4. Groups could for now be reflected in the captions panel with headings. 
5. Enabling grouping for organizational purposes only would already be extremely helpful(!) Implementing hierarchical settings could perhaps be examined in the next step?
6. Eager to hear your thoughts.

PR #1356, opened 2023-10-04
switch to grid layout for multiple choice buttons

jschleic:
to allow multi-line texts in translations

in svedish and german translation the buttons for "Open link in..." are longer. This results in multiline-buttons that the fixed height cannot handle.

grafik

I first thought about shortening translations - but how can one make sure, that every string in every language fits within the buttons? Thus I suggest to switch to the `grid` layout, which is already used by the button-bar. 

The only difference between `.button-bar` and `.umap-multiplechoice` is now the grid-gap.

### DIFF
--- a/umap/static/umap/base.css
+++ b/umap/static/umap/base.css
@@ -342,18 +342,23 @@ input.switch:checked ~ label:before {
 input.switch:checked ~ label:after {
     margin-left: 3em;
 }
-.button-bar {
+.button-bar, .umap-multiplechoice {
     margin-top: 5px;
     text-align: center;
     display: grid;
-    grid-gap: 7px;
     width: 100%
+}
+.button-bar {
+    grid-gap: 7px;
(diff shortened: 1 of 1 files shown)


umap/static/umap/base.css
```diff
@@ -342,19 +342,24 @@ input.switch:checked ~ label:before {
...
     width: 100%
 }
+.button-bar {
+    grid-gap: 7px;
+}
 .button-bar.half {
     grid-template-columns: 1fr 1fr;
 }
-.button-bar.third {
+.button-bar.third, .umap-multiplechoice.by3 {
```
jschleic, line 357: I couldn't find where `.third` is actually used - maybe it could be dropped?
```suggestion
.umap-multiplechoice.by3 {
```

yohanboniface:
Sounds good, thanks :)

PR #1358, opened 2023-10-05
Heatmap improvements

yohanboniface:
A bunch of improvement to make heatmap layers more usable:

- fix issues in Range input
- use a Range for heatmap radius
- use patch from https://github.com/Leaflet/Leaflet.heat/pull/78

Before:

image

After:

image

### DIFF
--- a/umap/static/umap/js/umap.layer.js
+++ b/umap/static/umap/js/umap.layer.js
@@ -161,8 +161,11 @@ L.U.Layer.Heat = L.HeatLayer.extend({
       [
         'options.heat.radius',
         {
-          handler: 'BlurIntInput',
-          placeholder: L._('Heatmap radius'),
+          handler: 'Range',
+          min: 10,
+          max: 100,
+          step: 5,
+          label: L._('Heatmap radius'),
           helpText: L._('Override heatmap radius (default 25)'),
         },
       ],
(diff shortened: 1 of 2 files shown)


umap/static/umap/js/umap.layer.js
```diff
@@ -187,6 +182,84 @@ L.U.Layer.Heat = L.HeatLayer.extend({
     }
     this._updateOptions()
   },
+
+  _redraw: function () {
+    // Import patch from https://github.com/Leaflet/Leaflet.heat/pull/78
+    // Remove me when this get merged and released.
```
davidbgk, line 188: I'm really inclined to rewrite/modernize (for…of, etc) that part together and submitting a patch to the upstream PR but maybe it's intentional on the Leaflet side to keep a bigger set of supported browsers?

yohanboniface: I don't think it's intentional, maybe it's because it's an old patch (7 years old) ?

I'm ok to rewrite it in our side, but not interested to submitting a patch: I think the PR is quite dead now, not even sur the user who submitted is still around and would accept it. Leaflet.heat has no release since a very long time. I'd like to do one (as a Leaflet maintainer I've the merge bit there to), but it may take a bit of time to get there, since I need to coordinate with Mourner, which has other things in mind I think.

So, my recommendation would be to not touch the patch (so not to diverge), but I'm ok if you prefer modernizing it on our side. Then, next step would be to merge that PR upstream and at this point modernize the code. But it's an unconventional situation and I don't really see a clear virtuous way to proceed…

Issue #334, opened 2016-06-24
[wishlist] highlight active marker / polygon 


druxstr:
If your map has a lot of markers or polygons, highlighting the selected one could be helpful.
When using the sidepanel, you can't visually get which one is currently selected.

If your map has also multipoints or multipolygons, highlighting seems anoivable : 
otherwise you won't see if a specific element is part of a multi or not.

This could be done by adding a colored border to the element for instance.

MIAnalytics:
This is a really good idea, is this something that could be easily added?

yohanboniface:
Fixed in #1359

PR #1359, opened 2023-10-05
highlight selected feature

jschleic:
I'd like to suggest some more interactive highlighting of paths. What do you think about:

https://github.com/umap-project/umap/assets/903453/04babeeb-bee3-4950-b43f-6b1d842d9faa

A very simple implementation uses the popup events to increase opacity and weight.
On maps with multiple lines one can see much easier, where the line starts and ends.
This implementation also highlights the lines of polygons.

Do you think, such a feature is feasible?
Should parameters be configurable?


yohanboniface:
Nice proposal, thanks! :+1:  

I don't think we need this to be configurable at this stage.

I wonder if we can release this without having an highlight for markers to (remember that the "popup" can be a side panel, o no directly associated to the related feature), and in this case how to deal with those, maybe touching the shadow ? I guess the answer will be different for each type of marker (may be done in CSS, after adding a class in JS).

For polygons, maybe also increase fill opacity ? In some situations polygons may have no border.

cc @Aurelie-Jallut @davidbgk 

cf #334

Aurelie-Jallut:
In consultation I want to differentiate the point / zone / line activated.e
- enlarge markers 
- Add contour to line
- Add contour to polygon (and possibly increase opacity)

jschleic:
Thank you very much for the quick feedback!

I don't see how an outline in a different color around a path could be implemented. Except adding the same path again behind the first one.

Regarding markers: Does the following mock-up go into the direction you were thinking of @Aurelie-Jallut ?

https://github.com/umap-project/umap/assets/903453/63036c2b-583e-46f5-9c81-879ea6edd9ae

I will continue on this as time allows...

yohanboniface:
> I don't see how an outline in a different color around a path could be implemented. Except adding the same path again behind the first one.

Increasing weight is fine!

> Regarding markers: Does the following mock-up go into the direction you were thinking of @Aurelie-Jallut ?

Sounds good to me! But usually markers have full opacity, so I think we'll need increasing the shape for all, or maybe playing with the shadow (bigger / changing color / moving it away from the marker so it appears like raised…)?

Thanks again for your work, much appreciated :)

umap/static/umap/js/umap.features.js
```diff
@@ -822,13 +822,15 @@ L.U.PathMixin = {
 
   highlightPath: function () {
     this.parentClass.prototype.setStyle.call(this, {
+      fillOpacity: Math.sqrt(this.getDynamicOption('fillOpacity')),
       opacity: 1.0,
       weight: 1.3 * this.getDynamicOption('weight'),
     })
   },
 
   resetPath: function () {
```
yohanboniface, line 831: Have you tried calling `redraw` instead of a custom function here ?

davidbgk:
> Sounds good to me! But usually markers have full opacity, so I think we'll need increasing the shape for all, or maybe playing with the shadow (bigger / changing color / moving it away from the marker so it appears like raised…)?

Maybe we could reduce the opacity of all other paths/points when only one is selected/hovered to make it stand out?

yohanboniface:
> Maybe we could reduce the opacity of all other paths/points when only one is selected/hovered to make it stand out?

This can be a lot of paths/points, not sure we want this extra computation :/

jschleic:
One issue is still open:

After adding a marker in edit mode and switching back to display mode the marker highlight doesn't work any longer. The `highlight` callback is called and the css class `umap-icon-active` is added in the debugger but not in DOM?! Any ideas @yohanboniface ?
grafik
grafik

jschleic:
> After adding a marker in edit mode and switching back to display mode the marker highlight doesn't work any longer. 

The culprit seems to be `this.map.removeLayer(this.layer)` triggered during `save()`:
https://github.com/umap-project/umap/blob/219b3e17dd4c37142614805bdc2dae7d9565f962/umap/static/umap/js/umap.layer.js#L1525
A manual `hide() show()` cycle in the layer box has the same effect.

Thus I have to keep the `highlight` callback in the marker class.

yohanboniface:
Thanks a lot! I'm travelling now, I'll have a look during the week-end :)

yohanboniface:
So, I've had a look and I'm fine with the logic :)

I'm not totally convinced by the rendering part. I'm making a summary of what's actually implemented in the PR, so maybe @Aurelie-Jallut can have an overview and help us on the finishing part :)

Path:

Sounds good to me!

[Screencast from 2023-10-21 12-20-05.webm](https://github.com/umap-project/umap/assets/146023/29193880-608b-4ac0-abed-c0c68a4e4735)

Polygon:

Sounds good to me!

[Screencast from 2023-10-21 12-20-28.webm](https://github.com/umap-project/umap/assets/146023/2b5b8eb3-c439-4997-81b2-555b0edd88c8)

Circle:

Sounds good to me!

[Screencast from 2023-10-21 12-20-46.webm](https://github.com/umap-project/umap/assets/146023/f9227b31-198e-4618-b24f-5914494c02da)

Default:

Shadow becomes white, which does not work on this white background.

[Screencast from 2023-10-21 12-23-33.webm](https://github.com/umap-project/umap/assets/146023/dc55762f-0ac6-4bd1-95bf-a587daefffd8)

I've tried to play with multiple shadow, but anything I can do is damn ugly, eg.:

[Screencast from 2023-10-22 10-37-05.webm](https://github.com/umap-project/umap/assets/146023/90ce168b-ad5f-4ded-ad73-f3d48f25b31f)

Drop:

Same white shadow issue.

[Screencast from 2023-10-21 12-22-35.webm](https://github.com/umap-project/umap/assets/146023/e1f5b775-03a4-424a-97eb-25d43007e96f)

Pin:

The ball grows but it becomes a bit square. If we get to a good working shadow highlight for Default and Drop, maybe we could use the same here, and there will be no need for increasing the ball size ?

[Screencast from 2023-10-21 12-21-47.webm](https://github.com/umap-project/umap/assets/146023/eefa4eab-811e-4d61-96c6-ef06be8897a0)

jschleic:
Square'ish *Pin* radius fixed, thanks!

What about following the "larger"-strategy for *Drop* and *Default* as well? That would look like:

https://github.com/umap-project/umap/assets/903453/9f9de770-1db4-4858-b755-8ce968501de4

For *Default* and *Drop* markers I would suggest to change to svg `path` from the composite `container`+`arrow` anyways. Semi-opaque markers look composite with circle+arrow. An svg would then allow an easier outline border around the marker as @Aurelie-Jallut suggested in https://github.com/umap-project/umap/pull/1359#issuecomment-1750232087 - but that's imho a separate PR.

Another option is to decrease the default marker opacity. To make highlighted markers with `opacity: 1` stand out. That would of course only work for new maps.

P.S. Are there possibly old maps out there without any opacity set (neither in the map options nor the path)?? The [test case](https://github.com/umap-project/umap/blob/219b3e17dd4c37142614805bdc2dae7d9565f962/umap/static/umap/test/_pre.js#L260-L277) data has such `null` polygon opacity resulting in an invisible polygon on click with this PR applied - is that an issue we should add an `if(opacity)`?

yohanboniface:
> What about following the "larger"-strategy for Drop and Default as well?

Makes sense! Let me check with Aurelie before putting more energy. Hopefully this Friday (we had a big event last week, that explains our time for synchronizing and giving you useful feedback, and I'm sorry for that).

> For Default and Drop markers I would suggest to change to svg path from the composite container+arrow anyways. 

Why not, but a few questions:

- could we add a shadow on a path ?
- could we easily integrate the pictogram/letter ?
- what about rendering performance (this HTML/CSS version has quite poor performance, and we should makes this better at some point, hence my question :) )

> P.S. Are there possibly old maps out there without any opacity set

Honestly I'm not sure about this point, so I'd say let's be more defensive just in case :)

Thanks again for your work!

yohanboniface:
@jschleic I just discussed with Aurélie and she suggests we should use same pattern for all, which is increasing the shape, so:

- path: increase the weight and opacity, as you did
- polygon: same
- circle: same
- pin: increase the ball and the shadow (but without changing the colour)
- drop: increase the shape, and shadow, and opacity when it's not already 1
- default: same

Does that make sense for you ? Do you still have time and energy to continue the work ? I think we are really close the final point, but feel free!

Thanks again :)

jschleic:
Sounds good!

I will finish that next week.

yohanboniface:
> to change to svg path from the composite container+arrow anyways.

Just found this, so pasting it here for later use: http://iatkin.github.io/leaflet-svgicon/

umap/static/umap/js/umap.core.js
```diff
@@ -161,7 +161,7 @@ L.Util.detectFileType = (f) => {
 }
 
 L.Util.usableOption = (options, option) =>
-  options[option] !== undefined && options[option] !== ''
```
jschleic, line 164: To make `getDynamicOption()` return the default I had to exclude `null` here. I tested with the test case data from `_pre.js`. Is there any other special meaning for `null`, that should be `usable`?

yohanboniface: Yes, in all ternary fields, like https://github.com/umap-project/umap/blob/master/umap/static/umap/js/umap.forms.js#L796-L802

yohanboniface:
Another discussion about using SVG markers that may be of interest: https://github.com/Leaflet/Leaflet/pull/8766

dkbg:
Thanks for implementing the feature. One issue I've noticed is that it can often be difficult to differentiate between highlighted and non-highlighted markers. I came here searching for whether this had been implemented, I thought that the version of uMap I was using didn't have the feature, not realising that the highlighting feature was there, just not at all obvious.

Here's a screenshot for reference. One of these markers is highlighted, but it isn't obvious which one.
image

jschleic:
@dkbg Thanks for your feedback!

As discussed above, adding a colored countour,  would imho require implementation of svg markers. That would be a rather big efford to re-implement all the marker classes. Any help welcome :-)

For now you could try to reduce the default *icon opacity* to something like 0.6 - that makes the active marker stand out like in the video above: https://github.com/umap-project/umap/pull/1359#issuecomment-1751369723

clwacker:
Thanks for the tip. 
But changing the opacity has little effect when you choose circles to represent the points on the map.

PR #1362, opened 2023-10-06
Allow to hide a datalayer from the caption list

yohanboniface:
In some situations, we don't want all datalayers to be displayed in the caption list, for example because the layer only contains "non meaningful" data: aesthetics one, or technical, or administrative…

Not sure about the property name…


davidbgk (approved):
Naming looks explicit enough to me 👍

PR #1375, opened 2023-10-24
Fix max not loading when defaultView=latest and datalayer has no data

### LINKED ISSUE #1374: Can't load and edit a map
**Describe the bug**
After configuring layer-specific editing permissions, the map can no longer be edited. Several errors are displayed in the console log.

**To Reproduce**
1. Configure a layer-specific editing permissions.
2. Save and exit
3. Edit that map back

**Screenshots**
Error log, while logged in : 
image

Error log, while logged out :
image

https://umap.openstreetmap.fr/en/map/poster-caleg_952507

yohanboniface:
When the default datalayer has no data, we need a fallback

fix #1374

### DIFF
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -697,6 +697,7 @@ L.U.Map.include({
         const datalayer = this.defaultDataLayer(),
           feature = datalayer.getFeatureByIndex(-1)
         if (feature) feature.zoomTo()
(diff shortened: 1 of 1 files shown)


davidbgk (approved):
I don't get how it prevents all errors reported from #1374 console.

yohanboniface:
AFAIK, they are all related to the map not being initialized fully (no center/zoom).

PR #1381, opened 2023-10-26
Increase maxZoomLimit to 24

yohanboniface:
Even if this limit is not a hard limit, just a warning, we now have been proven some maps using custom backgrounds use zooms greater than 20.

Eg. this maps which allows zoom until 23:

https://umap.openstreetmap.fr/fr/map/cimetiere-asnieres-la-giraud_716488


davidbgk (approved):
🍂

PR #1393, opened 2023-11-07
Add triangle icon to form fieldset to make toggle clearer

yohanboniface:
[Screencast from 2023-11-07 17-52-14.webm](https://github.com/umap-project/umap/assets/146023/74fc71bc-8a39-4887-813b-aa66ca45e800)


umap/static/umap/img/source/24-white.svg
davidbgk, line 1: Do we still need the change on the `24` sizes?

davidbgk: New Inkscape version leads to new rounded values 😅

davidbgk (approved):

PR #1395, opened 2023-11-08
Refactor icon selector: use tabs, make options more explicit

yohanboniface:
image

TODO:
- [x] integration tests
- [x] make it open on the correct tab when a value is already set
- [ ] do not display preview icon if value is inherited
- [x] good highlight and selected effects
- [x] refactor symbol search
- [x]  decide if we keep the automagic "switch to white" feature (cf #1387)


davidbgk (approved):

yohanboniface:
>  do not display preview icon if value is inherited

That would need a refactor of how inheritance works, so I'll let it for another step

ttooff:
Hi, since this update, the text font size in the icon has increased. Can that be reverted, or better, configured?

yohanboniface:
Yes, see https://github.com/umap-project/umap/pull/1434

Thanks for raising the point!

PR #1396, opened 2023-11-08
Full map download endpoint

davidbgk:
The premise of an API.

TODO:

* [x] permissions
* [x] make the JS use that view


yohanboniface:
:rocket:

PR #1401, opened 2023-11-10
Set a default favicon

davidbgk:
It was more complicated than expected.

Resources:
* https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs
* https://adamj.eu/tech/2022/01/18/how-to-add-a-favicon-to-your-django-site/


umap/urls.py
```diff
@@ -183,7 +183,15 @@
     re_path(r"^user/(?P<identifier>.+)/$", views.user_maps, name="user_maps"),
     re_path(r"", include(i18n_urls)),
 )
-urlpatterns += (path("stats/", cache_page(60 * 60)(views.stats), name="stats"),)
+urlpatterns += (
```
yohanboniface, line 186: Given those are static files, can't we let nginx serve them ?

davidbgk: We can put these routes before they hit Django if we control the instance but can we expect every deployment rely on nginx and admin will have the correct skills x rights to edit the configuration?

yohanboniface: I'd say it's a matter of documentation, we already have a bunch of Nginx configuration in the "Ubuntu" install process. But Can't we make that those file live under `/static/` so it automatically works ?

davidbgk: From what I understand, the favicon has to be at the root of the server `/favicon.ico` https://en.wikipedia.org/wiki/Favicon

yohanboniface: Ok, that's annoying. I thought that adding those meta would allow us to customize the path…

In this case, should we put all those files in a `/static/favicon/` subfolder, and then document a simple rule in nginx ?

I'm ok to add those URLs/view in Django, but I'm just trying to challenge a less intrusive scenario. :)

davidbgk: After discussion:

1. make a redirection to favicon.ico only + document the route to add in nginx
2. serve all others in static/favicons
3. create a view for the webmanifest to declare paths related to the static_url

PR #1408, opened 2023-11-15
Fix accordion arrow not displayed on Chromium

yohanboniface:
Screenshot from 2023-11-15 16-51-25
Screenshot from 2023-11-15 16-51-16


davidbgk (approved):
What about details/summary? 😈

yohanboniface:
> What about details/summary? 😈

Yep, let's do that, but in another move :)

PR #1409, opened 2023-11-15
More targeted CSS rule for button color

yohanboniface:
Otherwise it override forms button (which are in dark mode and thus need to keep the clear grey color)

image

image


yohanboniface:
Button and link in the share box does not have the exact same font rendering, but that's over my CSS skill and energy to fix that…

image

davidbgk (approved):

PR #1410, opened 2023-11-15
Add tests for favicon-related views

davidbgk:


umap/views.py
```diff
@@ -422,7 +422,7 @@ def _urls_for_js(urls=None):
 
 
 def simple_json_response(**kwargs):
-    return HttpResponse(json.dumps(kwargs))
+    return HttpResponse(json.dumps(kwargs), content_type="application/json")
```
davidbgk, line 425: @yohanboniface I made a PR just for that line, does it look OK to you?

yohanboniface: It does!

yohanboniface (approved):

PR #1413, opened 2023-11-18
Integrate with Github CI

almet:
- [x] Create Github Actions to run the tests
- [x] Make github actions use postgres
- [x] Make Playwright work
- [x] Support for python 3.12
- [x] Format the files 
- [x] Make the linter happy


almet:
I don't understand why the CI is not using the proper database configuration.

image

shows that it uses the "postgres" database, whereas I'm defining this otherwise in `tests/settings.py`:

```python

if os.environ.get("GITHUB_ACTIONS", False) == "true":
    DATABASES = {
        "default": {
            "ENGINE": "django.contrib.gis.db.backends.postgis",
            "NAME": "umap",
            "USER": "umap",
            "PORT": 5432,
            "PASSWORD": "umap",
        }
    }
```

almet:
The error is now, trying to fix it by specifying the `HOST`.
```
psycopg2.OperationalError: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
```

almet:
Now upgrading the postgres version to 12, as its the minimum required version for Django.

almet:
Now installing playwright :-)

almet:
Playwright is failing with stuff like:

```
______________________ test_map_load_with_owner[chromium] ______________________

anonymap = <Map: test map>
live_server = <LiveServer listening at http://localhost:50485>
owner_session = <Page url='http://localhost:50485/en/map/test-map_141'>

    def test_map_load_with_owner(anonymap, live_server, owner_session):
        owner_session.goto(f"{live_server.url}{anonymap.get_absolute_url()}")
        map_el = owner_session.locator("#map")
        expect(map_el).to_be_visible()
        enable = owner_session.get_by_role("button", name="Edit")
>       expect(enable).to_be_visible()
E       AssertionError: Locator expected to be visible
E       Actual value: None 
E       Call log:
E       LocatorAssertions.to_be_visible with timeout 5000ms
E       waiting for get_by_role("button", name="Edit")
```

yohanboniface:
> Playwright is failing with stuff like:

May be that the HTML is not the map page one, but a 404 or anything like that ?

.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,59 @@
...
+      if: matrix.dependencies == 'minimal'
+    - name: Install dependencies
+      run: |
+        sudo apt update
+        sudo apt install libgdal-dev
+        python -m pip install --upgrade pip
+        python -m pip install -e ".[dev,test]"
+        playwright install --with-deps
+    - name: run tests
+      run: pytest .
```
yohanboniface, line 50: I'd use `-vv` to have greater details and the test name in the list ? Maybe even use `make test` (which is `-vx`, but why not) ?

almet: Failing fast would help, I think, and verbose means more info. always good. Let's do this :-)

pyproject.toml
```diff
@@ -47,17 +45,20 @@ dependencies = [
...
   "djlint==1.31.0",
   "mkdocs==1.5.2",
+  "vermin==1.5.2",
+  "pymdown-extensions==10.4",
+  "isort==5.12",
 ]
 test = [
   "factory-boy==3.2.1",
-  "playwright==1.38.0",
+  "playwright>=1.39,<2",
```
almet, line 57: This is needed to be able to support python 3.12.

pyproject.toml
```diff
@@ -23,16 +23,14 @@ classifiers = [
...
-  "Programming Language :: Python :: 3.7",
   "Programming Language :: Python :: 3.8",
   "Programming Language :: Python :: 3.9",
   "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
 ]
 dependencies = [
-  "Django>=4.1",
+  "Django>=4.2,<5",
```
almet, line 33: The test suite fails otherwise with Django 4.1 (see #1426)

pyproject.toml
```diff
@@ -47,17 +45,20 @@ dependencies = [
...
 test = [
   "factory-boy==3.2.1",
-  "playwright==1.38.0",
+  "playwright>=1.39,<2",
   "pytest==6.2.5",
   "pytest-django==4.5.2",
-  "pytest-playwright==0.4.2",
-
+  "pytest-playwright>=0.4.3,<1",
+  "pytest-xdist>=3.5.0,<4",
```
almet, line 61: This is useful to speedup the tests.

umap/tests/settings.py
```diff
@@ -5,15 +5,17 @@
 SECRET_KEY = "justfortests"
 COMPRESS_ENABLED = False
 FROM_EMAIL = "test@test.org"
-EMAIL_BACKEND = 'django.core.mail.backends.locmem.EmailBackend'
+EMAIL_BACKEND = "django.core.mail.backends.locmem.EmailBackend"
 
-if "TRAVIS" in os.environ:
+if os.environ.get("GITHUB_ACTIONS", False) == "true":
```
almet, line 10: Configuration for Github Actions.

umap/tests/test_views.py
```diff
@@ -1,17 +1,18 @@
 import json
 import socket
-from datetime import date, timedelta
+from datetime import date, datetime, timedelta
```
almet, line 3: This fixes warning about datetime that should be time aware.

umap/settings/base.py
```diff
@@ -244,9 +244,7 @@
 UMAP_MAPS_PER_SEARCH = 25
 UMAP_MAPS_PER_PAGE_OWNER = 10
 UMAP_SEARCH_CONFIGURATION = "simple"
-UMAP_FEEDBACK_LINK = (
-    "https://wiki.openstreetmap.org/wiki/UMap#Feedback_and_help"  # noqa
-)
+UMAP_FEEDBACK_LINK = "https://wiki.openstreetmap.org/wiki/UMap#Feedback_and_help"  # noqa
```
davidbgk, line 247: Is the `# noqa` still relevant?

almet: fixed

davidbgk (approved):
🎁 🙌

Issue #1414, opened 2023-11-21
Improvements on database


anayrat:
Hello, 
As discussed with @yohanboniface.
I spotted several slow queries:

# Slow count

```
2023-11-21 08:15:40.126 UTC [3450782-1]: db=umap,user=umap,app=[unknown],client=[local],xid=0 LOG:  duration: 1484.497 ms  statement: SELECT COUNT(*) AS "__count" FROM "umap_map" WHERE ("umap_map"."share_status" = 1 AND ST_Distance("umap_map"."center", ST_GeogFromWKB('\x0101000020e610000000000000000000400000000000804940'::bytea)) > 1000.0 AND NOT ("umap_map"."id" = 1))
```
Here is the execution plan :
```
 Aggregate  (cost=171354.78..171354.79 rows=1 width=8) (actual time=276.758..276.759 rows=1 loops=1)
   Output: count(*)
   Buffers: shared hit=497515 dirtied=8
   ->  Index Scan using leaflet_storage_map_optim on public.umap_map  (cost=0.42..170633.90 rows=288351 width=0) (actual time=0.013..250.208 rows=518021 loops=1)
         Filter: (umap_map.id <> 1)
         Rows Removed by Filter: 1
         Buffers: shared hit=497515 dirtied=8
 Query Identifier: -5852652082130588930
 Planning:
   Buffers: shared hit=7
 Planning Time: 0.198 ms
 Execution Time: 276.788 ms
(13 rows)
```

(Is is faster than first query due to Postgres tuning (shared_buffers increased) ).
As we can see, this query count more than 500K rows very often. Postgres have to no choice to both read the index and the table due to this filter `NOT ("umap_map"."id" = 1)`.

We can improve with ways:

- Remove the filter, thus, postgres could use existing index `"leaflet_storage_map_optim" btree (modified_at) WHERE share_status = 1 AND st_distance(center, st_geomfromewkt('SRID=4326;POINT(2 51)'::text)::geography) > 1000.0::double precision` 
- Add the condition in a partial index if the condition is always the same : `"leaflet_storage_map_optim" btree (modified_at) WHERE share_status = 1 AND st_distance(center, st_geomfromewkt('SRID=4326;POINT(2 51)'::text)::geography) > 1000.0::double precision AND  NOT ("umap_map"."id" = 1)`
[...]

PR #1415, opened 2023-11-21
Do not exclude DEMO and SHOWCASE maps anymore from home

yohanboniface:
Those maps can be excluded by setting their share_status=OPEN

cf #1414


davidbgk (approved):
I didn't find any documentation relative to these two settings, is that intentional?

yohanboniface:
Sort of :p

Those are "non assumed features", so undocumented :p

PR #1422, opened 2023-11-23
Restore missing buttons in caption bar

yohanboniface:
Those button, being width: 100% and display: block were not displayed AND were pushing out also the slideshow buttons.

Before:
image

After:

image


davidbgk (approved):
Oups :(

PR #1434, opened 2023-11-26
Reset increasing icon text size

yohanboniface:
There are some usage with 5 chars that do not fit anymore. Let's reset for now and maybe introduce later a way for the user to configure the size.

### In current master:
Screenshot from 2023-11-26 17-18-42

### With the reset in this PR:
Screenshot from 2023-11-26 17-18-31

See also https://lists.openstreetmap.org/pipermail/umap/2023-November/000545.html

### DIFF
--- a/umap/static/umap/map.css
+++ b/umap/static/umap/map.css
@@ -1307,7 +1307,6 @@ a.add-datalayer:hover,
     vertical-align: middle;
     color: white;
     font-weight: bold;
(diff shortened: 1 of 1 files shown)


ttooff:
@yohanboniface Thanks for the commit! Also nice to see you used one of my maps as example!

davidbgk (approved):

Issue #1363, opened 2023-10-07
Making description field available in the properties table of layers


iriman:
**Is your feature request related to a problem? Please describe.**
For polygons that consist of many points (like borders) it usually takes long and it also slows down the browser while loading the feature to edit properties of it. "Table of Properties" for layer (right-most icon) ![image](https://github.com/umap-project/umap/assets/8467903/90dcee60-7206-4f21-8991-080fb44b043b) works great for this to avoid extra load on browser but it doesn't have description field.

**Describe the solution you'd like**
Making description field available in the properties table. thanks!

yohanboniface:
I think going this way would mean replacing our home made table by a more powerful one, like:
- https://gridjs.io/
- https://www.datagridxl.com/
- https://sheetjs.com/
- https://www.ag-grid.com/
- https://tabulator.info/

It's on our radar, while not in the top priorities at the moment. But of course, PR are welcome! :)

yohanboniface:
It's now possible to edit description in the table editor.

PR #1438, opened 2023-11-27
Always use fullwidth table editor

yohanboniface:
Adding a property when table is not in fullwidth is buguish, so let's make it simple.

In the future, we'd replace this table editor with a dedicated lib.

cf #1363


davidbgk (approved):
👍

PR #1439, opened 2023-11-28
(WIP) Add Bubbles layer type

### LINKED ISSUE #1399: Add another type of data layer : proportional circles
The addition of cloropleth data layers triggers another idea : showing point data as proportional circles, i.e. circles whose area is proportional to the value of a numerical field. 

- Load point data in a layer, with a field that contains a numerical value e.g. population.
- Select "Proportional circles" as the type of layer
- Define the field used to calculate the area of each circle, and a factor to apply so that circle sizes can be adjusted
- Each point is displayed as a circle, and can be selected to show a popup.
- It may be necessary to adjust the size of circles to the zoom level, so that they do not hide each other at a low zoom level => need for another factor ?

**Additional context**
Example map : 
image

yohanboniface:
fix #1399

This is a very first look. But it's not as simple as I thought, because:
- as for a heatmap, we need a special representation (each feature should be a circle, not a normal icon nor a polygon)
- as for a choropleth, we want to dynamically take control over the rendering (we need all the features to be able to compute the bubbles sizes)
- as for a normal layer, we want all the features to react to interaction, specifically click to open a popup, but also tooltip, highlight (?), etc.

In this first implementation, I've created a new type of layer, that creates a CircleMarker for each feature, but doing so we lose all the interactions. I've tried to set a sort of proxy so that a click on the circle will fire a click on the feature, but as this feature is actually not on the map, this does not work naturally.
Also, at this point, the styling is only done once, so editing layer's style will not be updated live in the map.

A few more thinking is needed, mainly to decide how to handle the circle: should it be a "proxy layer" as now, or should we try to transform a feature on the fly (make sure it is a marker of type circle, but as for now our circle are html, not svg…), and in this case, should we magically handle polygons and polylines (taking their center), or only markers (which could be easier, and possibly enough) ?

image

### DIFF
--- a/umap/static/umap/js/umap.forms.js
+++ b/umap/static/umap/js/umap.forms.js
@@ -381,6 +381,7 @@ L.FormBuilder.LayerTypeChooser = L.FormBuilder.Select.extend({
     ['Cluster', L._('Clustered')],
     ['Heat', L._('Heatmap')],
     ['Choropleth', L._('Choropleth')],
+    ['Bubble', L._('Bubbles')],
   ],
 })
 
--- a/umap/static/umap/js/umap.layer.js
+++ b/umap/static/umap/js/umap.layer.js
@@ -116,6 +116,123 @@ L.U.Layer.Cluster = L.MarkerClusterGroup.extend({
   },
 })
 
+L.U.Layer.Bubble = L.FeatureGroup.extend({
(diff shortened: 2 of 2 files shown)


yohanboniface:
Replaced by #2053

PR #1445, opened 2023-12-02
Make it easier to open file dialog

yohanboniface:
Failed attempt to make it easier to open the file dialog for import:

- tried to add a keyboard shortcut, but it does not work in Firefox (it does in Chrom*, and while it seems it [should](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#detecting_cancellations))
- tried to focus the field and use autofocus, so once the import panel is open one can press Enter to open the dialog, but it does not work either

As I started working on this, I thought this was the opportunity to move the importer panel to a separate class, as we did for the browser panel.


umap/static/umap/js/umap.importer.js
```diff
@@ -0,0 +1,167 @@
...
+
+  open: function () {
+    if (!this.container) this.build()
+    this.fileInput.showPicker()
+    this.map.ui.openPanel({ data: { html: this.container }, className: 'dark' })
+  },
+
+  openFiles: function () {
+    this.open()
+    this.fileInput.click()
```
davidbgk, line 126: Even with [showPicker](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/showPicker), you'll not be able to open it with a keyboard shortcut because it requires a [user activation](https://developer.mozilla.org/en-US/docs/Web/Security/User_activation) and it doesn't count as one.

umap/static/umap/js/umap.core.js
```diff
@@ -426,6 +426,7 @@ L.U.Keys = {
   I: 73,
   L: 76,
   M: 77,
+  O: 79,
```
davidbgk, line 429: We could use the [standard mapping](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) instead of using our custom one.

umap/static/umap/js/umap.importer.js
davidbgk: Actually:

> [keydown](https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event) (except for the `Esc` key nor a shortcut key reserved by the user agent) — [Features gated by user activation](https://developer.mozilla.org/en-US/docs/Web/Security/User_activation)

And we do use that event but it doesn't look to work anyway 🤷 Maybe because it's a reserved shortcut?

davidbgk: One option might be to create [our own fake input](https://stackoverflow.com/a/52757538) that fills the dedicated/regular input once files are selected but it's ugly/hackish…

umap/static/umap/js/umap.core.js
yohanboniface: Good point! Let's do that in a separate PR ?

davidbgk (approved):

PR #1447, opened 2023-12-03
fix: no data in facet filters when set as default panel

k-3st3ban:
When facet search is set as the default panel, if using remote data, no filter data was appearing in the panel.


yohanboniface:
Makes sense, thanks! :)

PR #1449, opened 2023-12-05
Replace "Ctrl" by "Cmd" under macOS and style the "kbd" tag in keyboard shortcuts labels

yohanboniface:


umap/static/umap/js/umap.core.js
```diff
@@ -433,6 +433,64 @@ L.U.Keys = {
...
+    },
+  },
+
+  displayLabel: function (action, withKbdTag=true) {
+    let {shortcut, label} = this.SHORTCUTS[action]
+    if (withKbdTag) {
+      shortcut = shortcut.split('+').map((el) => `<kbd>${el}</kbd>`).join('+')
+    }
+    const modifier = this.isMacOS ? 'Cmd' : 'Ctrl'
+    label += ` ${shortcut.replace('Modifier', modifier)}`
```
davidbgk, line 490:
```suggestion
    label += ` (${shortcut.replace('Modifier', modifier)})`
```

I think keeping the keyboard shortcut between parenthesis is more readable.

davidbgk (commented):
Is that intentional not to put `kbd` tags in that case?

Capture d’écran, le 2023-12-11 à 15 11 13

When/if applied, the CSS line-height/padding for that tag needs to be reduced.

yohanboniface:
> Is that intentional not to put `kbd` tags in that case?

Not intentional, but this box is derived from the buttons, and use the title element, which does not contain the tag of course. I did not put energy on changing that point, because I thought we'd totally change this box to have all the shortcuts, etc. But let's refactor to use something else than the title, you're right.

umap/static/umap/js/umap.core.js
yohanboniface: Once the `kbd` tag is styled, I find the parenthesis noisy:

image


vs 

image

What do you think ?

davidbgk (approved):

PR #1450, opened 2023-12-06
[chore] fix docs requirements

almet:
I didn't catch that docs has a separated requirements.txt file for readthedocs.


almet:
Seems to build correctly, see https://readthedocs.org/projects/umap-project/builds/22766696/

PR #1451, opened 2023-12-06
[ci] Add a separate target for docs

almet:
This adds a separate target for docs, to ensure that the docs will be build the same way readthedocs does it.


almet:
This works as intended: the CI was failing (detecting the docs couldn't build) prior to #1450, and is now passing with it.

PR #1453, opened 2023-12-06
Better changelog rendering

davidbgk:
Before:

Capture d’écran, le 2023-12-06 à 12 46 54

After:

Capture d’écran, le 2023-12-06 à 12 47 00

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,133 +2,133 @@
 
 ## 1.11.1 - 2023-11-27
 
-* Reset increasing icon text size by @yohanboniface in https://github.com/umap-project/umap/pull/1434
-* But back blue links in popup and small credit by @yohanboniface in https://github.com/umap-project/umap/pull/1425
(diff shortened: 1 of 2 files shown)


davidbgk:
Documentation: https://facelessuser.github.io/pymdown-extensions/extensions/magiclink/

yohanboniface:
Can it be even more magical ? :)

I mean, those full links are those coming from a copy-paste from the autogenerated changelog in Github, could it be that magic link or something else would display them in the short form automagically ? Otherwise, it means we need to change those links manually, no ? Which would not be ideal :(

davidbgk:
> I mean, those full links are those coming from a copy-paste from the autogenerated changelog in Github, could it be that magic link or something else would display them in the short form automagically ? Otherwise, it means we need to change those links manually, no ? Which would not be ideal :(

🪄 we can copy-paste the rendered HTML and it looks to work! 💖

almet:
Looks good to me 👍

yohanboniface (approved):

PR #1454, opened 2023-12-07
Refactor Share & Download UI for better usability

jschleic:

Users of our map often complain they didn't find the gpx download option. We embed a umap on our club website. But having to click the share button to download, then scroll down and find the `geojson` dropdown to select `gpx` seems to be not really intuitive.

We'd like to have an intuitive, easy-to-use download option of gpx files.

* Splitting the gpx / kml download option into a "download" dialog with download-icon would probably be most intuitive. But further increasing the number of buttons on the left hand side and an essentially empty dialog seems not the way to go.
* Add gpx output to the new server API could be an option, so that we could place a link to the gpx-download in a more prominent place inside custom maps (e.g. sidebar text or node popup). May be still desirable for special use-cases but imho a cleaner share-dialog is useful for everyone.

While working at the PR I've further unified the different other dialogs.

**This PR proposes:**
* Tooltips for all the buttons now _match_ the headline on top of the dialogs (with #1449 the headings should probably also be filled from the new constants)
* Re-order the share and download formats from simple use cases (share link to friends) to complex ones (embed iframe into custom website)
* directly show available formats as buttons instead of hiding them in a dropdown
* Add a label explaining the actions a bit more
* Finetuned CSS to fix dark-theme hangovers

pr-download

### DIFF
--- a/umap/static/umap/base.css
+++ b/umap/static/umap/base.css
@@ -162,13 +162,15 @@ textarea {
     padding: 7px;
 }
 select {
+    border: 1px solid #222;
     width: 100%;
     height: 28px;
     line-height: 28px;
-    color: #efefef;
-    border: 1px solid #222;
-    background-color: #393F3F;
     margin-top: 5px;
+}
+.dark select {
+    color: #efefef;
+    background-color: #393F3F;
 }
 select[multiple="multiple"] {
     height: auto;
@@ -225,7 +227,8 @@ button.flat,
     font-size: 10px;
     border-radius: 0 2px;
 }
-.content .helptext {
+.content .helptext,
+#umap-ui-container .help-text {
     background-color: #eee;
     color: #000;
 }
@@ -294,15 +297,18 @@ input[type="file"] + .error {
     text-align: left;
     display: block;
     cursor: pointer;
-    background-color: #232729;
+    background-color: #eee;
     height: 30px;
     line-height: 30px;
-    color: #fff;
     margin: 0;
     font-family: fira_sans;
     font-weight: normal;
     font-size: 1.2em;
     padding: 0 5px;
+}
+.dark .fieldset.toggle .legend {
+    background-color: #232729;
+    color: #fff;
 }
 .fieldset.toggle .legend:before {
     background-repeat: no-repeat;
@@ -311,10 +317,13 @@ input[type="file"] + .error {
     width: 24px;
     line-height: 24px;
     display: inline-block;
-    background-image: url('./img/16-white.svg');
+    background-image: url('./img/16.svg');
     vertical-align: bottom;
     content: " ";
     background-position: -144px -76px;
+}
+.dark .fieldset.toggle .legend:before {
+    background-image: url('./img/16-white.svg');
 }
 .fieldset.toggle.on .legend:before {
     background-position: -144px -51px;
@@ -401,6 +410,7 @@ input.switch:checked ~ label:after {
 .umap-multiplechoice.by5 {
     grid-template-columns: 1fr 1fr 1fr;
 }
+.button-bar.by4,
 .umap-multiplechoice.by4 {
     grid-template-columns: 1fr 1fr 1fr 1fr;
 }
@@ -448,7 +458,8 @@ input.switch:checked ~ label:after {
 }
 .umap-field-iconUrl .action-button,
 .inheritable .define,
-.inheritable .undefine {
+.inheritable .undefine,
+.copy-button {
     float: right;
     width: initial;
     min-height: 18px;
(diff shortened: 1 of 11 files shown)


davidbgk:
@jschleic this is something we should better document but the source of svg icons are in `umap/static/umap/img/source/*.svg` which are editable with inkscape and then `make icons` will generate the final optimized ones 😉

It would (hopefully) avoid such a big diff.

davidbgk:
Thanks for your PR!

>  We'd like to have an intuitive, easy-to-use download option of gpx files.

I wonder if a dedicated URL, like we did for the full (umap) backup and as you suggested, would be pertinent for your use-case to be able to make your own custom buttons. In the meantime, I appreciate that 4-buttons version! Both are possible though and could be part of a two-step PRocess.

yohanboniface:
cc @Aurelie-Jallut could you have a look ?

:+1: @jschleic Thanks a lot for this :) Indeed, this dialog needs a refactor. Would you mind doing a first separate PR with the technical part and things unrelated to the new UI proposal ? So we can merge it quickly and concentrate then on the UI/UX part.

Also, what do you think of moving this panel to a dedicated class as we did recently for [browser](https://github.com/umap-project/umap/blob/master/umap/static/umap/js/umap.browser.js) and [importer](https://github.com/umap-project/umap/blob/master/umap/static/umap/js/umap.importer.js) ?

I really like your proposal, but I'd like Aurélie (the UX of the team) to have a look to! So I'd say lets first focus on non UI/UX changes, so we can then focus on a smaller diff. What do you think ?

jschleic:
Thanks for the quick response! I've separated out the headline / title / svg stuff which concerns the consistency of all dialogs (also UI).
Waiting for feedback from @Aurelie-Jallut here; I will rebase after #1455 is merged.

jschleic:
> Also, what do you think of moving this panel to a dedicated class as we did recently for [browser](https://github.com/umap-project/umap/blob/master/umap/static/umap/js/umap.browser.js) and [importer](https://github.com/umap-project/umap/blob/master/umap/static/umap/js/umap.importer.js) ?

Good suggestion. I moved the `renderShareBox` code into a separate file together with helpers I think are related.

yohanboniface:
Thanks!

One related issue I've been reported recently: when no SHORT_URL is set, then there is not "simple" share link: only the iframe and the link that includes all the query string.

I'll mention this pull to @Aurelie-Jallut. We need her review to proceed, I'd say.

Aurelie-Jallut:
Hi @jschleic 
Thank you very much for this  !
I completely agree on the usefulness of using the same headings on the buttons and in the title. 

In relation to your proposal, I think we should still group Share a link and integrate an iframe with the card settings functions. But if we add a tab system, we'll be able to see the download links in different formats more directly.
I've added a screenshot of a low-definition model to illustrate my idea.
What do you think?

For the Download section, we can directly add the links with the Download title, which still directs the user to the area.

(At a later stage, we could also look at the contents of the setting box ;) )

image

yohanboniface:
>  I've added a screenshot of a low-definition model to illustrate my idea.

Like it!
Two points:
- we need to keep a message on the download part saying "Only active layers will be downloaded" (or something like this)
- I'd keep the "settings" only in the "Embed" part, so to have:
    - in the "share" part: only the base link of the map, shortened if there is a short domain, otherwise normal link
    - in the "embed" part: both the iframe code and the direct link of the iframe (so including the query string resulting from the settings), and of course the settings form

Thoughts ?

Aurelie-Jallut:
Only active layers will be downloaded => doesn't this concern Full backup?

If the parameters also concern the basic map link, it seems important to me to display it, doesn't it?

Aurelie-Jallut:
image

Aurelie-Jallut:
if it doesn't make sense to use export / download, we  can just keep one of the two : 
image

yohanboniface:
> if it doesn't make sense to use export / download, we can just keep one of the two

Even better! :)

davidbgk:
This new interface might be a good candidate to pursue the experimentation of #1461 😇

yohanboniface:
Very nice!

If you don't want to deal with the tabs, you can just put one box below the other as a first step. And we'll see later, if you have more time/energy, or one of us.

Anyway thanks already for this work :)

yohanboniface:
One small suggestion: instead of repeating "data" for each format, add it in the help label "Only visible layers' data" or something like this ?

jschleic:
> If you don't want to deal with the tabs, you can just put one box below the other as a first step. And we'll see later, if you have more time/energy, or one of us.

Great! So let's leave it as single page without tabs for the moment.

yohanboniface:
I'm about to merge but one question/suggestion: what about putting the copy button in the same line of the related input here ?

image

I think using flex should work, and allow translations of the word "copy" to be longer/shorter without troubles. Otherwise an icon may do the trick.

I can also merge like this and do this enhancement in a later step!

yohanboniface:
I'll take it! Thanks for your work @jschleic :)

PR #1461, opened 2023-12-12
Template-based importer (vs. using L.DomUtil)

davidbgk:
After a discussion with @almet we realized that it would be possible to write HTML in templates instead of building everything on the JS side using `L.DomUtil`, this is a first small step to verify that it's both possible and pertinent.

There are still some rough edges but let's discuss the direction in itself first!

### DIFF
--- a/umap/static/umap/js/modules/global.js
+++ b/umap/static/umap/js/modules/global.js
@@ -1,10 +1,20 @@
 import * as L from '../../vendors/leaflet/leaflet-src.esm.js'
 import URLs from './urls.js'
 import Browser from './browser.js'
+import Importer from './importer.js'
 import { Request, ServerRequest, RequestError, HTTPError, NOKError } from './request.js'
 // Import modules and export them to the global scope.
 // For the not yet module-compatible JS out there.
 
 // Copy the leaflet module, it's expected by leaflet plugins to be writeable.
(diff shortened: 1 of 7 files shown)


yohanboniface:
Templates are tempting, but I feel like until we need all those `querySelector` the final benefit is small, while it also adds a small indirection / complexity (declaring HTML in some file, and using it in another).
Also, given Leaflet itself is programmatic, this means introducing another pattern alongside.
I liked your [experiment](http://tmp.larlet.fr/leaflet-map/) much as a direction, even if of course the step would be much bigger.
At this stage, the direction is not totally clear to me.

davidbgk:
> At this stage, the direction is not totally clear to me.

We might want to go a step further and turn that form into a web component in itself which is able to do its stuff (`select` build from data, checks, auto-selected fields on upload) on itself prior to dispatch a custom event with the result to uMap's core JS.

I'd love to give it a try 😇

yohanboniface:
Sounds appealing! But what would be the browser support ?

davidbgk:
[Custom elements](https://caniuse.com/custom-elementsv1) are not fully baseline compliant :'( But it's coming!

davidbgk:
Done in a different way since then.

PR #1463, opened 2023-12-13
Add support for JS modules (+module for URLs handling)

almet:
This is a test for how we could use JS modules in uMap.

Because modules are not used everywhere, we need to bind the loaded modules to `window` fr now. Also, the mocha test runner needs to be run inside a server, to avoid CORS limitations for local file systems.

I'm not really sure it's the direction we should take, but I hope it could bring good discussions :-)


umap/static/umap/js/bindModules.mjs
```diff
@@ -0,0 +1,3 @@
+import URLs from "./modules/urls.mjs"
+
+window.URLs = URLs;
```
almet, line 3: We have to bind the loaded modules to something in order to access it on the other side.

umap/templates/umap/map_init.html
```diff
@@ -1,7 +1,9 @@
 {% load umap_tags %}
 <div id="map"></div>
 <!-- djlint:off -->
-<script type="text/javascript">
-  let MAP = new L.U.Map("map", {{ map_settings|notag|safe }})
+<script defer type="text/javascript">
+  window.addEventListener('load', (event) => {
```
almet, line 5: Had to add this, otherwise the objects wouldn't be exported to `window`.

umap/static/umap/js/modules/urls.mjs
```diff
@@ -0,0 +1,44 @@
...
+        if (this.urls.hasOwnProperty(urlName)) {
+            console.log("urls", this.urls, urlName, params);
+            return L.Util.template(this.urls[urlName], params);
+        } else {
+            throw `Unable to find a URL for route ${urlName}`;
+        }
+    }
+
+    // Update if map_id is passed, create otherwise.
+    map_save({ map_id }) {
```
almet, line 29: Just figured out this won't work as the object might be null.

umap/static/umap/js/umap.js
```diff
@@ -1960,17 +1940,17 @@ L.U.Map.include({
...
         lat: e.latlng.lat,
         lng: e.latlng.lng,
         locale: L.locale,
         zoom: this.getZoom(),
       }
-      window.open(L.Util.template(url, params))
-    }
-    return
+    );
+    if (url)
```
yohanboniface, line 1952: We had a discussion about this simple `if` case with @davidbgk. We did not ended with a clear consensus, but I made this proposal:
- if the `if` is followed by a single instruction AND prettier keeps it on a single line, let's use a one liner without the curly bracket
- other wise put use curly brackets and carriage return

umap/static/umap/js/modules/urls.mjs
```diff
@@ -0,0 +1,38 @@
...
+        if (typeof params !== 'object')
+            params = { map_id: params };
+        if (params.map_id)
+            return this.get("map_update", params.map_id);
+        return this.get("map_create");
+    }
+
+    // Update the layer if pk is passed, create otherwise.
+    datalayer_save(params) {
+        if (typeof params !== 'object')
```
yohanboniface, line 32: Why not always passing params as an object ?

umap/static/umap/js/modules/urls.mjs
```diff
@@ -0,0 +1,38 @@
...
+            throw `Unable to find a URL for route ${urlName}`;
+        }
+    }
+
+    // Update if map_id is passed, create otherwise.
+    map_save(params) {
+        if (typeof params !== 'object')
+            params = { map_id: params };
+        if (params.map_id)
+            return this.get("map_update", params.map_id);
```
yohanboniface, line 26: Why not just passing `params` here ?

almet: ✅

umap/static/umap/js/modules/urls.mjs
almet: It allows to make the calls easier.

```urls.get('route_name', map_id)``` rather than ```urls.get('route_name', {map_id: map_id})```

File: umap/static/umap/js/umap.js (hunk 4 above)
Comment (almet) [reply umap/static/umap/js/umap.js:1952]: Thanks for mentioning it :-)

File: umap/static/umap/js/bindModules.mjs (hunk 1 above)
Comment (vinyll) [reply umap/static/umap/js/bindModules.mjs:3]: missing newline? ⏎

Not use about `.mjs` vs `.js` but a reading is here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#aside_%E2%80%94_.mjs_versus_.js

File: umap/static/umap/js/modules/urls.mjs [hunk 7]
```diff
@@ -0,0 +1,38 @@
...
+
+    get(urlName, params) {
+        if (typeof params !== 'object')
+            params = { map_id: params };
+
+        if (typeof this[urlName] === "function")
+            return this[urlName](params);
+
+        if (this.urls.hasOwnProperty(urlName)) {
+            console.log("urls", this.urls, urlName, params);
```
vinyll: is this a debug command? If so it could rather be `console.debug`

umap/static/umap/js/modules/urls.mjs
```diff
@@ -0,0 +1,40 @@
+export default class URLs {
+    constructor(serverUrls) {
+        this.urls = serverUrls;
+    }
+
+    get(urlName, params) {
+        if (typeof params !== 'object')
+            params = { map_id: params };
```
vinyll: I guess you want to ensure that the `params` is an object with a `map_id` property.
`typeof []` also return 'object'.
`if(!params.map_id)` would seem more robust.
However this is not very clear to me and I may be misunderstanding

umap/static/umap/js/modules/urls.mjs
vinyll: This piece of check seems to be here and there.
Would it make sense to have a uniformizing function such as

```
function getMapObject(idOrObject) {
  if (!idOrObject.map_id)
    return { map_id: idOrObject }
  else
    return idOrObject
}
```

Then calling it whenever we need to check about the `params` object|string

umap/static/umap/js/modules/urls.mjs
```diff
@@ -0,0 +1,40 @@
...
+
+    // Update the layer if pk is passed, create otherwise.
+    datalayer_save(params) {
+        if (typeof params !== 'object')
+            params = { map_id: params };
+        if (params.pk)
+            return this.get("datalayer_update", params);
+        return this.get("datalayer_create", params);
+    }
+}
```
vinyll, line 30: missing newline ⏎

umap/static/umap/js/umap.js
```diff
@@ -90,6 +90,7 @@ L.U.Map.include({
 
     // After calling parent initialize, as we are doing initCenter our-selves
     if (geojson.geometry) this.options.center = this.latLng(geojson.geometry)
+    this.urls = new window.URLs(this.options.urls)
```
vinyll, line 93: Where does the `options` object comes from and how can we ensure the `urls` property exists?
Same question for other properties of `options`

umap/static/umap/js/umap.js
```diff
@@ -272,10 +273,10 @@ L.U.Map.include({
         url.searchParams.delete('edit')
         history.pushState({}, '', url)
       }
-      if (L.Util.queryString('download'))
-        window.location = L.Util.template(this.options.urls.map_download, {
-          map_id: this.options.umap_id,
-        })
+      if (L.Util.queryString('download')) {
+        let download_url = this.urls.get('map_download', { map_id: this.options.umap_id })
```
vinyll, line 277: It seems not to be means to change and should then be `const` instead of `let`

umap/static/umap/js/modules/urls.mjs
almet: I guess that's because I'm trying to mimic how python works. I've made changes to always pass an object.

umap/static/umap/js/umap.js
```diff
@@ -352,7 +353,7 @@ L.U.Map.include({
       document.body,
       'umap-caption-bar-enabled',
       this.options.captionBar ||
-        (this.options.slideshow && this.options.slideshow.active)
+      (this.options.slideshow && this.options.slideshow.active)
```
vinyll, line 356: what is the `slideshow` object and how can we make sure it has the `active` property?

umap/static/umap/js/umap.js
```diff
@@ -1169,42 +1170,25 @@ L.U.Map.include({
...
   star: function () {
     if (!this.options.umap_id)
       return this.ui.alert({
         content: L._('Please save the map first'),
         level: 'error',
       })
-    let url = L.Util.template(this.options.urls.map_star, {
-      map_id: this.options.umap_id,
-    })
+    let url = this.urls.get('map_star', { map_id: this.options.umap_id })
```
vinyll, line 1191: `const`

umap/static/umap/js/umap.js
```diff
@@ -1960,17 +1940,13 @@ L.U.Map.include({
...
-      const params = {
-        lat: e.latlng.lat,
-        lng: e.latlng.lng,
-        locale: L.locale,
-        zoom: this.getZoom(),
-      }
-      window.open(L.Util.template(url, params))
-    }
-    return
+    let url = this.urls.get('routing', {
```
vinyll, line 1943: `const`

umap/static/umap/js/umap.layer.js
```diff
@@ -1531,8 +1514,9 @@ L.U.DataLayer = L.Evented.extend({
     formData.append('settings', JSON.stringify(this.options))
     // Filename support is shaky, don't do it for now.
     const blob = new Blob([JSON.stringify(geojson)], { type: 'application/json' })
-    formData.append('geojson', blob)
-    this.map.post(this.getSaveUrl(), {
+    formData.append('geojson', blob);
+    let saveUrl = this.map.urls.get("datalayer_save", { map_id: this.map.options.umap_id, pk: this.umap_id })
```
vinyll, line 1518: `const`

umap/static/umap/test/URLs.mjs
```diff
@@ -0,0 +1,48 @@
+import URLs from "../js/modules/urls.mjs";
+
+describe('URLs', () => {
```
vinyll, line 3: Quotes are sometimes single or double and I'm not sure what rule applies in your code

umap/static/umap/test/URLs.mjs
```diff
@@ -0,0 +1,48 @@
...
+    // Mock server URLs that will be used for testing
+    const mockServerUrls = {
+        'map_create': '/maps/create',
+        'map_update': '/maps/{map_id}/update',
+        'datalayer_create': '/maps/{map_id}/datalayers/create',
+        'datalayer_update': '/maps/{map_id}/datalayers/{pk}/update'
+    };
+
+    let urls = new URLs(mockServerUrls);
```
vinyll, line 12: `const`

umap/templates/umap/content.html
```diff
@@ -37,7 +37,7 @@
 {% endblock content %}
 {% block bottom_js %}
   {{ block.super }}
-  <script type="text/javascript">
+  <script defer type="text/javascript">
```
almet, line 40: Will remove this, shouldn't be useful.

umap/static/umap/test/URLs.mjs
```diff
@@ -0,0 +1,48 @@
...
+    describe('datalayer_save()', () => {
+        it('should return the create URL if no pk is provided', () => {
+            expect(urls.datalayer_save({ map_id: '123' })).to.be.equal('/maps/123/datalayers/create');
+        });
+
+        it('should return the update URL if a pk is provided', () => {
+            expect(urls.datalayer_save({ map_id: '123', pk: '456' })).to.be.equal('/maps/123/datalayers/456/update');
+        });
+    });
+});
```
vinyll, line 48: newline ⏎

vinyll (commented):
Just to pretend I know 🤓

yohanboniface:
Great work, thanks!
I'm fine with the direction and the implementation. Though, for this to be merged, I've two concerns:
- it seems to me that we are loading twice the Leaflet code, I think this is a no go (but I may miss something)
- the event listener on "load" seems weird to me, I wonder which impacts this can have and if there isn't a "better" way to deal with the initial issue

About loading Leaflet twice, I guess the only way to prevent this is to make a full switch, and loading only the esm version. So that would mean either:
- changing all Leaflet usage to switch to importing module
- reexposing "manually" a global L with all the needed property, until the full switch is done

Given the later is more gradual and we use a lot of plugins we do not control, I'd be in favor of the second option.

Thoughts ?

yohanboniface:
WIP of plugin status:

plugin|importable|last commit|maintainers
--|--|--|--
[leaflet-contextmenu](https://github.com/aratcliffe/Leaflet.contextmenu/)|no|4 years ago| others
[leaflet-editable](https://github.com/Leaflet/Leaflet.Editable/)|no|last year| Leaflet
[leaflet-editinosm](https://github.com/yohanboniface/Leaflet.EditInOSM/)|no|5 years ago| us
[leaflet-formbuilder](https://github.com/yohanboniface/Leaflet.FormBuilder)|no|last month| us
[leaflet-fullscreen](https://github.com/Leaflet/Leaflet.fullscreen)|no|5 years ago| Leaflet
[leaflet-hash](https://github.com/mlevans/leaflet-hash)|no|10 years ago| others
[leaflet-i18n](https://github.com/umap-project/Leaflet.i18n)|no|4 months ago| us
[leaflet-iconlayers](https://github.com/ScanEx/Leaflet-IconLayers)|no|5 years ago| others
[leaflet-loading](https://github.com/ebrelsford/Leaflet.loading)|no|2 years ago| others
[leaflet-measurable](https://github.com/umap-project/Leaflet.Measurable)|no|8 months ago| us
[leaflet-minimap](https://github.com/Norkart/Leaflet-MiniMap)|no|4 years ago| others
[leaflet-toolbar](https://github.com/Leaflet/Leaflet.toolbar)|no|5 years ago| Leaflet
[leaflet.heat](https://github.com/Leaflet/Leaflet.heat)|no|2 months ago| Leaflet
[leaflet.locatecontrol](https://github.com/domoritz/leaflet-locatecontrol)|no|2 months ago| others
[leaflet.markercluster](https://github.com/Leaflet/Leaflet.markercluster)|no|7 months ago| Leaflet
[leaflet.path.drag](https://github.com/Leaflet/Path.Drag.js/)|no|6 years ago| Leaflet
[leaflet.photon](https://github.com/komoot/leaflet.photon)|no|5 days ago| us

yohanboniface:
This are the `L.` properties expected by uMap code (not the plugins; some properties of the list do not come from Leaflet but from some plugins OR from uMap itself):

```
{'L.Bounds',
 'L.Browser',
 'L.Control',
 'L.DomEvent',
 'L.DomUtil',
 'L.Editable',
 'L.FeatureGroup',
 'L.FormBuilder',
 'L.GeoJSON',
 'L.GeoUtil',
 'L.HeatLayer',
 'L.Icon',
 'L.LatLng',
 'L.LineUtil',
 'L.Map',
 'L.Marker',
 'L.MarkerCluster',
 'L.MarkerClusterGroup',
 'L.MeasureControl',
 'L.Path',
 'L.PhotonSearch',
 'L.Point',
 'L.Polygon',
 'L.Polyline',
 'L.Popup',
 'L.TileLayer',
 'L.Toolbar',
 'L.ToolbarAction',
 'L.U',
 'L.UmapSingleton',
 'L.Util',
 'L._',
 'L.bind',
 'L.control',
 'L.createObjectURL',
 'L.extend',
 'L.lang',
 'L.latLng',
 'L.latLngBounds',
 'L.locale',
 'L.setLocale',
 'L.setOptions',
 'L.stamp',
 'L.svg'}
```

yohanboniface:
### Properties set by plugins
#### Leaflet.MarkerCluster
- `L.DistanceGrid`
- `L.MarkerCluster`
- `L.MarkerClusterGroup`
- `L.MarkerClusterNonAnimated`
- `L.QuickHull`
- `L.markerClusterGroup`
#### Leaflet.Editable
- `L.Editable`
#### Leaflet.FormBuilder
- `L.FormBuilder`
#### Leaflet.Hash
- `L.Hash`
- `L.hash`
#### Leaflet.Heat
- `L.HeatLayer`
#### Leaflet.Measurable
- `L.Measurable`
- `L.MeasureControl`
- `L.MeasureLine`
- `L.MeasureVertex`
#### leaflet.path.drag
- `L.PathDraggable`
#### Leaflet.Photon
- `L.PhotonBase`
- `L.PhotonBaseSearch`
- `L.PhotonReverse`
- `L.PhotonSearch`
#### Leaflet.Toolbar
- `L.Toolbar`
- `L.ToolbarAction`
- `L.toolbar`
- `L.toolbarActio`
#### uMap
- `L.U`
- `L.UmapSingleton`
- `L.lang`
#### Leaflet.i18n
- `L._`
- `L.i18n`
- `L.locale`
- `L.locales`
- `L.registerLocale`
- `L.setLocale`


### Properties from Leaflet

```
{
 'L.Bounds',
 'L.Browser',
 'L.Canvas',
 'L.Circle',
 'L.Class',
 'L.Control',
 'L.DivIcon',
 'L.DomEvent',
 'L.DomUtil',
 'L.Draggable',
 'L.Evented',
 'L.FeatureGroup',
 'L.GeoJSON',
 'L.GeoUtil',
 'L.Handler',
 'L.Icon',
 'L.LatLng',
 'L.LatLngBounds',
 'L.LayerGroup',
 'L.LineUtil',
 'L.Map',
 'L.Marker',
 'L.Mixin',
 'L.Path',
 'L.Point',
 'L.Polygon',
 'L.Polyline',
 'L.Popup',
 'L.Projection',
 'L.Rectangle',
 'L.TileLayer',
 'L.Util',
 'L.bind',
 'L.control',
 'L.extend',
 'L.featureGroup',
 'L.latLng',
 'L.latLngBounds',
 'L.map',
 'L.point',
 'L.polyline',
 'L.rectangle',
 'L.setOptions',
 'L.stamp',
 'L.svg',
 }

**Comment by yohanboniface** [comment]
It is not that simple: a `type=module` script is loaded asynchronously, so it cannot expose the `L.` (and properties) before classic script will need it.

File: umap/static/umap/js/modules/urls.js [hunk 20]
```diff
@@ -0,0 +1,29 @@
+import { Util } from './vendors.js'
```
yohanboniface, line 1: I'm not fan of this import, as we lose the origin of the `Util` variable in the process.

I'd suggest, either:
- loading `Util` (and such) from `path/to/leaflet.esm.js` everywhere
- exposing namespace in vendors, not variables; so one would type `import L from './vendors` and then use `L.Util`, just like in classic mode

umap/templates/umap/map_init.html
yohanboniface: As far as I understand:
- `defer` will not work for local script (i.e. without a `src` attribute)
- module type scripts are executed in defer mode, which means after the DOM is ready

I guess this explain why you needed the `load` event here. Maybe we could use [`DomContentLoaded`](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event) instead ?

One other option to consider (but not for this PR I'd say), is to have a dedicated script (`/map/xxxx/load.js`), which could then make it work in defer mode, and could be an elegant way for other people to include a uMap map in their HTML ?

umap/static/umap/js/modules/urls.js
almet: Yep, I like the second one better, and I agree 👍

umap/templates/umap/map_init.html
almet: Interesting, `defer` is not useful here then. `DomContentLoaded` seems the way (if it works), and will allow us to run the script without having to wait for other resources to be ready (images, etc).

yohanboniface (approved):

davidbgk (commented):
Did we check that it still works fine with embedded maps?

umap/templates/umap/map_init.html
```diff
@@ -1,7 +1,9 @@
 {% load umap_tags %}
 <div id="map"></div>
 <!-- djlint:off -->
-<script type="text/javascript">
-  let MAP = new L.U.Map("map", {{ map_settings|notag|safe }})
+<script defer type="text/javascript">
+  window.addEventListener('DOMContentLoaded', (event) => {
+    let MAP = new L.U.Map("map", {{ map_settings|notag|safe }});
```
vinyll, line 6: What is the scope of this?
Why `let`?
Does it become a global variable?

umap/templates/umap/map_init.html
```diff
@@ -1,7 +1,9 @@
...
+  });
```
vinyll, line 7: Does it make sense to have it twice in map_fragment and map_init?

davidbgk: Solved with #1516

yohanboniface:
It seems we need to better escape the settings (`"` in the description that makes the HTML invalid):

```
<umap-fragment data-settings="{&quot;type&quot;: &quot;Feature&quot;, &quot;geometry&quot;: {&quot;type&quot;: &quot;Point&quot;, &quot;coordinates&quot;: [6.173458099365235, 49.114866157159355]}, &quot;properties&quot;: {&quot;name&quot;: &quot;Diaporama des arbres remarquables \u00e0 Metz&quot;, &quot;zoom&quot;: 14, &quot;easing&quot;: false, &quot;zoomTo&quot;: 19, &quot;licence&quot;: &quot;&quot;, &quot;miniMap&quot;: false, &quot;overlay&quot;: null, &quot;showLabel&quot;: true, &quot;slideshow&quot;: {}, &quot;tilelayer&quot;: {&quot;tms&quot;: false, &quot;name&quot;: &quot;OSM-eu&quot;, &quot;maxZoom&quot;: 18, &quot;minZoom&quot;: 0, &quot;attribution&quot;: &quot;Map data ODbL © OSM contributors / tiles CC-by-SA © OSM-FR&quot;, &quot;url_template&quot;: &quot;https://tile.openstreetmap.bzh/eu/{z}/{x}/{y}.png&quot;}, &quot;captionBar&quot;: false, &quot;longCredit&quot;: &quot;&quot;, &quot;popupShape&quot;: &quot;Default&quot;, &quot;description&quot;: &quot;Cette carte en cours de r\u00e9alisation int\u00e8gre \n- Le fichier des arbres remarquables d\u00e9pos\u00e9 par Metz M\u00e9tropole en open data sur le portail [[https://www.data.gouv.fr/fr/organizations/metz-metropole/|data.gouv.fr]]\n- Des images des [[https://archives.metz.fr|Archives de la Ville de Metz]]\n(autorisation sans utilisation commerciale)\n- Des photographies et films personnels\n\n- Cette carte est pr\u00e9sent\u00e9e sur \n[[https://insidemap.eu/faire-sa-carte-diaporama-darbre-en-arbre-a-metz/|Insidemap.eu]]\n- Pour suivre sa fabrication, avec vinber et cartograf, voir l" excellent="" forum="" d'osm="" :\n[[http:="" forum.openstreetmap.fr="" viewtopic.php?t="7094|Tuto" de="" diaporama="" pour="" umap="" ?]]\n\n*seul="" un="" calque="" est="" compl\u00e9t\u00e9*="" :="" celui="" les="" arbres="" remarquables="" jardin="" des="" r\u00e9gates="" -="" boufflers="" et="" square="" giraud.="" en="" jaune="" pas="" encore="" photographie="" ni="" film.="" \n\n**un="" sujet="" d'actualit\u00e9**="" \nenvoy\u00e9="" sp\u00e9cial.="" le="" secret="" \n{{{https:="" youtube.com="" embed="" eh6rnaqspto}}}\n\n\n\n\n",="" "limitbounds":="" {},="" "morecontrol":="" true,="" "onloadpanel":="" "none",="" "zoomcontrol":="" "captionmenus":="" "embedcontrol":="" "scalecontrol":="" "searchcontrol":="" "measurecontrol":="" "scrollwheelzoom":="" false,="" "datalayerscontrol":="" "fullscreencontrol":="" "tilelayerscontrol":="" "displaypopupfooter":="" "popupcontenttemplate":="" "#="" {name}\n{description}",="" "permanentcreditbackground":="" "tilelayers":="" [{"id":="" 1,="" "name":="" "positron",="" "url_template":="" "https:="" cartodb-basemaps-{s}.global.ssl.fastly.net="" light_all="" {z}="" {x}="" {y}.png",="" "minzoom":="" 0,="" "maxzoom":="" 18,="" "attribution":="" "&copy;="" [[http:="" www.openstreetmap.org="" copyright|openstreetmap]]="" contributors,="" &copy;="" [[https:="" carto.com="" attributions|carto]]",="" "rank":="" "tms":="" false}],="" "datalayers":="" [{"name":="" "calque="" 1",="" "editmode":="" "disabled",="" "browsable":="" "incaption":="" "displayonload":="" "id":="" 42,="" "permissions":="" {"edit_status":="" 0}},="" {"name":="" "c="" botanique",="" "color":="" "cadetblue",="" "iconclass":="" "drop",="" "popupshape":="" "panel",="" "remotedata":="" 43,="" "b="" epars="" (\u00e0="" d\u00e9sactiver)",="" 44,="" (en="" cours)",="" 45,="" "a="" citadelle="" hypercentre",="" "showlabel":="" 46,="" 0}}],="" "urls":="" {"map_short_url":="" "="" m="" {pk}="" ",="" "ajax-proxy":="" ajax-proxy="" "password_change":="" change-password="" "password_change_done":="" change-password-done="" "map_download":="" map="" {map_id}="" download="" "stats":="" stats="" "login":="" fr="" login="" "login_popup_end":="" popup="" end="" "logout":="" logout="" "map_geojson":="" geojson="" "map_anonymous_edit_url":="" anonymous-edit="" {signature}",="" "pictogram_list_json":="" pictogram="" json="" "datalayer_view":="" datalayer="" "datalayer_versions":="" versions="" "datalayer_version":="" {name}",="" "map":="" {slug}_{map_id}",="" "map_new":="" new="" "map_create":="" create="" "map_star":="" star="" "user_dashboard":="" me",="" "user_profile":="" me="" profile",="" "map_update":="" update="" settings="" "map_update_permissions":="" permissions="" "map_attach_owner":="" owner="" "map_delete":="" delete="" "map_clone":="" clone="" "datalayer_create":="" "datalayer_delete":="" "datalayer_permissions":="" "map_send_edit_link":="" send-edit-link="" "datalayer_update":="" "routing":="" "http:="" directions?engine="osrm_car&amp;route={lat},{lng}&amp;locale={locale}#map={zoom}/{lat}/{lng}&quot;," "ajax_proxy":="" ?url="{url}&amp;ttl={ttl}&quot;," "search":="" photon.komoot.io="" api="" ?"},="" "static_url":="" static="" "hash":="" "attributioncontrol":="" "umapattributioncontrol":="" "nocontrol":="" "umap_id":="" 10,="" "default_iconurl":="" img="" marker.png",="" "prefix":="" "",="" "page":="" "2"}}'="">
  <div id="map2_10" class="map_fragment"></div>
</umap-fragment>
```

umap/templates/umap/map_init.html
almet: This is a (previously introduced) hack to have access to MAP everywhere, which was broken with these changes. I've updated accordingly.

almet:
To me we're ready to introduce these changes!

yohanboniface:
:tada: :bomb: :rocket:

davidbgk:
Maybe we need a release with just that change 😅 😇

PR #1467, opened 2023-12-14
Use real tile URL in tests

yohanboniface:
It's easier to debug integration tests screenshots (otherwise background is only grey).


davidbgk (commented):
How does it affect tests duration? 🤔

yohanboniface:
Good question!

This branch: 212 passed in 51.13s
Master: 212 passed in 49.41s

Only integrations ones:

This branch: 47 passed in 47.04s
Master: 47 passed in 42.88s

So it may not be insignificant…

davidbgk (approved):

PR #1471, opened 2023-12-15
[docs] Add some testing information

almet:


davidbgk (approved):
❤️

Makefile
```diff
@@ -73,10 +73,8 @@ vendors:
 	npm run vendors
 installjs:
 	npm install
-testjsfx:
```
yohanboniface, line 76: Not sure why you removed this ?

davidbgk: This is not cross-platform

almet: It's because, with the ES modules you need a browser to run. Instead of using firefox directly, my proposal here is to print an URL, and then you can open it the way you want.

Would that work for you?

almet:
But I'm sorry, I don't remember hitting the merge button here. Was waiting for @yohanboniface review first. Not sure when that happened 😳

Makefile
yohanboniface: Hehe, not a big deal :)

I remember seeing that `open path/to/index.html` would be cross platform, no ? 

At least, I do use `make testjsfx`, so I'd prefer if it still works :)

davidbgk: As far as I understand, it has been merged with `testjs` so it still works somehow but maybe the `mocha-phantomjs` call was still useful?

almet: Ah, if `open` works that would be perfect, but I think it's not cross-platform. Does it work on your OS?

yohanboniface: In mine, yes :)

almet: > maybe the mocha-phantomjs call was still useful?

It wasn't working for me. If it's useful for somebody I can put it back of course :-)

yohanboniface: No, that one is broken, it can be removed for sure

almet: See #1472

PR #1472, opened 2023-12-16
build: Update makefile testjs to open the browser

almet:
Following the discussion on #1471, here are proposed changes to the `Makefile`.

- Runs a subprocess, and setup a trap to ensure ctrl+c can stop the server
- Uses [`python -m webbrowser`](https://docs.python.org/3/library/webbrowser.html) to open a new tab (cross compatible)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
 .DEFAULT_GOAL := help
 
+JS_TEST_URL := http://localhost:8001/umap/static/umap/test/index.html
+
 .PHONY: install
 install: ## Install the dependencies
 	python3 -m pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -74,7 +76,19 @@ vendors:
 installjs:
 	npm install
 testjs: node_modules
-	@echo "See the tests at http://localhost:8001/umap/static/umap/test/" && python -m http.server 8001
+	@{ \
```
yohanboniface, line 79: For my understanding, can you explain what that trap/wait is doing ? :)

yohanboniface (approved):

Makefile
almet: Sure,

[`trap`](https://man7.org/linux/man-pages/man1/trap.1p.html) will issue the `kill` command when receiving a signal (here a SIGINT).

[`wait`](https://man7.org/linux/man-pages/man1/wait.1p.html) waits for the job to complete.

To break things down:

- setups a callback with `trap`
- run the server in the background with `&`
- open the browser
- `wait` for the pid to be killed (which will be the case after it's killed with the trap)

yohanboniface: Ok thanks!
Not sure to understand who/what will send the SIGINT ?

almet: `CTRL + C` sends a `SIGINT` 😀 (and `CTRL + D` a `SIGTERM`)

davidbgk: It works on my env 👍

PR #1474, opened 2023-12-16
fix: accept long floats for latitude and longitude

almet:
Without "step: any", long floats are considered invalid (see the capture)

image

### DIFF
--- a/umap/static/umap/js/umap.features.js
+++ b/umap/static/umap/js/umap.features.js
@@ -680,8 +680,26 @@ L.U.Marker = L.Marker.extend({
   appendEditFieldsets: function (container) {
     L.U.FeatureMixin.appendEditFieldsets.call(this, container)
     const coordinatesOptions = [
-      ['_latlng.lat', { handler: 'FloatInput', label: L._('Latitude') }],
-      ['_latlng.lng', { handler: 'FloatInput', label: L._('Longitude') }],
+      [
+        '_latlng.lat',
+        {
+          handler: 'FloatInput',
+          label: L._('Latitude'),
+          step: 'any',
+          min: -90,
+          max: 90,
+        },
+      ],
+      [
+        '_latlng.lng',
+        {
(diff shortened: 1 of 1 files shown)


umap/static/umap/js/umap.features.js
```diff
@@ -680,8 +680,26 @@ L.U.Marker = L.Marker.extend({
...
-      ['_latlng.lat', { handler: 'FloatInput', label: L._('Latitude') }],
-      ['_latlng.lng', { handler: 'FloatInput', label: L._('Longitude') }],
+      [
+        '_latlng.lat',
+        {
+          handler: 'FloatInput',
+          label: L._('Latitude'),
+          step: 'any',
+          min: -90,
+          max: 90,
```
almet, line 690: I've added the max and min latitudes and longitudes here as well.

almet:
Just found out about https://github.com/umap-project/umap/commit/4936a97218528d2f75e0af0fd4ebe9187a95b605

Not sure what's going on here :-) Maybe the underlying `Leaflet.Formbuilder` doesn't merge `steps` correctly? (`FloatInput default `step` could be overwritten by the options we pass?`)

yohanboniface:
Humm, is that issue in prod or in local ? If in local, maybe you does not have an up to date node_modules ?

yohanboniface:
Sounds ok in prod, as far as I can see:

image

almet:
It's in local, but my modules are up to date. I'll try to reproduce on prod to see.

almet:
I do not reproduce on the OSMFR instance at least. Will investigate later.

almet:
> In that ecosystem, you cannot rely on updates, you need to rm -rf node_modules first 🙈

Ah, good to know, thanks 👍

almet:
I confirm this was due to the files not up to date on my side. Sorry for the noise (but I ended up learning a bit about `Leaflet.Formbuilder` along the way, so I'm happy 🤓

PR #1475, opened 2023-12-16
Iconlayers polish

yohanboniface:
Found some edge cases of the new tilelayer switcher this morning while testing it a bit more:
- when importing a map with a background not in the 10 first layers
- when choosing a tilelayer after the 10 first from the selector in edit mode

A bit of prettier noise…

### DIFF
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -166,7 +166,7 @@ L.U.Map.include({
     this.help = new L.U.Help(this)
 
     if (this.options.hash) this.addHash()
-    this.initTileLayers(this.options.tilelayers)
+    this.initTileLayers()
     // Needs tilelayer to exist for minimap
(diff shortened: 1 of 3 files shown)


davidbgk:
> A bit of prettier noise…

Maybe we should force it through the CI?

umap/static/umap/js/umap.js
```diff
@@ -649,8 +651,18 @@ L.U.Map.include({
   },
 
   eachTileLayer: function (callback, context) {
-    if (this.customTilelayer) callback.call(context, this.customTilelayer)
-    this.tilelayers.forEach((layer) => callback.call(context, layer))
+    const urls = []
+    const callOne = (layer) => {
```
davidbgk, line 655: 😉

yohanboniface:
> Maybe we should force it through the CI?

Yes!

Issue #1459, opened 2023-12-11
[suggestion] List maps as tile, with checkboxes to delete them all


Shohreh:
Hello,

Currently, listing and possibly deleting maps is slow, as the user must possibly scroll down several times to see them all.*

I suggest;
1. Display maps as a tile, each map with a checkbox next to it so as to delete several maps in one go
2. Only display Name and Last Save with option to sort each column. Fetching the maps takes too much time, and Title is good enough to know what it is — provide link to map in case the user wants to actually see it.

Cheers,

![image](https://github.com/umap-project/umap/assets/1541861/a851ffcc-d0fd-4733-9fd4-bc1b2039484d)

davidbgk:
@Shohreh what about adding a Delete option on the My maps page in the last column of the table?

<img width="1216" alt="Capture d’écran, le 2023-12-11 à 10 07 26" src="https://github.com/umap-project/umap/assets/3556/8c3fbc0c-1ca6-4756-9535-00a13634798d">

yohanboniface:
I see several points here:

- map miniature makes the page too slow => we'd like a preview image instead, but this is hard to do without adding a complex stack; Should we hide the miniature in the meantime ?
- Scrolling with a "more" button is not very friendly => should we have a normal paginator instead ? (prev 1 2 3 … next) ?
- 10 maps at a time may not be enough ?
- we'd like some mass action, with check boxes on maps, so to: delete / download many at once, cf #1430

Shohreh:
I guess hiding it before finding a better way does the job. The issue is that is that is slows down things a lot, while the user simply wants to delete a bunch of maps.

To select maps, I'd rather have them displayed all on one page instead of having to scroll or hit prev/next. For better performance, just displaying their title + date last updated is good enough.

A "delete/download" option would definitely be nice IMHO.

davidbgk:
> while the user simply wants to delete a bunch of maps.

I wonder if that's a popular use-case, adding some friction to such a destructive function might actually be a feature 🤔

yohanboniface:
Closing here, we should be good now!

Shohreh:
I don't see checkboxes on each line.

Is there a way to select several maps, and delete them all in one go?

![image](https://github.com/umap-project/umap/assets/1541861/ef4fea3b-6d28-416e-bb55-3111bfd7eabc)

davidbgk:
This is not a use case we consider to be common and this is a destructive one so we decided to keep the extra step.

PR #1478, opened 2023-12-19
Preview map only on click in user’s dashboard

davidbgk:
Refs #1459

This is a proposal to reduce the size/time for that view. The map is only loaded when the user clicks the Preview button.

Before:

Capture d’écran, le 2023-12-19 à 09 34 20

After:
Capture d’écran, le 2023-12-19 à 14 20 08

The preview opens in an almost full-screen dialog element with blurred background:

Capture d’écran, le 2023-12-19 à 14 24 06

Let's discuss :) Once accepted, I'll put styles and scripts to their regular places.

### DIFF
--- a/umap/models.py
+++ b/umap/models.py
@@ -1,4 +1,3 @@
-import json
 import os
 import time
 
@@ -12,6 +11,7 @@ from django.urls import reverse
 from django.utils.translation import gettext_lazy as _
 
 from .managers import PublicManager
+from .utils import _urls_for_js
 
 
 # Did not find a clean way to do this in Django
@@ -196,8 +196,6 @@ class Map(NamedModel):
 
     @property
     def preview_settings(self):
-        from .views import _urls_for_js
-
         layers = self.datalayer_set.all()
         datalayer_data = [c.metadata() for c in layers]
         map_settings = self.settings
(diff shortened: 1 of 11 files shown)


umap/models.py
```diff
@@ -193,6 +194,40 @@ class Map(NamedModel):
     objects = models.Manager()
     public = PublicManager()
 
+    @property
+    def unique_id(self):
+        return f"map_{self.pk}"
+
+    @property
+    def preview_settings(self):
+        from .views import _urls_for_js
```
yohanboniface, line 203: Move to some helper file ?

umap/models.py
```diff
@@ -193,6 +194,40 @@ class Map(NamedModel):
     objects = models.Manager()
     public = PublicManager()
 
+    @property
+    def unique_id(self):
+        return f"map_{self.pk}"
```
yohanboniface, line 199: This sounds like more related to HTML than models, shouldn't we keep `map_` inside the html and just use `pk` ?

yohanboniface (commented):
Sounds good!  :+1:

davidbgk:
Current status:

Capture d’écran, le 2023-12-27 à 13 30 51

PR #1479, opened 2023-12-20
Add a popup template to showcase OpenStreetMap data

yohanboniface:
image

It's a popup template that know how to display data coming from OpenStreetMap. At this stage, it's very basic and only display contact informations, but I'd like to push such a minimalist version and see how users use it before putting more energy on it.

What's not done and may be in further steps:
- display a i18n label (bakery, swimming pool)
- automatically display an icon if none is set on the layer
- show opening hours, or at least show if the shop/amenity is open
- show more specific tags like the cuisine for a restaurant, the organic tag, etc.
- connect to panoramax/mapilary to have a picture…

There is no intention to cover all and every possible case, but only to focus on the main tags (shop/amenity…).

### DIFF
--- a/umap/static/umap/js/umap.popup.js
+++ b/umap/static/umap/js/umap.popup.js
@@ -270,24 +270,9 @@ L.U.PopupTemplate.OSM = L.U.PopupTemplate.Default.extend({
     const color = this.feature.getDynamicOption('color')
     title.style.backgroundColor = color
     const iconUrl = this.feature.getDynamicOption('iconUrl')
-    let icon
-    if (
-      L.Util.isPath(iconUrl) ||
-      L.Util.isRemoteUrl(iconUrl) ||
-      L.Util.isDataImage(iconUrl)
-    ) {
-      icon = L.DomUtil.add('img', 'popup-icon', title)
-      icon.src = iconUrl
-    } else {
-      icon = L.DomUtil.add('span', 'popup-icon', title)
-      icon.textContent = iconUrl
-    }
-    if (L.DomUtil.contrastedColor(title, color)) {
-      if (L.Util.isPath(iconUrl) && iconUrl.endsWith('.svg')) {
-        icon.style.filter = 'invert(1)'
-      }
-      title.style.color = 'white'
-    }
+    let icon = L.U.Icon.makeIconElement(iconUrl, title)
+    L.U.Icon.setIconContrast(icon, title, iconUrl, color)
+    if (L.DomUtil.contrastedColor(title, color)) title.style.color = 'white'
     L.DomUtil.add('span', '', title, this.getName())
     const street = props['addr:street']
     if (street) {
--- a/umap/static/umap/js/umap.popup.js
+++ b/umap/static/umap/js/umap.popup.js
@@ -271,6 +271,7 @@ L.U.PopupTemplate.OSM = L.U.PopupTemplate.Default.extend({
     title.style.backgroundColor = color
     const iconUrl = this.feature.getDynamicOption('iconUrl')
     let icon = L.U.Icon.makeIconElement(iconUrl, title)
+    L.DomUtil.addClass(icon, 'icon')
     L.U.Icon.setIconContrast(icon, title, iconUrl, color)
     if (L.DomUtil.contrastedColor(title, color)) title.style.color = 'white'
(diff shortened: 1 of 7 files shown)


umap/static/umap/js/umap.popup.js
```diff
@@ -251,3 +251,70 @@ L.U.PopupTemplate.GeoRSSLink = L.U.PopupTemplate.Default.extend({
...
+    className: 'umap-openstreetmap',
+  },
+
+  getName: function () {
+    const props = this.feature.properties
+    if (L.locale && props[`name:${L.locale}`]) return props[`name:${L.locale}`]
+    return props.name
+  },
+
+  renderBody: function () {
```
davidbgk, line 266: Is the body rendered only when you click the marker?

davidbgk (commented):
I don't get where OSM proprs are actually fetched?

umap/static/umap/js/umap.popup.js
yohanboniface: Yes!

yohanboniface:
> I don't get where OSM proprs are actually fetched?

`props` are just the normal feature properties. This template is supposed to be used when the features comes from OSM, either by an import or a remote URL (eg. overpass).

umap/static/umap/js/umap.popup.js
```diff
@@ -251,3 +251,70 @@ L.U.PopupTemplate.GeoRSSLink = L.U.PopupTemplate.Default.extend({
...
+    const props = this.feature.properties
+    const container = L.DomUtil.add('div')
+    const title = L.DomUtil.add('h3', 'popup-title', container)
+    const color = this.feature.getDynamicOption('color')
+    title.style.backgroundColor = color
+    const iconUrl = this.feature.getDynamicOption('iconUrl')
+    let img
+    if (L.Util.isPath(iconUrl) || L.Util.isRemoteUrl(iconUrl) || L.Util.isDataImage(iconUrl)) {
+      img = L.DomUtil.add('img', 'popup-icon', title)
+      img.src = iconUrl
```
davidbgk, line 276: I wonder if the issue we found while pairing with images (re)downloaded for each marker is not related to the fact that we dynamically load URLs like this. Maybe using a cloned HTML template would help? It has to be tested though.

umap/static/umap/js/umap.popup.js
```diff
@@ -251,3 +251,70 @@ L.U.PopupTemplate.GeoRSSLink = L.U.PopupTemplate.Default.extend({
...
+    } else {
+      img = L.DomUtil.add('span', 'popup-icon', title)
```
davidbgk, line 278: Is it still an image?

umap/static/umap/js/umap.popup.js
```diff
@@ -251,3 +251,70 @@ L.U.PopupTemplate.GeoRSSLink = L.U.PopupTemplate.Default.extend({
...
+    if (props.mobile) {
+      L.DomUtil.add('div', '', container, L.DomUtil.element('a', {href: `tel:${props.mobile}`, textContent: props.mobile}))
+    }
+    const email = props.email || props['contact:email']
+    if (email) {
+      L.DomUtil.add('div', '', container, L.DomUtil.element('a', {href: `mailto:${email}`, textContent: email}))
+    }
+    const id = props['@id']
+    if (id) {
+      L.DomUtil.add('div', 'osm-link', container, L.DomUtil.element('a', {href: `https://www.openstreetmap.org/${id}`, textContent: L._('See on OpenStreetMap')}))
```
davidbgk, line 316: I'm pretty sure PrettierJS would not allow that long line 😉

umap/static/umap/js/umap.popup.js
yohanboniface: Renamed in `icon` :)

umap/static/umap/js/umap.popup.js
yohanboniface: Fixed, sorry!

umap/static/umap/js/umap.popup.js
yohanboniface: Maybe! But AFAIK we've seen this issue only when using the html2cancas. In normal mode, the img is fetched only once, for what I can see.

davidbgk (approved):

Issue #1001, opened 2022-06-02
Allow element position edition by GPS coordinates


brunetton:
I'm trying to manually place elements on a map with a good precision (not importing from geojson, but using UI)

I'm copying GPS coordinates (example `50.92350,2.11022`) from another online tool (https://www.geoportail.gouv.fr) and I'd like to be able to quickly paste them to a marker on map.

For now I've to separate longitude from latitude, and copy/paste them separately on the properties of my newly created marker:
![image](https://user-images.githubusercontent.com/86322/171655263-97146b69-7bfb-4489-8b6b-c079b0300786.png)

I'd like to have another field titled "GPS coordinates" that is the representation of the coordinates as GPS format (lat and log separated by a comma) that could be used to:
- quickly copy coordinates to standard format
- quickly paste coordinates from standard format to set the new marker position (from https://www.openstreetmap.org for example)

This is somehow related to #1000

yohanboniface:
This is possible by typing the coordinates in the search box and clicking on the pen:

![Image](https://github.com/user-attachments/assets/412840f8-8d4c-49f1-8655-8d34a3b6aaf3)

brunetton:
Thank you :)

PR #1480, opened 2023-12-21
Allow to type a latlng in the search box

### LINKED ISSUE #1000: Accept gps coordinates in search input
Hi

I'm using uMap for some projects (and thanks for this great tool !). But I really misses the ability to center the map to given coordinates. Entering GPS coordinates (example `48.85839, 2.29458`) isn't recognized:

image

yohanboniface:
When entering a string in a form of a latlng, we'll do two things:
- add a result that allow to center the map on this latlng (and to create a new point if edit mode is enabled)
- do a reverse search and display it result if any

image

image

fix #1000
cf #1001

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31,7 +31,7 @@
         "leaflet.locatecontrol": "^0.79.0",
         "leaflet.markercluster": "^1.5.3",
         "leaflet.path.drag": "0.0.6",
-        "leaflet.photon": "0.8.0",
+        "leaflet.photon": "0.9.0",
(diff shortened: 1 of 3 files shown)


davidbgk (approved):
This is one of these advanced feature that are hard to discover, maybe doable with an explicit label: `Search a place name (or coordinates)`?

davidbgk (approved):

brunetton:
Thank you ! Can't wait to use it in production 🥳

PR #1490, opened 2023-12-27
Be more explicit on changed fields when updating choropleth form

yohanboniface:
The `postUpdate` method of the Choropleth layer is called after any form field change, even if this field is not in the dedicated choropleth helper. So the previous check was too broad, and it would try to fetch the breaks input value on any form helper, which would fail if someone change any "non choropleth" property (like the colour)

### DIFF
--- a/umap/static/umap/js/umap.layer.js
+++ b/umap/static/umap/js/umap.layer.js
@@ -9,7 +9,7 @@ L.U.Layer = {
     return []
   },
 
-  postUpdate: function () {},
+  onEdit: function () {},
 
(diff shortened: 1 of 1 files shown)


davidbgk (approved):
Would it be pertinent to have a dedicated `postUpdateChoropleth`?

yohanboniface:
> Would it be pertinent to have a dedicated postUpdateChoropleth?

I had a look at this scenario, but I've seen that there is some case (actually one…) where a layer want to update itself after another property has been changed (`color` in this case), so I've preferred to keep the system generic.

Issue #1489, opened 2023-12-26
Support date properties in facet search


flammermann:
Currently, facet search only supports textual properties and allows filtering features by checking boxes with values for certain properties. Features with matching values in the respective property are shown, others are hidden.

I would like to extend that to also support date (continuous) properties.

Proposed changes:

- Change format in "Facet key" option from "\<key>|\<label>" to "\<key>|\<label>|\<type>"
- Possible types are "checkbox" and "date", default if not provided is "checkbox"
- If type is "checkbox" render elements for "Facet search" panel as usual
- If type is "date" render input fields for min and max
- Prepopulate the min and max fields with the minimum and maximum values available for the property across all features
- On update of min and/or max field, show/hide matching/not matching features

I already implemented an equivalent functionality for myself in version 1.3.7. Unfortunately, I missed to contribute it to the project and would now like to re-implement the functionality and contribute it straight away. ;)

Also, I already opened another feature request for numerical properties.

davidbgk:
Done in #1763

PR #1491, opened 2023-12-27
Support date and number properties in facet search

flammermann:
Implementation of https://github.com/umap-project/umap/issues/1489 and https://github.com/umap-project/umap/issues/1488


umap/static/umap/js/umap.forms.js
```diff
@@ -804,18 +804,67 @@ L.FormBuilder.FacetSearch = L.FormBuilder.Element.extend({
...
+    const min = this.options.criteria['min'];
+    const max = this.options.criteria['max'];
+
+    // Create labels for min and max inputs
+    this.minLabel = L.DomUtil.create('label', '', this.container);
+    this.minLabel.innerHTML = 'Start';
+    this.minLabel.htmlFor = `date_${this.name}_min`;
+
+    this.minInput = L.DomUtil.create('input', '', this.container);
+    this.minInput.type = 'datetime-local';
```
davidbgk, line 834: Nice, I didn't know about [that input type](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local) 👍

umap/static/umap/js/umap.controls.js
```diff
@@ -741,22 +741,50 @@ L.U.Map.include({
...
 
-    const knownValues = {}
+    const facetCriteria = {}
 
     keys.forEach((key) => {
-      knownValues[key] = []
-      if (!this.facets[key]) this.facets[key] = []
+      if (facetKeys[key]["type"] === "date") {
+        if (!facetCriteria[key]) facetCriteria[key] = {
+	  "min": undefined,
```
davidbgk, line 752: There are a couple of weird indentations, can you run PrettierJS against it?

flammermann: You are right, there are some tabs instead of spaces. Thanks for the hint, I will fix it.

umap/static/umap/js/umap.features.js
```diff
@@ -483,12 +483,30 @@ L.U.FeatureMixin = {
     return false
   },
 
+  parseDateField: function (value) {
+    if (parseFloat(value).toString() === value.toString()) {
```
yohanboniface, line 487: `value` may be undefined, which would make `parseFloat` to fail. As a general advice, given the data uMap will load can come from user inputs and from remote source, we have to be very defensive. The feature may not work, but the map should not crash, so the user can fix the data or settings.

umap/static/umap/js/umap.controls.js
```diff
@@ -741,21 +741,49 @@ L.U.Map.include({
...
+      }
     })
 
     this.eachBrowsableDataLayer((datalayer) => {
       datalayer.eachFeature((feature) => {
         keys.forEach((key) => {
           let value = feature.properties[key]
-          if (typeof value !== 'undefined' && !knownValues[key].includes(value)) {
-            knownValues[key].push(value)
+          if (facetKeys[key]["type"] === "date") {
```
yohanboniface, line 775: Should we distinguish between `date` and `datetime` ? I'm thinking about the form UI, but maybe also the parsing process.

umap/static/umap/js/umap.core.js
```diff
@@ -706,7 +706,7 @@ L.U.Help = L.Class.extend({
   slugKey: L._('The name of the property to use as feature unique identifier.'),
   filterKey: L._('Comma separated list of properties to use when filtering features'),
   facetKey: L._(
-    'Comma separated list of properties to use for facet search (eg.: mykey,otherkey). To control label, add it after a | (eg.: mykey|My Key,otherkey|Other Key)'
+    'Comma separated list of properties to use for facet search (eg.: mykey,otherkey). To control label, add it after a | (eg.: mykey|My Key,otherkey|Other Key). To control type, add it after another | (eg.: mykey|My Key|checkbox,otherkey|Other Key|date)'
```
yohanboniface, line 709: Let's make explicit the valid values for `type` ?

umap/static/umap/js/umap.forms.js
```diff
@@ -804,18 +804,67 @@ L.FormBuilder.FacetSearch = L.FormBuilder.Element.extend({
...
 
+L.FormBuilder.FacetSearchDate = L.FormBuilder.Element.extend({
+  build: function () {
+    this.container = L.DomUtil.create('div', 'umap-facet', this.parentNode);
+    const min = this.options.criteria['min'];
+    const max = this.options.criteria['max'];
+
+    // Create labels for min and max inputs
+    this.minLabel = L.DomUtil.create('label', '', this.container);
+    this.minLabel.innerHTML = 'Start';
```
yohanboniface, line 830: Maybe, instead of `Start`/`End` we could have `From` and `to`  in the same line ?

yohanboniface (commented):
Thanks a lot for this work!
I've made a few comments.
Could you also add some basic integration tests (in umap/test/integration) ? I'd say one basic case and one case where the data does not match the facet key. If you need help for this, please ask :)

Thanks again, that's a very nice contribution! :)

yohanboniface:
For the record, here is what the date form looks like currently in the facet form on my computer:

image

umap/static/umap/js/umap.js
```diff
@@ -2033,7 +2033,10 @@ L.U.Map.include({
   getFacetKeys: function () {
     return (this.options.facetKey || '').split(',').reduce((acc, curr) => {
       const els = curr.split('|')
-      acc[els[0]] = els[1] || els[0]
+      acc[els[0]] = {
+        "label": els[1] || els[0],
+        "type":  els[2] || "checkbox"
```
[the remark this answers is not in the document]
davidbgk: Default for an `<input>` element [looks to be](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types) `text`.

umap/static/umap/js/umap.features.js
```diff
@@ -483,12 +483,30 @@ L.U.FeatureMixin = {
...
+      value = parseFloat(value);
+      if (value < 10000000000) {
```
davidbgk, line 489: We need at least comments to explain these two values `10000000000` and `1000`.

flammermann:
Sorry, for the late response. Thanks for all the comments, will look into them now. ;)

flammermann:
> For the record, here is what the date form looks like currently in the facet form on my computer:
> 
> image

Yes, I found that too during my testing. It's not optimal but the user can use the date picker to choose the date and the keyboard (numbers or arrows) to enter/change the time. Also, I would expect/hope that Firefox eventually fixes their implementation to fully support datetime-local. Thus, I went with it despite the lack of support from Firefox.

umap/static/umap/js/umap.js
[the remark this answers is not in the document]
flammermann: I see your point and also have some ideas:

1. Stick to data format (date, time, datetime, number, string, enum, ...) and assume the required input field format (min/max fields for dates, times, numbers, text for fuzzy matching strings, checkboxes for enum).
2. Extend the functionality even further and have both data format and input field format. With reasonable defaults. This would allow even more scenarios (e.g. date with sliders instead of min/max fields, enum with radio instead of checkboxes).

For now, I went for option 1 because option 2 can be done as an extension of option 1 in another improvement PR.

I did not adjust the naming in `umap.forms.js` because there the word checkbox refers to the input field type, so it seems correct to me.

If your happy with that, please resolve this conversation.

umap/static/umap/js/umap.features.js
```diff
@@ -483,12 +483,30 @@ L.U.FeatureMixin = {
     return false
   },
 
+  parseDateField: function (value) {
```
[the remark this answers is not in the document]
flammermann: Regarding your inputs here:

1. Location of the code: Actually, I was not quite sure where to put such a function, so I just put it somewhere. Maybe you could provide a small definition for the different files and examples where to to put what for which reason in the developer documentation. And yes, nothing need to be instantiated, so I moved it as you suggested.
2. Support for ISO 8601: Actually this is given. I added some documentation on the function below. While writing the documentation, I figured that this code does exactly what I need it to do, but is generally not really good. Maybe it would be better for the user to explicitly state what kind of date is contained in the feature property. Can you give some feedback here?
3. Tests: I did not familiarize myself yet with that part of you code. Is there any explanation/documentation on writing such tests? If not, I will just work through it myself and figure something out. ;)

```
// the Date() constructor can handle various inputs to create a date
// - value: epoch (unix timestamp) in milliseconds
// - dateString: ISO 8601 formatted (YYYY-MM-DDTHH:mm:ss.sssZ)
// - dateObject: JS date object
// - multiple parameters for different date fields (year, month, ...)
//
// a mix of those options shall be supported without the user having
// to specify the exact format, since umap is based on json the type
// of the feature property value can only be
// - string, number or boolean
// - object or array
// - null
//
// therefore, the following inputs shall be covered
// - epoch (unix timestamp) in milliseconds as number
// - epoch (unix timestamp) in seconds as number
// - epoch (unix timestamp) in milliseconds as string
// - epoch (unix timestamp) in seconds as string
// - date in ISO 8601 format as string
//
// this function tries to guess the format of the feature property value
// and adjust it a little before passing it to Date() constructor
//
parseDateField: function (value) {
  if (value != null && parseFloat(value).toString() === value.toString()) {
    // if the string representation of the feature property value is
    // the same with and without being parsed as a float, the value is
    // a number (either of type number or string)
    //
    // numbers are assumed to be epochs (unix timestamps) but so far
    // it is unclear whether it is in seconds, milliseconds or nanoseconds
    //
    // without user input it can never be determined with certainty, but
    // by making some assumptions and sacrificing some small date ranges,
    // it is possible to work around that
    //
    value = parseFloat(value);
    if (Math.abs(value) < 10000000000) {
      // if the absolute value of that number is smaller than 10000000000,
      // it is assumed to be in seconds and must be multiplied by 1000
      //
      value = value * 1000;
    } else if (Math.abs(value) > 10000000000000) {
      // if the absolute value of that number is bigger than 10000000000000,
      // it is assumed to be in nanoseconds and must be divided by 1000
      //
      value = value / 1000;
    }
    // in all other cases the number is assumed to be in milliseconds
  }
  // in all other cases the value is passed to the Date() constructor
  // without modification and the constructor will try to make something out of it
  //
  // this can either result in something proper (e.g. string containing dateString),
  // something wrong (e.g. boolean) or an invalid date (e.g. object, array, null,
  // string not containing dateString)
  //
  return new Date(value);
}
```

umap/static/umap/js/umap.features.js
flammermann: See this comment: https://github.com/umap-project/umap/pull/1491#discussion_r1439922296

Not happy with this part of the implementation, happy to receive more feedback/suggestions from you.

umap/static/umap/js/umap.features.js
flammermann: You are absolutely right. Actually, I already saw that problem shortly after this pull request and resolved it on the spot. It's part of this commit: https://github.com/umap-project/umap/pull/1491/commits/96f808383291b6b880752e67f1e53bb349018999

umap/static/umap/js/umap.controls.js
flammermann: Yes, again I agree. We should distinguish. But maybe go even further letting the user define the data type as well as the input field type. Since we have another conversation going on about this, I will resolve this conversation.

https://github.com/umap-project/umap/pull/1491#discussion_r1439898242

umap/static/umap/js/umap.forms.js
flammermann: Start and End were supposed to be generic and also usable for numbers but since we have to have an own L.FormBuilder.Element deviate for those, we could also have nicer labels. I would go with From and Until for dates and times and Min and Max for numbers.

I also enhanced the implementation of L.FormBuilder.FacetSearchDate with a table. Looks nicer now.

image

umap/static/umap/js/umap.js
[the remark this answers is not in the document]
flammermann: I had some more minutes and actually implemented option 2 directly. This also gives us more flexibility for refactoring the parseDateField function (see https://github.com/umap-project/umap/pull/1491#discussion_r1439922296).

On the fly I also restricted the allowed values for dataType and inputType as well as allowed radio to be a inputType for enum dataType. Radio works quite ok, however right now there is no possibility to "deselect" the radio and show all features again. I will look into that tomorrow.

umap/static/umap/js/umap.features.js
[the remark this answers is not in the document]
yohanboniface: For the tests, just have a look a how it's done on this file, and add a new `describe` for your new util:

https://github.com/umap-project/umap/blob/master/umap/static/umap/test/Util.js

umap/static/umap/js/umap.core.js
```diff
@@ -296,6 +296,18 @@ L.Util.normalize = function (s) {
     .replace(/[\u0300-\u036f]/g, '')
 }
 
+L.Util.parseDateField = function (value) {
```
yohanboniface, line 299: `parseDate` ? Not sure what "field" is in this context :)

umap/static/umap/js/umap.js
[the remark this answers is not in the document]
yohanboniface: Thanks for exploring that right away! :)

I'm not convinced about adding two parameters, dataType and inputType. I think that in an ideal world, the data type comes from the data iself, or from the layer, but we don't need to set the data type at the moment of computing the facets.
I'll like to implement a sort of json-schema like in uMap at some point, but of course we are not there yet.
So in the meantime, I'd only keep one parameter in the facet definition, which is the input type, and I'd derive the data type from it until we get the type from elsewhere.
I'd suggest for a first step to support those input types:
- checkbox (default) => data is enum, so we only check for equality
- radio (nice suggestion) => same data check
- minmax => two input min and max, and we check for the value to be inside (including boundaries)
- date => two inputs start and end, and we have a date picker
- datetime => same, but the date picker includes the time (if the browser supports it…)

I'm also concerned about keeping this nice enhancement as simple as possible, so we can let users use it and we can better priorise what to do next from their feedback. Also, once we have implement some syntax and users start using it, it means we need to support it for long.

What do you think ? :)

yohanboniface (commented):
:+1:

umap/static/umap/js/umap.js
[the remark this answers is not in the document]
flammermann: I like the idea of allowing JSON input to define facets. The | splitting is becoming rather cumbersome and also brings the disadvantage of having to define all previous parameters, even if the user does not care about it (e.g. have to define label, otherwise I cannot define type). The only reason, I did not do this straight away were concerns about backwards compatibility.

I also like the idea of getting the data type from somewhere else. Brainstorming here a little: In think the easiest approach would be for the user to be able to define one data type per feature property name across all layers. We cannot have different data types per layer because facet search acts across all layers. Maybe we could still implement the data type definition per feature property name per layer and handle deviating data types during the facet search (e.g. excluding incompatible data types, defining groups of data types that work together, etc.)

As mentioned in one of my other comments, I am also rather unhappy with the complexity of the parseDateField function. Let's reduce the complexity of this change even further by not trying to be clever when the data is parsed but rely on existing JS functionality. However, I would change the input type naming a little to make things more transparent.

- checkbox (default) => one checkbox input per unique value, check for equality, data is passed through toString() method, should work for everything
- radio => same as checkbox but with radio input
- number_minmax => two number inputs for min and max, check for the value to be inside (including boundaries), data is passed through parseFloat function, NaNs are considered a match (otherwise features would disappear if property is empty)
- date_minmax => same as number_minmax but with date inputs and data is passed through Date() constructor
- datetime_minmax => same as date_minmax but with datetime-local inputs

Sounds good?

umap/static/umap/js/umap.features.js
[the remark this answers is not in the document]
flammermann: Perfect, thanks a lot. Seems pretty straight forward. I will reduce complexity of the function and then add some tests. :)

umap/static/umap/js/umap.core.js
flammermann: There is no particular reason for the Field, it just sounded right. But I really don't have any strong feeling on the naming here. ;)

umap/static/umap/js/umap.js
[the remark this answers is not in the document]
flammermann: Yes, keeping the `|` splitting for this PR.

umap/static/umap/js/umap.features.js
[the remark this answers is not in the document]
flammermann: Function became obsolete, decision made that for now only those data types that are supported by the Date() constructor are used.

umap/static/umap/js/umap.core.js
flammermann: Function became obsolete. Decision was made to only support data types that are covered by the Date() constructor itself.

flammermann:
I refactored the code according to your suggestions. Support for number properties is now also given. I think the PR is in pretty good shape now. Thanks for the help and review! :)

yohanboniface:
> Could you also add some basic integration tests (in umap/test/integration) ? I'd say one basic case and one case where the data does not match the facet key. 

I've [pushed](https://github.com/umap-project/umap/commit/ecd4c73a7ef93b7a40a6290c1e8636e68c4589b0) as basic facet search test, so you can base new ones on this :)

umap/static/umap/js/umap.controls.js
```diff
@@ -741,21 +741,57 @@ L.U.Map.include({
...
+            if (["number"].includes(inputType)) value = parseFloat(value);
+            if (!isNaN(value) && (isNaN(facetCriteria[key]["min"]) || facetCriteria[key]["min"] > value)) {
+              facetCriteria[key]["min"] = value
+            }
+            if (!isNaN(value) && (isNaN(facetCriteria[key]["max"]) || facetCriteria[key]["max"] < value)) {
+              facetCriteria[key]["max"] = value
+            }
+          } else {
+            value = String(value)
+            value = (value.length ? value : "empty string")
```
yohanboniface, line 791: I'd say we want to translate this string:
```suggestion
            value = (value.length ? value : L._("empty string"))
```

umap/static/umap/js/umap.controls.js
```diff
@@ -771,13 +807,12 @@ L.U.Map.include({
...
-
-    const fields = keys.map((current) => [
-      `facets.${current}`,
+    const fields = keys.map((key) => [
+      `facets.${key}`,
       {
-        handler: 'FacetSearch',
-        choices: knownValues[current],
-        label: this.getFacetKeys()[current],
+        handler: ["date", "datetime-local", "number"].includes(facetCriteria[key]["inputType"]) ? 'FacetSearchMinMax' : 'FacetSearchChoices',
```
yohanboniface, line 813: I'd keep the handler choice out of the object for readability.

umap/static/umap/js/umap.features.js
```diff
@@ -485,10 +485,30 @@ L.U.FeatureMixin = {
...
+          min = parseFloat(min)
+          max = parseFloat(max)
+          value = parseFloat(value)
+        }
+        if (!isNaN(min) && !isNaN(value) && min > value) return false
+        if (!isNaN(max) && !isNaN(value) && max < value) return false
+      } else {
+        const choices = criteria["choices"]
+        value = String(value)
+        value = (value.length ? value : "empty string")
```
yohanboniface, line 510:
```suggestion
        value = (value.length ? value : L._("empty string"))
```

umap/static/umap/js/umap.forms.js
```diff
@@ -802,20 +804,116 @@ L.FormBuilder.FacetSearch = L.FormBuilder.Element.extend({
...
+    this.maxInput.type = this.inputType
+    this.maxInput.id = `${this.inputType}_${this.name}_max`
+    this.maxInput.step = '1'
+    if (max != null) {
+      this.maxInput.valueAsNumber = max.valueOf()
+      this.maxInput.dataset.value = max
+    }
+
+    if (["date", "datetime-local"].includes(this.inputType)) {
+      this.minLabel.innerHTML = 'From'
```
yohanboniface, line 871:
```suggestion
      this.minLabel.innerHTML = L._('From')
```

umap/static/umap/js/umap.forms.js
```diff
@@ -802,20 +804,116 @@ L.FormBuilder.FacetSearch = L.FormBuilder.Element.extend({
...
+      this.maxLabel.innerHTML = 'Until'
```
yohanboniface, line 872:
```suggestion
      this.maxLabel.innerHTML = L._('Until')
```

umap/static/umap/js/umap.forms.js
```diff
@@ -802,20 +804,116 @@ L.FormBuilder.FacetSearch = L.FormBuilder.Element.extend({
...
+    }
+
+    if (["datetime-local"].includes(this.inputType)) {
+      this.minInput.step = '0.001'
+      this.maxInput.step = '0.001'
+    }
+
+    if (["number"].includes(this.inputType)) {
+      if (min != null && max != null) {
+        // calculate step from significant digits of min and max values
```
yohanboniface, line 888: This should be extracted in a small util (in umap.core.js) and tested :)

umap/static/umap/js/umap.forms.js
```diff
@@ -802,20 +804,116 @@ L.FormBuilder.FacetSearch = L.FormBuilder.Element.extend({
...
+    if (min != null) {
+      this.minInput.valueAsNumber = min.valueOf()
+      this.minInput.dataset.value = min
+    }
+
+    this.maxTr = L.DomUtil.create('tr', '', this.table)
+
+    this.maxTdLabel = L.DomUtil.create('td', '', this.maxTr)
+    this.maxLabel = L.DomUtil.create('label', '', this.maxTdLabel)
+    this.maxLabel.innerHTML = 'Max'
```
yohanboniface, line 857:
```suggestion
    this.maxLabel.innerHTML = L._('Max')
```

umap/static/umap/js/umap.forms.js
```diff
@@ -802,20 +804,116 @@ L.FormBuilder.FacetSearch = L.FormBuilder.Element.extend({
...
+    this.inputType = this.options.criteria["inputType"]
+
+    const min = this.options.criteria['min']
+    const max = this.options.criteria['max']
+
+    this.minTr = L.DomUtil.create('tr', '', this.table)
+
+    this.minTdLabel = L.DomUtil.create('td', '', this.minTr)
+    this.minLabel = L.DomUtil.create('label', '', this.minTdLabel)
+    this.minLabel.innerHTML = 'Min'
```
yohanboniface, line 840:
```suggestion
    this.minLabel.innerHTML = L._('Min')
```

yohanboniface (commented):
We are on good track, thanks! :)

yohanboniface:
hey @flammermann any time/interest in finishing this PR ? :)

flammermann:
> hey @flammermann any time/interest in finishing this PR ? :)

Hi @yohanboniface, yes, I am still interested in finishing it. Unfortunately, my server crashed hard (harddrive/raid failure) shortly after my last comment/commit, so I could not continue. :/ But you got perfect timing, I finished recovery and rebuilding everything last Saturday and can now finally pick up this PR again. :)

Btw. I got lucky because I could export a lot of work on one of my maps that was not yet backed up properly from the still open browser tab. ;)

umap/static/umap/js/umap.forms.js
flammermann: Can you recheck this one, please? I moved the functionality to umap.core.js as you suggested.

umap/static/umap/js/umap.controls.js
flammermann: I moved the code out of the object. There are now three variables in the mapping function, which are used in the object definition. Is that what you meant?

yohanboniface:
Work continues in #1763

PR #1492, opened 2023-12-28
Fix preview of TMS TileLayer preview

yohanboniface:
This is not the ideal fix, instead we'd prefer to use `TileLayer.getTileUrl`, but this is not possible yet.

cf https://github.com/Leaflet/Leaflet/pull/9201

cf https://lists.openstreetmap.org/pipermail/umap/2023-December/000557.html

### DIFF
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -29,7 +29,12 @@ L.Map.mergeOptions({
   name: '',
   description: '',
   displayPopupFooter: false,
-  demoTileInfos: { s: 'a', z: 9, x: 265, y: 181, r: '' },
+  // When a TileLayer is in TMS mode, it needs -y instead of y.
+  // This is usually handled by the TileLayer instance itself, but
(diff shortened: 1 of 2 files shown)


umap/static/umap/js/umap.js
```diff
@@ -29,7 +29,7 @@ L.Map.mergeOptions({
   name: '',
   description: '',
   displayPopupFooter: false,
-  demoTileInfos: { s: 'a', z: 9, x: 265, y: 181, r: '' },
+  demoTileInfos: { s: 'a', z: 9, x: 265, y: 181, '-y': 181, r: '' },
```
davidbgk, line 32: I don't get that part 🤔

yohanboniface: Just pushed a comment. Does it make clearer the change ? :)

davidbgk: Yup 👍

davidbgk (approved):

PR #1493, opened 2023-12-28
Add a small box-shadow to tilelayer preview

### LINKED ISSUE #1485: [maps] Make it more visible; Add Transport map
Hello,

It's halfway between a bug and a suggestion.

The Maps section of uMap changed recently.

1. It's a bit hard to see. I suggest making it more visible (add border?)
3. The Transport map is missing from the list. I'm probably not the only one to find that map very useful.

Cheers,

image

---
Edit: To choose a map not shown in View mode, switch to Edit mode and select it through the "Change tilelayers" button. If not already available, I suggest letting the user choose what maps fit in the ten shown in View mode, ideally through a simple, in place solution — ie. withougt having to dive into the Settings section after switching to Edit mode.

yohanboniface:
In some situation, the preview is very similar to the background, so it's not visible enough.

fix #1485

image

image

### DIFF
--- a/umap/static/umap/map.css
+++ b/umap/static/umap/map.css
@@ -164,7 +164,7 @@
 .leaflet-iconLayers-layer {
     width: 38px;
     height: 38px;
-    box-shadow: none;
+    box-shadow: 0px 0px 2px #444;
     border: 1px solid #bbb;
     border-radius: 4px;
 }


davidbgk (approved):
Looks better, I was thinking maybe to add a background/shadow too when all icon layers are opened to improve the visibility?

yohanboniface:
The same box-shadow applies :)

image

davidbgk:
Alright, LGTM!

PR #1496, opened 2023-12-29
Only load Leaflet once, as module

yohanboniface:
cc @vinyll @almet 

(I cannot run the frontend tests, because it seems the fix to run them from a running browser is not on this branch. @almet if you have the opportunity to rebase the source branch, I'll rebase here after :) )


yohanboniface:
Done directly in branch #1463

PR #1497, opened 2023-12-30
fix dirty flags when re-ordering layers

### LINKED ISSUE #375: Layer order broke on update
Hi, 

Following the 23/9/16 update, layer ordering broke in "About" & "Data browser"

Layers display in random order. Even on cloned maps, layer ordering is now different.

Reordering layers with the new graphical _doesn't_ fix the problem.

Plus, reordering this way doesn't produce coherent results.

I've ordered layers from 01 to 10 on two similar maps (one original, one cloned). On the original, layers are now ordered in this way :
- 4 A
- 2 B
- 3 C
- 1 D 

Where alphabetic order overcomes numeric order. On the clone, no order appears, neither alphabetic nor numeric.

Newly created maps don't seem affected ([example](https://umap.openstreetmap.fr/en/map/ordering-bug_103889)). But most of my already existing maps are. Noticeably, my private maps aren't affected.

jschleic:
The index of the top layer in the view starts with zero, while the rank of the layers count backwards. Thus moving the second-last to the last position should set the dirty flag of rank 0 and 1. Instead the former implementation set the dirty flag for layers >= 5, when moving "Red" to the bottom in [the test map](https://umap.openstreetmap.fr/fr/map/test-ordering_994821) - resulting in the wrong layers saved.

Fixes #375

### DIFF
--- a/umap/static/umap/js/umap.controls.js
+++ b/umap/static/umap/js/umap.controls.js
@@ -632,12 +632,14 @@ L.U.DataLayersControl = L.Control.extend({
       function (e) {
         const layer = this.map.datalayers[e.src.dataset.id],
           other = this.map.datalayers[e.dst.dataset.id],
-          minIndex = Math.min(e.initialIndex, e.finalIndex)
+          minIndex = Math.min(layer.getRank(), other.getRank()),
+          maxIndex = Math.max(layer.getRank(), other.getRank())
         if (e.finalIndex === 0) layer.bringToTop()
         else if (e.finalIndex > e.initialIndex) layer.insertBefore(other)
         else layer.insertAfter(other)
         this.map.eachDataLayerReverse((datalayer) => {
-          if (datalayer.getRank() >= minIndex) datalayer.isDirty = true
+          if (datalayer.getRank() >= minIndex && datalayer.getRank() <= maxIndex)
+            datalayer.isDirty = true
         })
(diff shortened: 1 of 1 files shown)


yohanboniface:
Sounds good to me. Thanks a lot for spotting this and providing the fix :)

umap/static/umap/js/umap.controls.js
```diff
@@ -632,12 +632,14 @@ L.U.DataLayersControl = L.Control.extend({
...
-          minIndex = Math.min(e.initialIndex, e.finalIndex)
+          minIndex = Math.min(layer.getRank(), other.getRank()),
+          maxIndex = Math.max(layer.getRank(), other.getRank())
         if (e.finalIndex === 0) layer.bringToTop()
         else if (e.finalIndex > e.initialIndex) layer.insertBefore(other)
         else layer.insertAfter(other)
         this.map.eachDataLayerReverse((datalayer) => {
-          if (datalayer.getRank() >= minIndex) datalayer.isDirty = true
+          if (datalayer.getRank() >= minIndex && datalayer.getRank() <= maxIndex)
+            datalayer.isDirty = true
```
davidbgk, line 642: Why not `datalayer.isDirty = datalayer.getRank() >= minIndex && datalayer.getRank() <= maxIndex`?

PR #1498, opened 2024-01-02
Align copy button to input in share box

yohanboniface:
cf #1454

image

(Let's make iframe textarea copiable in a later step.)

cc @jschleic :)


jschleic (approved):
Looks great! Thanks for continuing :-)

PR #1516, opened 2024-01-09
Set a umap-fragment web component for lists

davidbgk:


umap/static/umap/js/umap.components.js
```diff
@@ -0,0 +1,13 @@
+class UmapFragment extends HTMLElement {
```
yohanboniface, line 1: Maybe move in a `components` subfolder ? `components/fragment.js` ?

yohanboniface (commented):
Sounds clean!

It's not exactly clear to me what's the support of this. Is that this feature ? https://caniuse.com/custom-elementsv1

umap/static/umap/js/umap.components.js
```diff
@@ -0,0 +1,13 @@
...
+  connectedCallback() {
+    new L.U.Map(this.firstElementChild.id, JSON.parse(this.dataset.settings))
```
almet, line 3: I suppose the reason the id is in the inner `div` (rather than directly in the `umap-fragment`) is that umap currently expect a `div`element? 🤔

umap/templates/umap/content.html
```diff
@@ -68,12 +68,6 @@
           callback: function (data) {
             const container = this.parentNode
             container.innerHTML = data
```
almet, line 70: At the moment, this is including all the retrieved page inside the current one(I guess it's why it's still a draft :-))

umap/templates/umap/js.html
```diff
@@ -58,4 +58,5 @@
         <script src="{{ STATIC_URL }}umap/js/umap.share.js" defer></script>
         <script src="{{ STATIC_URL }}umap/js/umap.js" defer></script>
         <script src="{{ STATIC_URL }}umap/js/umap.ui.js" defer></script>
+        <script src="{{ STATIC_URL }}umap/js/umap.components.js" defer></script>
```
almet, line 61: I'm trying to understand how do we guarantee that map are loaded when all the context is ready. Do I understand correctly that this `umap.component` is expected to be included after all the other ones, so the dom gets parsed then only?

umap/templates/umap/content.html
davidbgk: All the retrieved page is actually an HTML fragment mainly with the list of web components with correct ids/settings.

almet: ah okay, thanks for the clarification :-)

davidbgk:
> It's not exactly clear to me what's the support of this. Is that this feature ? https://caniuse.com/custom-elementsv1

This is it, and we can add an optional polyfill if don't feel confident enough of the current support: https://github.com/ungap/custom-elements

umap/templates/umap/js.html
davidbgk: If I understand correctly [this article](https://lamplightdev.com/blog/2021/03/15/when-are-the-constructor-and-connectedcallback-methods-called-when-creating-a-custom-element/), the `connectedCallback()` from the `umap-fragment` custom element will be called only when the DOM is evaluated within the page.

Even if I delay by 5sec the registration of the custom element:

`setTimeout(() => register(UmapFragment, 'umap-fragment'), 5000)`

Once loaded, the maps will eventually appear. I can even put the whole `UmapFragment` class definition within the timeout and it will still work (which is a bit magical!).

umap/static/umap/js/umap.components.js
davidbgk: Yes, `Map()` is expecting an `id` for the container and I reused the already defined `class` to avoid a bigger change. Somehow we have a cleaner separation between data (settings) related to the component and id/class for the rendering of the map.

One option could have been to generate the `div` from `umap-fragment` parameters but this way we have the HTML part in an HTML file :)

umap/templates/umap/js.html
davidbgk: The [Custom element lifecycle callbacks](https://developer.mozilla.org/en-US/docs/Web/API/Web_Components/Using_custom_elements#custom_element_lifecycle_callbacks)

davidbgk: > So a `<script defer src="yourelement.js">` does the job; **but** will run **after** all DOM is created,
your components are now created (very) late. So you now have to deal with (more) FOUCs. — https://stackoverflow.com/a/70952159

Maybe something to consider, even for #1463

almet (approved):
yay

PR #1518, opened 2024-01-10
Add IDs on features

almet:
Just to have it somewhere in case

### DIFF
--- a/umap/static/umap/js/modules/utils.js
+++ b/umap/static/umap/js/modules/utils.js
@@ -1,0 +1,8 @@
+export function generateId() {
+  // A real implementation would benefit from another id type.
+  // Using uppercase + lowercase + digits, here's the collision risk
+  // For 6 chars, 1 in 100 000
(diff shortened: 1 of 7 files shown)


umap/static/umap/js/modules/utils.js
```diff
@@ -0,0 +1,8 @@
+export function generateId() {
+  // A real implementation would benefit from another id type.
+  // Using uppercase + lowercase + digits, here's the collision risk
+  // For 6 chars, 1 in 100 000
+  // For 5 chars, 5 in 100 000
+  // for 4 chars, 500 in 100 000
+  return (((1 + Math.random()) * 0x10000) | 0).toString(16).substring(1)
```
davidbgk, line 7: Does it start with a letter (vs. number) to make it html-id/class compatible?

almet: Good idea :-)

PR #1529, opened 2024-01-15
docs: document how dependencies are handled

almet:


almet:
Subsequent to https://github.com/umap-project/umap/pull/1514

docs/dev/dependencies.md
```diff
@@ -0,0 +1,22 @@
...
+
+Dependencies are **not** pinned, but a `^` character is used instead ([defined
+by node semver](https://github.com/npm/node-semver#caret-ranges-123-025-004)),
+meaning the next minor or patch versions will be installed (but not the next major)
+
+The installed libs needs to be:
+
+- Added to the `package.json`
+- Added in the `scripts/vendorsjs.sh` script
+- Loaded in the HTML templates.
```
davidbgk, line 21: + Added in `umap/test/index.html`
+ `make vendors`

?

almet: > Added in umap/test/index.html

With the modules it's not a requirement anymore, so it really depends what you want to do.

`make vendors` indeed!

almet: (updated on master :see_no_evil:)

almet: > make installjs vendors

Is there an use case for running `make vendors` without running `installjs` first? (I'm wondering if we can make one a dependency of the other)

yohanboniface: I don't see any use case to run only one but no the other, let's do that :)

PR #1531, opened 2024-01-15
Introduce UMAP_HOME_FEED to control which maps are shown on the home page

yohanboniface:
For now we have only three modes:
- latest, which is the default and shows the last updated maps
- highlighted, which shows only the map that have been starred by a least one staff member
- None, which does not show any map

### DIFF
--- a/docs/config/settings.md
+++ b/docs/config/settings.md
@@ -218,6 +218,13 @@ ready for production use (no backup, etc.)
 
 Link to show on the header under the "Feedback and help" label.
 
+#### UMAP_HOME_FEED
+
+Which feed to display on the home page. Three valid values:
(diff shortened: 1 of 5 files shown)


docs/config/settings.md
```diff
@@ -218,6 +218,13 @@ ready for production use (no backup, etc.)
 
 Link to show on the header under the "Feedback and help" label.
 
+#### UMAP_HOME_FEED
+
+Which feed to display on the home page. Two valid values:
```
davidbgk, line 223: I count three :p

davidbgk (approved):

docs/config/settings.md
yohanboniface: Now me too! :p

PR #1539, opened 2024-01-16
Use custom storage instead of django-compressor

davidbgk:
This is a really early stage.

* [x] Deal with the commented `favicon.ico` which tries to reach `staticfiles_storage.url("umap/favicons/favicon.ico")`
* [x] Check that all paths are well rewritten
* [x] Handle [dynamic {locale}.js path](https://github.com/umap-project/umap/pull/1539/files#diff-b642b030e74563d2d5e096cdb16de5c85eb02fb18191734f8995a4fa51246212R30), not sure how it can work, maybe with intermediary variable?
* [x] Fix tests :)

The `support_js_module_import_aggregation` toggle + wrong regexps for our prettier semicolons-less config were hard to find!

### DIFF
--- a/umap/utils.py
+++ b/umap/utils.py
@@ -2,6 +2,7 @@ import gzip
 import os
 
 from django.conf import settings
+from django.contrib.staticfiles.storage import ManifestStaticFilesStorage
 from django.urls import URLPattern, URLResolver, get_resolver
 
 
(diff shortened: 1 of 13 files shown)


umap/utils.py
```diff
@@ -162,3 +163,53 @@ def merge_features(reference: list, latest: list, incoming: list):
         merged.append(item)
 
     return merged
+
+
+class UmapManifestStaticFilesStorage(ManifestStaticFilesStorage):
```
yohanboniface, line 168: Is this supposed to be registered in some setting ? I've tried the branch, but the statics have still their plain name.

yohanboniface:
>  Deal with the commented favicon.ico which tries to reach staticfiles_storage.url("umap/favicons/favicon.ico")

I'm missing the issue here: it seems to work well locally, either with DEBUG=True or False and using an nginx proxy, the path to favicon.ico seems to be updated in the served HTML, I see that even favicon.ico exists in the static dir, so even if some browser wants that path explicitly it will find it. So I guess I'm missing the point ;)

davidbgk:
> So I guess I'm missing the point ;)

Well, in a previous attempt it was an issue during the `collectstatic` command, maybe another setting was wrong because now it works for me too!

davidbgk:
> now it works for me too!

Not so fast, tests are failing because of that.

yohanboniface (approved):

PR #1549, opened 2024-01-21
Use minified versions + sourcemaps of vendor statics

davidbgk:
We need to find a way to keep in sync `css.html` + `js.html` with `test/index.html`. Maybe generating the page with Django too?


yohanboniface:
> We need to find a way to keep in sync css.html + js.html with test/index.html. Maybe generating the page with Django too?

Right! Now that we need the `index.html` to be served, why not serving it by Django ?

umap/storage.py
```diff
@@ -38,34 +38,19 @@ class UmapManifestStaticFilesStorage(ManifestStaticFilesStorage):
         ),
     )
 
-    # https://github.com/django/django/blob/0fcee1676c7f14bb08e2cc662898dee56d9cf207↩
```
yohanboniface, line 41: This change is unclear to me.

davidbgk: Previously, we were skipping `.map` files from the default regexp but now that these files are copied it's OK to get back to default and hash these filenames too.

yohanboniface (approved):

PR #1551, opened 2024-01-22
fix: build browser once features are on the map, not before

### LINKED ISSUE #1519: "Browse data" button becomes unfunctional when filter has no result
**Describe the bug**
in title

**To Reproduce**
1. Go to 'https://umap.openstreetmap.fr/en/map/map_943303' 
2. Click on 'Browse data'
3. Type "x". nothing is found.
4. Click on 'About'
5. Click on 'Browse data' button.
6. Button doesn't work. 

 - OS: [Windows]
 - Browser [Firefox]
 - Version [121.0]

**Additional context**
On step 3 data disappeared from the map, and after clearing filter box map remains empty.

yohanboniface:
Features title may contain variables, which include geographical ones (center, lat, lon…), and in this case the feature must be on the map to be able to compute them (eg. the polygon center).

fix #1519

### DIFF
--- a/umap/static/umap/js/umap.browser.js
+++ b/umap/static/umap/js/umap.browser.js
@@ -110,13 +110,13 @@ L.U.Browser = L.Class.extend({
     const formContainer = L.DomUtil.create('div', '', container)
     const dataContainer = L.DomUtil.create('div', 'umap-browse-features', container)
 
-    const appendAll = () => {
+    const rebuildHTML = () => {
       dataContainer.innerHTML = ''
       this.map.eachBrowsableDataLayer((datalayer) => {
         this.addDatalayer(datalayer, dataContainer)
       })
     }
-    const resetLayers = () => {
+    const redrawDataLayers = () => {
       this.map.eachBrowsableDataLayer((datalayer) => {
         datalayer.resetLayer(true)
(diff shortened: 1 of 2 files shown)


umap/static/umap/js/umap.browser.js
```diff
@@ -129,16 +129,16 @@ L.U.Browser = L.Class.extend({
...
       callback: (e) => {
         if (e.helper.field === 'options.inBbox') {
-          if (this.options.inBbox) this.map.on('moveend', appendAll)
-          else this.map.off('moveend', appendAll)
+          if (this.options.inBbox) this.map.on('moveend', rebuildHTML)
+          else this.map.off('moveend', rebuildHTML)
         }
-        appendAll()
-        resetLayers()
+        redrawDataLayers()
```
davidbgk, line 135: Here are the two switched lines :)

davidbgk (approved):

umap/static/umap/js/umap.browser.js
yohanboniface: Yes, with the renaming, that was tricky to catch!

I thought about making two commits for that, but… I know you'd catch it anyway :p

PR #1555, opened 2024-01-23
chore: move xhr management to a module and refactor

yohanboniface:
It sort of work for common cases, but edge cases needs work, specifically the login flow.

- [x] better error handling
- [x] 412
- [ ] dependency on UI
- [ ] login required when saving the map/datalayers


umap/static/umap/js/modules/request.js
```diff
@@ -0,0 +1,157 @@
+// Uses `L._`` from Leaflet.i18n which we cannot import as a module yet
+import { Evented, DomUtil } from '../../vendors/leaflet/leaflet-src.esm.js'
+
+export class HTTPError extends Error {
```
almet, line 4: Should all the exception inherit from the same base class, so we can intercept all the ones from this module alltogether?

umap/static/umap/js/modules/request.js
```diff
@@ -0,0 +1,157 @@
...
+  constructor(message) {
+    super(message)
+    this.name = this.constructor.name
+  }
+}
+
+export class NokError extends Error {
```
almet, line 11: nit: should this be in caps (`NOKError`) ?

almet: Documenting what's the difference between `HTTPError` and `NOKError` could be useful :-)

umap/static/umap/js/modules/request.js
```diff
@@ -0,0 +1,160 @@
...
+    } catch (error) {
+      console.error(error)
+      this.fire('dataload', { id: id })
+      throw new HTTPError(error.message)
+    }
+    if (!response.ok) {
+      this.fire('dataload', { id: id })
+      throw new NOKError(response)
+    }
+    // TODO
```
davidbgk, line 45: I guess it's done?

umap/static/umap/js/modules/request.js
```diff
@@ -0,0 +1,160 @@
...
+export class NOKError extends RequestError {
+  constructor(response) {
+    super(response.status)
+    this.response = response
+    this.status = response.status
+    this.name = this.constructor.name
+  }
+}
+
+const BaseRequest = Evented.extend({
```
davidbgk, line 23: I'm still wondering if it's a good idea to extend the Leaflet logic here, let's talk one last time about that :)

umap/static/umap/js/umap.js
```diff
@@ -1093,64 +1061,62 @@ L.U.Map.include({
...
-                callback: () => {
-                  L.Util.copyToClipboard(data.permissions.anonymous_edit_url)
-                  this.ui.alert({
-                    content: L._('Secret edit link copied to clipboard!'),
-                    level: 'info',
-                  })
-                },
-                callbackContext: this,
+    const uri = this.urls.get('map_save', { map_id: this.options.umap_id })
+    const [data, response, error] = await this.server.post(uri, {}, formData)
```
almet, line 1065: Could it be possible to do the `this.fire('dataloading', e))` here? (I'm trying to see if we could get rid of events at the request level)

umap/static/umap/js/umap.js
```diff
@@ -1093,64 +1061,62 @@ L.U.Map.include({
...
+    if (!error) {
+      let duration = 3000,
```
almet, line 1067: (and here for the  `dataload` maybe?)

davidbgk (approved):

PR #1564, opened 2024-01-29
tests: ask playwright to way for ajax before checking the page state

yohanboniface:
Should fix the instable test.


yohanboniface:
Does not work as expected: test is still failing intermittently :(

PR #1565, opened 2024-01-29
tests: fix test_collaborative_editing_draw_markers

yohanboniface:
We needed to:
- make playwright to wait for ajax requests
- add a sleep to prevent saving two datalayer's version in the same second


davidbgk (approved):
😴

PR #1568, opened 2024-01-29
lint: use eslint with a browserslist configuration

davidbgk:


package.json
```diff
@@ -60,5 +62,8 @@
     "simple-statistics": "^7.8.3",
     "togpx": "^0.5.4",
     "tokml": "0.4.0"
-  }
+  },
+  "browserslist": [
+    "> 0.5%, last 2 versions, Firefox ESR, not dead, not op_mini all"
```
davidbgk, line 67: Excluding Opera mini might be not the best strategy, Opera is still [used in Africa](https://gs.statcounter.com/browser-market-share/all/africa) (latest release March, 2015 though).

If we want to support it, we have to implement our own URL parsing: https://caniuse.com/url

yohanboniface: Are we supporting it with our latest stable release ?

PR #1569, opened 2024-01-29
feat: Adding Sentry errors reporter support

davidbgk:
We might want to configure it for the JS side too: https://docs.sentry.io/platforms/javascript/

### DIFF
--- a/docs/config/customize.md
+++ b/docs/config/customize.md
@@ -100,3 +100,15 @@ There are three settings you can play with to control that:
     # primary key and not expose the username (which may be private or may change too
     # often for URL persistance)
     USER_URL_FIELD = "username"
+
+
+## Custom header and/or footer scripts
+
+You can populate the content of you own `umap/header.html` and `umap/footer.html`
+templates with `<script>` elements, it might be useful for instance 
+for analytics (footer) or bug reporting like Sentry (header).
+
+By default, these files are empty.
+You must define you own paths for extra templates following
+[Django’s documentation for this particular setting](https://docs.djangoproject.com/en/4.2/ref/settings/#templates).
+
--- a/umap/templates/registration/login.html
+++ b/umap/templates/registration/login.html
@@ -2,6 +2,7 @@
 {% load umap_tags i18n %}
 {% block extra_head %}
   {% umap_css %}
+  {{ block.super }}
 {% endblock extra_head %}
 {% block body_class %}
   login
--- a/umap/templates/base.html
+++ b/umap/templates/base.html
@@ -12,6 +12,7 @@
           content="{% trans "uMap lets you create maps with OpenStreetMap layers in a minute and embed them in your site." %}">
     <meta name="keywords" content="map, umap">
     {% block extra_head %}
+      {% include "umap/header.html" %}
     {% endblock extra_head %}
     <meta name="viewport"
           content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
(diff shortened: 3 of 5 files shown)


pyproject.toml
```diff
@@ -45,14 +45,15 @@ dependencies = [
...
-  "mkdocs==1.5.3",
+  "hatch==1.9.1",
+  "isort==5.12",
   "mkdocs-material==9.4.14",
-  "vermin==1.6.0",
+  "mkdocs==1.5.3",
   "pymdown-extensions==10.4",
-  "isort==5.12",
+  "ruff==0.1.6",
+  "sentry-sdk==1.39.2",
```
davidbgk, line 55: I put it as a `dev` dependency to not force it but we might want to create a dedicated `optional-dependencies` section.

almet: having this as a `dev` dependency seems off to me (we shouldn't need dev tools on the server we deploy to)

davidbgk: Alright, I moved it to a dedicated section 👍 

🔖  update the deployment script!

yohanboniface:
Does this means that the Sentry access is valid for any uMap instance ?

davidbgk:
> Does this means that the Sentry access is valid for any uMap instance ?

I applied for Sentry's open-source program specifically mentioning the OSM-FR instance. I guess it would be logical to submit a request per instance if necessary, and some instances should consider a paying option 🤷

yohanboniface:
In that case maybe the sentry python package should be installed by our deploy script ?

davidbgk:
> In that case maybe the sentry python package should be installed by our deploy script ?

Yes! https://github.com/umap-project/umap/pull/1569#discussion_r1470347736

davidbgk:
> * (basejs.html which is inherited from js.html with a super block)

It's hardly doable given that `js_html` is an inclusion tag calling the `js.html` template, I'll get back to the original idea of providing an extra template atop other imports.

docs/config/customize.md
```diff
@@ -99,4 +99,16 @@ There are three settings you can play with to control that:
...
     # primary key and not expose the username (which may be private or may change too
     # often for URL persistance)
-    USER_URL_FIELD = "username"
\ No newline at end of file
+    USER_URL_FIELD = "username"
+
+
+## Custom header and/or footer scripts
+
+You can populate the content of you own `umap/header.html` and `umap/footer.html`
```
yohanboniface, line 107: I'm not sure about the naming: `header.html` seems to me to be content oriented, but given it will be included in the `<head>` section, I'm tempted to call it `head.html`, `extra_head.html`…

yohanboniface (approved):

docs/config/customize.md
davidbgk: I choose the `header` name to make it the explicit counterpart of the existing `footer`, I don't know if editors/installers will have the knowledge to make the distinction between header/head/extra_head. I'm willing to change it if you think so.

umap/templates/registration/login.html
```diff
@@ -2,6 +2,7 @@
 {% load umap_tags i18n %}
 {% block extra_head %}
   {% umap_css %}
+  {{ block.super }}
```
almet, line 5: Cool, I didn't know about `block.super`.

docs/config/customize.md
yohanboniface: OK, let's go like that and see how people use it, if they use it ;)

PR #1587, opened 2024-02-08
chore: do not introduce a new global, use L and L.U already there

yohanboniface:


almet:
The rationale for having them separated in the first place:

- `L.U` for leaflet-related specific things; whereas
- `umap` is for modules exposed to the global namespace (so the non-module code out there can use it).

yohanboniface:
We can take ii the other way:
- `L` is for Leaflet related things
- `L.U` is for uMap related things

And both should disappear at some point inch Allah, which is the main reason why I'd prefer not to introduce a new global…

almet:
> L.U is for uMap related things

Maybe we can drop the `L` prefix then ? It seems off to have leaflet-unrelated-things inside the `L` namespace.

davidbgk (approved):

almet:
I was thinking dropping it for the non-leaflet stuff, but as we have some stuff related to leaflet, it could make sense to keep the two namespaces in my opinion.

PR #1593, opened 2024-02-12
feat: change breakpoint to switch black SVG img to white

yohanboniface:
I've been a report about changed pictograms, which pointed to this map:

image

I found that with those grey, the picto should have been white, so I had a look and increased a bit the threshold.

Before:
Screenshot from 2024-02-12 12-42-02

After:
Screenshot from 2024-02-12 12-41-49

### DIFF
--- a/umap/static/umap/js/umap.core.js
+++ b/umap/static/umap/js/umap.core.js
@@ -410,7 +410,8 @@ L.DomUtil.contrastedColor = (el, bgcolor) => {
   rgb = L.DomUtil.RGBRegex.exec(rgb)
   if (!rgb || rgb.length !== 4) return out
   rgb = parseInt(rgb[1], 10) + parseInt(rgb[2], 10) + parseInt(rgb[3], 10)
-  if (rgb < (255 * 3) / 2) out = 1
+  const threshold = 140
+  if (rgb < threshold * 3) out = 1
   if (bgcolor) _CACHE_CONSTRAST[bgcolor] = out
   return out
 }
--- a/umap/static/umap/js/umap.core.js
+++ b/umap/static/umap/js/umap.core.js
@@ -396,10 +396,28 @@ L.DomUtil.after = (target, el) => {
   return el
 }
 
+// From https://gist.github.com/Accudio/b9cb16e0e3df858cef0d31e38f1fe46f
+// convert colour in range 0-255 to the modifier used within luminance calculation
+L.DomUtil.colourMod = (colour) => {
+  const sRGB = colour / 255
+  let mod = Math.pow((sRGB + 0.055) / 1.055, 2.4)
+  if (sRGB < 0.03928) mod = sRGB / 12.92
+  return mod
+}
 L.DomUtil.RGBRegex = /rgb *\( *([0-9]{1,3}) *, *([0-9]{1,3}) *, *([0-9]{1,3}) *\)/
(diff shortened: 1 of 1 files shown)


Aurelie-Jallut:
We're good !
"A minimum contrast ratio of 3:1 between interface components or graphic elements carrying information and their background."
=> White / Grey (background) => contrast = 3.94
=> White / DimGrey (background) => contrat = 5.48

davidbgk (approved):

yohanboniface:
I tried to be more accurate in the choice, so I scripted a map with (named) colors close to the threshold (in parentesis the weight computed by umap, which is just a sum of RGB values).

Before this PR:

image

With threshold = 150

image

With threshold = 140

image

With my human eyes, I feel like 140 is the best choice, but I'd like your review again on this @davidbgk @Aurelie-Jallut :)

davidbgk:
We need to check contrast on the green markers at the bottom but it's easier from a live version, let's do that together :)

Aurelie-Jallut:
The number in brackets gives an indication of the weight of the color?
I don't quite understand the logic but I'm interested : )

I have the impression that there are colors above 140 that could accommodate a white picto like for example: Mediumpurple(159.33).
The color darkorange (131,67) on the other hand is at 1.37 in contrast with the white picto.

manuhabitela:
:wave: 

As Aurélie said the important thing is to have a 3:1 contrast ratio between icon and background. Usually green shades are kinda hard to match with white.

[Here is a totally overwhelming grid that checks contract of a bunch of colors of the final screenshot](https://contrast-grid.eightshapes.com/?version=1.1.0&background-colors=&foreground-colors=%23fff%2C%0D%0A%23000%2C%0D%0A%23808080%2C%0D%0A%23708090%2C%0D%0A%23bc8f8f%2C%0D%0A%23db7093%2C%0D%0A%239370db%2C%0D%0A%236a5acd%2C%0D%0A%234169e1%2C%0D%0A%2300bfff%2C%0D%0A%23ff8c00%2C%0D%0A%23cd853f%2C%0D%0A%23bdb76b%2C%0D%0A%239acd32%2C%0D%0A%2366cdaa%2C%0D%0A%23adff2f%2C%0D%0A%2300fa9a%2C%0D%0A%237fff00%2C%0D%0A%2300ff7f%2C%0D%0A%237cfc00&es-color-form__tile-size=compact&es-color-form__show-contrast=dnp) (note I didn't check all of them, focused on the ones with white text).

Don't worry you just have to look at the two first lines of the grid (or first two columns, as you wish). All the shown cells, which are the ones marked "DNP", do not have sufficient contrast with either black or white foreground.

Few ideas to be sure to generate OK colors:
- use a lib that uses the WCAG contrast ratio formula, like [color](https://github.com/Qix-/color), check the background color with both black and white foreground, and chose the one that has the better contrast ratio?
- do as what [Tanaguru](https://contrast-finder.tanaguru.com/result.html?foreground=%23FFFFFF&background=%2300FA9A&ratio=3&isBackgroundTested=true&algo=Rgb&distanceSort=asc) does and have a script return the "closest" accessible color to what you want

yohanboniface:
Mieux comme ça ? :)
Basé sur un algo de WGAG si j'ai bien tout lu Freud!

image

davidbgk (approved):

PR #1602, opened 2024-02-12
fix: custom storage for tests

davidbgk:
Setting `manifest_strict = False` will hopefully get rid of issues we have with the CI:

> ValueError: Missing staticfiles manifest entry for 'umap/favicons/icon-192.png'

See https://docs.djangoproject.com/en/5.0/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage.manifest_strict

### DIFF
--- a/umap/tests/integration/test_statics.py
+++ b/umap/tests/integration/test_statics.py
@@ -24,7 +24,7 @@ def test_javascript_have_been_loaded(
 ):
     settings.STORAGES["staticfiles"][
         "BACKEND"
-    ] = "umap.storage.UmapManifestStaticFilesStorage"
+    ] = "umap.storage.UmapManifestStaticFilesStorageNotStrict"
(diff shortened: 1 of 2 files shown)


umap/tests/integration/test_statics.py
```diff
@@ -24,7 +24,7 @@ def test_javascript_have_been_loaded(
 ):
     settings.STORAGES["staticfiles"][
         "BACKEND"
-    ] = "umap.storage.UmapManifestStaticFilesStorage"
+    ] = "umap.storage.UmapManifestStaticFilesStorageNotStrict"
```
yohanboniface, line 27: I don't think this line is even used: given the staticfiles is run in a fixture, changing the setting at this time will not do anything. So I think the current test is already broken.

PR #1607, opened 2024-02-13
chore: migrate to OAuth 2 for OpenStreetMap backend

### LINKED ISSUE #1279: Move from OAuth 1.0a to OAuth 2.0 for OSM logins
OSM's OAuth 1.0a is in the process of being deprecated, as discussed at https://github.com/openstreetmap/operations/issues/867. No timeline has been set, but we do not expect to shut off OAuth 1.0a this year. It would be good to move to OAuth 2.0 well before this time.

yohanboniface:
fix #1279


davidbgk (commented):
Don't we have to change / document the `AUTHENTICATION_BACKENDS` setting too?

PR #1613, opened 2024-02-16
fix: star control was using old post method

yohanboniface:
Also, only display it if the user is already logged in, given we do not deal anymore with login during ajax calls.


davidbgk (approved):
Yay for tests 😎

PR #1624, opened 2024-02-19
chore: raise error if any in storage post_process

yohanboniface:
I'm not totally sure it's the way to go (maybe ignoring here is fine ?), but it seems still more usefull to have a proper error instead of a python error a few lines forward.


davidbgk (approved):
😈

PR #1641, opened 2024-02-21
a11y: use role=alert for messages from Django and JS

davidbgk:
Also define a custom module+css for alerts (chore).

This is an early stage, more a discussion at this point.
It works with messages from Django and set dynamically via JS.
The old code is still present.

See:
* https://www.w3.org/WAI/ARIA/apg/patterns/alert/
* https://www.w3.org/WAI/ARIA/apg/patterns/alert/examples/alert/
* https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Alert_Role

### DIFF
--- a/umap/static/umap/js/modules/alerts.js
+++ b/umap/static/umap/js/modules/alerts.js
@@ -1,0 +1,54 @@
+export default class Alerts {
+  constructor() {
+    this.alertNode = document.querySelector('[role="alert"]')
+    const observer = new MutationObserver(this._callback.bind(this))
+    observer.observe(this.alertNode, { childList: true })
+    // On initial page load, we want to display messages from Django.
+    Array.from(this.alertNode.children).forEach(this._display.bind(this))
+  }
+
+  _callback(mutationList, observer) {
+    for (const mutation of mutationList) {
+      this._display(
+        [...mutation.addedNodes].filter((item) => item.tagName === 'P').pop()
+      )
+    }
+  }
+
+  _display(alert) {
+    const duration = alert.dataset?.duration || 3000
+    const level = alert.dataset?.level || 'info'
+    const wrapper = document.createElement('div')
+    const alertHTML = alert.cloneNode(true).outerHTML
+    wrapper.innerHTML = `
+    <div data-level="${level}" data-alert data-toclose>
+      ${alertHTML}
+      <button class="umap-close-link" type="button" data-close>
+        <i class="umap-close-icon"></i><span>${L._('Close')}</span>
+      </button>
+    </div>
+    `
+    const alertDiv = wrapper.firstElementChild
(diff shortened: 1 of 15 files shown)


davidbgk:
The `error` level should probably trigger an `Infinite` duration 🤔

davidbgk:
There are two remaining complex cases where there are buttons and logic within the alert:

* to send a link upon anonymous map creation 
* in case of conflict with data from the server when we save

umap/static/umap/js/modules/alerts.js
```diff
@@ -0,0 +1,54 @@
...
+    observer.observe(this.alertNode, { childList: true })
+    // On initial page load, we want to display messages from Django.
+    Array.from(this.alertNode.children).forEach(this._display.bind(this))
+  }
+
+  _callback(mutationList, observer) {
+    for (const mutation of mutationList) {
+      this._display(
+        [...mutation.addedNodes].filter((item) => item.tagName === 'P').pop()
+      )
```
almet, line 14: Why do we need to `.pop()` the element if we are already filtering it? 🤔

umap/static/umap/js/modules/alerts.js
```diff
@@ -0,0 +1,54 @@
...
+    <div data-level="${level}" data-alert data-toclose>
+      ${alertHTML}
+      <button class="umap-close-link" type="button" data-close>
+        <i class="umap-close-icon"></i><span>${L._('Close')}</span>
+      </button>
+    </div>
+    `
+    const alertDiv = wrapper.firstElementChild
+    this.alertNode.after(alertDiv)
+    if (isFinite(duration)) {
```
almet, line 33: Love using `Infinity` + `isFinite` 🤝

almet (approved):

umap/static/umap/js/modules/alerts.js
davidbgk: `filter()` might return multiple element and we want the first one.

umap/static/umap/js/modules/alerts.js
```diff
@@ -0,0 +1,54 @@
...
+      setTimeout(() => {
+        alertDiv.remove()
+      }, duration)
+    }
+  }
+
+  add(message, level = 'info', duration = 3000) {
+    this.alertNode.innerHTML = `
```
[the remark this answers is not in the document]
davidbgk: That's indeed a concern, I was wondering if that case actually happens 🤔

umap/static/umap/js/modules/alerts.js
almet: Thanks, I actually misread, and was trying to understand what was being poped-out (because it's removed from the list). Turns out it's not a problem here because the whole list only exists for the duration of the event.

PR #1645, opened 2024-02-21
fix: zoomControl rendered twice

### LINKED ISSUE #1644: Zoom control appears twice when map is embedded or shared (customised view)
**Describe the bug**
When a map is shared via the option "share this link to open a customised map view, using the default link provided, the + and - zoom controls appear twice in the top left of the shared map.
If the option "Display zoom control" is set to "Never", the zoom control is shown once.

The same happens when embedding the map - that was how I first noticed it, as the duplicated controls take up a lot of space in a small iframe for embedding in a blog.

It does not seem to depend on any particular map being edited, it happens on a completely new map, whether signed in or not.

**To Reproduce**
1. Go to https://umap.openstreetmap.fr/en/
2. Click on 'Create a Map.'
3. Click the Share icon
4. Go to the "Direct Link" option near the bottom and copy the default link provided, e.g. https://umap.openstreetmap.fr/en/map/new/?scaleControl=false&miniMap=false&scrollWheelZoom=false&zoomControl=true&editMode=disabled&moreControl=true&searchControl=null&tilelayersControl=null&embedControl=null&datalayersControl=true&onLoadPanel=undefined&captionBar=false&captionMenus=true
5. Paste the link into a browser. The icons at top left are + - + -. 

**Expected behavior**
The zoom control should only appear once given the default setting, and should respond appropriately to the Hidden and Never settings.

**Screenshots**
image

An example of an embedded map from the project documentation at https://docs.framasoft.org/fr/umap/ - you can see how the duplicated controls eat up space on the left:
image

 - Windows 11
 - Firefox, Edge

yohanboniface:
fix #1644

### DIFF
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -125,18 +125,12 @@ U.Map = L.Map.extend({
     // To be used in javascript APIs
     if (geojson.properties.lang) L.lang = geojson.properties.lang
 
-    // Don't let default autocreation of controls
-    const zoomControl =
-      typeof geojson.properties.zoomControl !== 'undefined'
-        ? geojson.properties.zoomControl
-        : true
-    geojson.properties.zoomControl = false
-    const fullscreenControl =
-      typeof geojson.properties.fullscreenControl !== 'undefined'
-        ? geojson.properties.fullscreenControl
-        : true
-    geojson.properties.fullscreenControl = false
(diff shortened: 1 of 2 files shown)


jschleic (approved):
Looks good and I can confirm that it fixes the duplicated zoom controls.

yohanboniface:
Thanks! I'll do a patch release soon!

PR #1648, opened 2024-02-22
chore: format templates

almet:
I'm not sure why this is not failing on master. I wonder if something is off with my machine.

### DIFF
--- a/umap/tests/test_map_views.py
+++ b/umap/tests/test_map_views.py
@@ -818,10 +818,10 @@ def test_oembed_map(client, map, datalayer):
 def test_oembed_link(client, map, datalayer):
     response = client.get(map.get_absolute_url())
     assert response.status_code == 200
+
     assert (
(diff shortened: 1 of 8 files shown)


umap/tests/test_map_views.py
```diff
@@ -818,10 +818,10 @@ def test_oembed_map(client, map, datalayer):
 def test_oembed_link(client, map, datalayer):
     response = client.get(map.get_absolute_url())
     assert response.status_code == 200
+
     assert (
-        '<link rel="alternate" type="application/json+oembed"'
-        in response.content.decode()
-    )
+        '<link rel="alternate"\n        type="application/json+oembed"'
```
almet, line 823: Not a big fan of having spaces in the tests here. Any suggestion welcome :-)

davidbgk: Maybe applying [dedent](https://docs.python.org/3/library/textwrap.html#textwrap.dedent) to the deocded response?

PR #1649, opened 2024-02-22
chore: add ids on features

almet:
Add an ID for each feature on creation, or when read from a GeoJSON.

Ids are persisted in the GeoJSON

### DIFF
--- a/umap/static/umap/js/umap.features.js
+++ b/umap/static/umap/js/umap.features.js
@@ -45,7 +45,7 @@ U.FeatureMixin = {
     this.addInteractions()
     this.parentClass.prototype.initialize.call(this, latlng, options)
   },
-  _checkId: function () {
+  _checkId: function (string) {
     return typeof string !== 'undefined'
   },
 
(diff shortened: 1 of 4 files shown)


umap/static/umap/js/umap.features.js
```diff
@@ -36,6 +45,9 @@ U.FeatureMixin = {
     this.addInteractions()
     this.parentClass.prototype.initialize.call(this, latlng, options)
   },
+  _checkId: function () {
```
yohanboniface, line 48: Missing a parameter ?

umap/static/umap/js/umap.features.js
```diff
@@ -36,6 +45,9 @@ U.FeatureMixin = {
     this.addInteractions()
     this.parentClass.prototype.initialize.call(this, latlng, options)
   },
+  _checkId: function (string) {
+    return typeof string !== 'undefined'
```
yohanboniface, line 49: Shouldn't we check that the id is in our internal format and not something else ? Or do we want to consume any non null string ? So people importing in uMap could control their ids ?

almet: You're correct. I've added a check to ensure ids are matching the expected format.

PR #1653, opened 2024-02-25
path was doubled

datendelphin:
If I wanted to import path "maki", that tried to open "maki/maki/pictogram.svg"


yohanboniface:
Thanks! :)

I did not hit this bug because I was using absolute path, but your fix sounds good.

PR #1665, opened 2024-02-29
fix: allow empty datalayers reference on merges.

almet:
Previously, an error was thrown when the reference datalayer had no `features`defined.

When looking for features, it now defaults to an empty list if the key doesn't exist.


davidbgk (approved):
👍 good catch!

PR #1666, opened 2024-02-29
fix: Replace Last-Modified with custom headers

almet:
`X-Datalayer-Version` and `X-Datalayer-Reference` are now used instead
of the `Last-Modified` and `If-Unmodified-Since` headers.

`Last-Modified` is granular to the second, which led to problems with
the versionning. The new system uses timestamps instead.

This commit also changes the way versions were created. Previously,
the associated version was coming from two different places:
the last modified time from the filesystem and a `time.time()` call
done when saving the model, which could result in the two getting
out of sync.

> [!NOTE]
> **This still has one issue** Requests coming at the same time can lead to conflicts. It's not clear if we want to create a lock system to handle this, or if we just consider this as an edge case.

-> It seems we want to consider this an edge-case for now, so we're good to go 👍 

## Todo

- [x] Make the tests pass
- [x] Update the headers to use CamelCase everywhere
- [x] Rename the headers to `X-Datalayer-Version` and `X-Datalayer-Reference-Version`

### DIFF
--- a/umap/static/umap/js/umap.layer.js
+++ b/umap/static/umap/js/umap.layer.js
@@ -677,7 +677,7 @@ U.DataLayer = L.Evented.extend({
     this._loading = true
     const [geojson, response, error] = await this.map.server.get(this._dataUrl())
     if (!error) {
-      this._last_modified = response.headers.get('last-modified')
+      this._reference_version = response.headers.get('X-Datalayer-Version')
       // FIXME: for now this property is set dynamically from backend
       // And thus it's not in the geojson file in the server
       // So do not let all options to be reset
@@ -1459,7 +1459,7 @@ U.DataLayer = L.Evented.extend({
     if (!this.isVisible()) return
     const bounds = this.layer.getBounds()
     if (bounds.isValid()) {
-      const options = {maxZoom: this.getOption("zoomTo")}
+      const options = { maxZoom: this.getOption('zoomTo') }
       this.map.fitBounds(bounds, options)
     }
   },
@@ -1577,8 +1577,8 @@ U.DataLayer = L.Evented.extend({
       map_id: this.map.options.umap_id,
       pk: this.umap_id,
     })
-    const headers = this._last_modified
-      ? { 'If-Unmodified-Since': this._last_modified }
+    const headers = this._reference_version
+      ? { 'X-Datalayer-Reference': this._reference_version }
       : {}
     await this._trySave(saveUrl, headers, formData)
     this._geojson = geojson
@@ -1596,7 +1596,7 @@ U.DataLayer = L.Evented.extend({
             label: L._('Save anyway'),
             callback: async () => {
               // Save again,
-              // but do not pass If-Unmodified-Since this time
+              // but do not pass the reference version this time
               await this._trySave(url, {}, formData)
             },
           },
@@ -1619,7 +1619,7 @@ U.DataLayer = L.Evented.extend({
         this.fromGeoJSON(data.geojson)
         delete data.geojson
       }
-      this._last_modified = response.headers.get('last-modified')
+      this._reference_version = response.headers.get('X-Datalayer-Version')
       this.setUmapId(data.id)
       this.updateOptions(data)
       this.backupOptions()
--- a/umap/tests/integration/test_collaborative_editing.py
+++ b/umap/tests/integration/test_collaborative_editing.py
@@ -196,3 +196,72 @@ def test_empty_datalayers_can_be_merged(context, live_server, tilelayer):
         sleep(1)
 
     expect(marker_pane_p2).to_have_count(2)
+
+
+def test_same_second_edit_doesnt_conflict(context, live_server, tilelayer):
+    # Let's create a new map with an empty datalayer
+    map = MapFactory(name="collaborative editing")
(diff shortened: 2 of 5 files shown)


almet:
This is using a specific in house versioning system rather than `Etags`, to avoid conflicts with the way gzip is handled for datalayers.

We might want at some point to get rid of the custom headers and replace them with Etags, but that's probably for later :-)

umap/views.py
```diff
@@ -958,20 +958,19 @@ class GZipMixin(object):
...
 
     @property
     def gzip_path(self):
         return Path(f"{self.path}{self.EXT}")
 
-    def compute_last_modified(self, path):
-        stat = os.stat(path)
-        return http_date(stat.st_mtime)
+    def read_version(self, path):
+        return path.stem.split("_")[-1].replace(".geojson", "")
```
davidbgk, line 968:
```suggestion
        # Remove optional .gz, then .geojson, then return the trailing version from path.
        return path.with_suffix("").with_suffix("").split("_")[-1]
```

A non-blocking suggestion I find a bit more readable. Even with the original proposal, a comment making explicit that the `path` can be either .gz suffixed or not might be useful (it took me some time to figure it out).

umap/views.py
```diff
@@ -1012,7 +1011,7 @@ def render_to_response(self, context, **response_kwargs):
             with open(path, "rb") as f:
                 # Should not be used in production!
                 response = HttpResponse(f.read(), content_type="application/geo+json")
-            response["Last-Modified"] = self.last_modified
+            response["X-Umap-Version"] = self.version
```
davidbgk, line 1014: Why is that UPPERCASED in JS and CamelCased in Python? Is that a convention I miss?

umap/views.py
almet: +1 for the comment and the change 👍

umap/views.py
almet: Thanks for pointing it out. I started with uppercase everywhere, and then went to camel case, but didn't update it everywhere.

umap/static/umap/js/umap.layer.js
```diff
@@ -677,7 +677,7 @@ U.DataLayer = L.Evented.extend({
     this._loading = true
     const [geojson, response, error] = await this.map.server.get(this._dataUrl())
     if (!error) {
-      this._last_modified = response.headers.get('last-modified')
+      this._reference_version = response.headers.get('X-Umap-Version')
```
yohanboniface, line 680: nit: `umap-version` seems a bit like it's the version of uMap itself. What about `x-Data-Version` or `X-DataLayer-Version` ?

umap/tests/integration/test_collaborative_editing.py
```diff
@@ -196,3 +196,72 @@ def test_empty_datalayers_can_be_merged(context, live_server, tilelayer):
...
+
+    # Click on the map, it will place a marker at the given position.
+    map_el_p2 = page_two.locator("#map")
+    map_el_p2.click(position={"x": 220, "y": 220})
+    expect(marker_pane_p2).to_have_count(1)
+
+    # Save the two tabs at the same time
+    with page_one.expect_response(DATALAYER_UPDATE):
+        save_p1.click()
+        sleep(0.1)  # Needed to avoid
```
yohanboniface, line 247: To avoid ? :)

umap/views.py
```diff
@@ -1012,17 +1012,16 @@ def render_to_response(self, context, **response_kwargs):
...
         return response
 
 
 class DataLayerVersion(DataLayerView):
     @property
     def path(self):
-        return "{root}/{path}".format(
-            root=settings.MEDIA_ROOT,
-            path=self.object.get_version_path(self.kwargs["name"]),
+        return Path(settings.MEDIA_ROOT) / self.object.get_version_path(
```
yohanboniface, line 1023: Maybe do this once and for all in `settings/__init__.py` ?

umap/views.py
```diff
@@ -1033,42 +1032,41 @@ class DataLayerCreate(FormLessEditMixin, GZipMixin, CreateView):
...
         return response
 
 
 class DataLayerUpdate(FormLessEditMixin, GZipMixin, UpdateView):
     model = DataLayer
     form_class = DataLayerForm
 
-    def has_been_modified_since(self, if_unmodified_since):
-        return if_unmodified_since and self.last_modified != if_unmodified_since
+    def has_changes_since(self, incoming_version):
```
yohanboniface, line 1047: Given we are using version number and not time anymore, maybe rename to `match_version`  or `is_latest`/`is_latest_version`?

yohanboniface (approved):

umap/static/umap/js/umap.layer.js
almet: I went with `X-DataLayer-Version` and `X-DataLayer-Reference`.

umap/views.py
almet: I prefer `has_changes_since` because we're passing the argument with the version. With the propositions you're making, it's less clear to me that it's supposed to be an argument passed.

umap/views.py
almet: Why not, but I prefer to have this in a separate pull request if possible :-)

davidbgk (approved):

PR #1687, opened 2024-03-12
feat: merge "datalayer browser" and "data browser"

### LINKED ISSUE #330: [UX issue] No direct way to get back to layers manager
It's currently not really easy (or at least not intuitive) how to get back to the list of layers when (for example) creating a new layer.

Steps to reproduce:
- open the layers manager,
- click on the "add new layer" button,
- modify the name and description of the layer.

In order for example to modify the order of the layers or change the description of an other layer, user currently need to click on the "manage layers" again.
Users may try to click on the "close button" and get the right panel to be hidden.

yohanboniface:
Small big PR!

Previously, we had both a "datalayers browser" and a "data browser", now those two are merged.

The datalayers browser looked like this:

image

The data browser looked like this:

image

The new browser has two modes, an expanded mode and a condensed mode.

Expended mode, which is roughly the same as the previous data browser:

image

In condensed mode:

image

One can switch from those two mode by a button in the panel:

image

Once the panel is in a given mode, it will stay in this mode unless the user changes it.

The button to open the browser is now on the right of the screen (while it was on the left before), and a new "caption" button appears:

image

In edit mode, the browser stays, and replaces the previous box to list the layers:

image

Again, it can be either in condensed mode or in expanded mode:

image

When editing a given layer, a new "back" button appears, to go back to the browser (fix #330):

image


EduNetEurope:
Looks great! 
Can we still display it on the right sight?

VierPaarVoeten:
Since the update it is impossible to change the color of the layer?

yohanboniface:
> Since the update it is impossible to change the color of the layer?

Can you be more explicit ?

yohanboniface:
>  Can we still display it on the right sight?

This is not planned. Can you share your use case ? :)

VierPaarVoeten:
[
Screenshot_20240508-210020

](url)

EduNetEurope:
> > Can we still display it on the right sight?
> 
> This is not planned. Can you share your use case ? :)

Thank you for the reply.
I guess we have all vertical menus on the right side and, personally, I find it more stylish to have it on the right side.
But it is a personal preference for sure.

yohanboniface:
> I would like to change the color like I did in the past. But can't find the
correct panel to do. 

Sorry, I still don't understand your issue. You want to change the layer color, correct ? Did you click on "Bewerken", then "Manager Layers", can't you edit your layer as always ?

VierPaarVoeten:
> > I would like to change the color like I did in the past. But can't find the
> > correct panel to do.
> 
> Sorry, I still don't understand your issue. You want to change the layer color, correct ? Did you click on "Bewerken", then "Manager Layers", can't you edit your layer as always ?

No I can't edit my layer as always. You can see it right [here](https://youtu.be/JkqV5JWC63I).

yohanboniface:
humm, you have to click here:

image

Or I still don't get your point :upside_down_face:

VierPaarVoeten:
There you van change the color of the layer. It is not clear because it
seems you change the color of the pin instead of the layer.

Thanks!

Op do 9 mei 2024 16:01 schreef Yohan Boniface ***@***.***>:

PR #1692, opened 2024-03-15
refactor: Separate ui rendering from data updates

almet:
`U.SCHEMA` now contains an `impacts` key, which makes it possible to specify what part of the UI is impacted by data changes.

A new `render` method has been added on `U.Map` and `U.DataLayer`, which is used to rerender the proper parts of the UI depending on the passed properties.

`U.FormBuilder` calls this `render()` method (if present), during form changes.

> [!Note]
> This approach is less specific than [a previous one](https://github.com/umap-project/umap/compare/master...feat/simple-crdts) in order to a) not tie the schema to specific callbacks and b) not repeat properties in two different places.
> 
> The resulting code is less conservative when considering which parts of the UI are updated when receiving updates, thus rerendering parts of the UI which might not be in need of rerender. The hope is that it won't be a problem, but let's keep our eyes open.

## Progress

- [x] Sort schema keys
- [x] Add all UI impacts on schema keys
- [x] Update FormBuilder instanciations to remove callbacks when possible
- [x] Ensure the tests pass
- [x] `ChoroplethLayer.onEdit` is expecting to be passed the builder
- [x] Use `impacts = []` everywhere (remove `['']` occurrences)
- [x] Change how the functional tests are working (see #1718 #1719) 
- [x] Make the JS unit tests pass
- [x] Make the functional tests pass


umap/static/umap/js/modules/schema.js
```diff
@@ -1,45 +1,52 @@
...
+  datalayersControl: {
+    impacts: ['ui'],
+    type: Boolean,
+    nullable: true,
+    handler: 'DataLayersControl',
+    label: translate('Display the data layers control'),
+    default: true,
   },
   defaultView: {
+    impacts: [], // no need to update the ui, only useful when loading the map
```
davidbgk, line 49: Is there a difference between `['']` (above) and `[]`? Maybe update the `// Possible impacts` comment to explain that?

umap/static/umap/js/umap.js
```diff
@@ -240,6 +241,42 @@ U.Map = L.Map.extend({
     this.on('click contextmenu.show', this.closeInplaceToolbar)
   },
 
+  render: function (fields) {
+    let impacts = U.Utils.getImpactsFromSchema(fields)
```
davidbgk, line 245: If the schema is static, don't we want to compute that once and for all? (vs. at each `render()` call)

umap/static/umap/js/modules/schema.js
almet: No, there shouldn't be, and the list should be updated to only have `[]`.

umap/static/umap/js/umap.js
almet: The schema is static but the list of passed fields isn't.`getImpactsFromSchema` goes trough the keys of the schema, returning what is impacted by the given fields.

I don't think the lookup will take much time. If it is, we might want to cache out the results, maybe. I'd say it's an optimisation for the future, when/if we need it. Would that be okay for you?

davidbgk: Alright 👍

umap/static/umap/js/modules/utils.js
```diff
@@ -22,3 +22,20 @@ export function checkId(string) {
   if (typeof string !== 'string') return false
   return /^[A-Za-z0-9]{5}$/.test(string)
 }
+
+export function getImpactsFromSchema(fields) {
```
yohanboniface, line 26: I'd make a few unitests (JS) for this one :)

umap/static/umap/js/umap.js
```diff
@@ -240,6 +241,42 @@ U.Map = L.Map.extend({
...
+          break
+      }
+    }
+  },
+
+  reindexEachDataLayer: function () {
+    this.eachDataLayer((datalayer) => datalayer.reindex())
+  },
+
+  renderVisibleDataLayers: function () {
```
yohanboniface, line 274: `redrawVisibleDataLayers` ?

umap/static/umap/js/umap.layer.js
```diff
@@ -594,6 +596,31 @@ U.DataLayer = L.Evented.extend({
     if (this.autoLoaded()) this.map.on('zoomend', this.onZoomEnd, this)
   },
 
+  render: function (fields, builder) {
```
yohanboniface, line 599: Maybe we could call the method `onDataChanged` to make it more explicit that this is not the default redering but a sort of "reflow" ?

umap/static/umap/js/modules/schema.js
```diff
@@ -1,45 +1,52 @@
...
-  miniMap: {
-    type: Boolean,
-    label: translate('Do you want to display a minimap?'),
-    default: false,
-  },
-  displayPopupFooter: {
-    type: Boolean,
-    label: translate('Do you want to display popup footer?'),
-    default: false,
+  color: {
```
yohanboniface, line 24: I suggest to add a dedicated `test_on_edit.py` file with some basic test (at least one per "impact"), that we can then complete later with some edge cases.

umap/static/umap/js/umap.forms.js
```diff
@@ -1033,6 +1033,7 @@ U.FormBuilder = L.FormBuilder.extend({
   setter: function (field, value) {
     L.FormBuilder.prototype.setter.call(this, field, value)
     if (this.options.makeDirty !== false) this.obj.isDirty = true
+    if ('render' in this.obj) this.obj.render([field], this)
```
yohanboniface, line 1036: this may call render on map when searching in the browser. So either this is only done when `makeDirty` is false, either we use a dedicated Builder class for the browser (simpler imho).

umap/static/umap/js/umap.js
```diff
@@ -240,6 +241,42 @@ U.Map = L.Map.extend({
...
+          this.initTileLayers()
+          break
+        case 'bounds':
+          this.handleLimitBounds()
+          break
+      }
+    }
+  },
+
+  reindexEachDataLayer: function () {
```
yohanboniface, line 270: `reindexDataLayers` ?

umap/static/umap/js/umap.layer.js
```diff
@@ -254,16 +254,18 @@ U.Layer.Choropleth = L.FeatureGroup.extend({
   },
 
   onEdit: function (field, builder) {
+    // Only compute the breaks if we're dealing with choropleth
+    if (!field.startsWith('options.choropleth')) return
     // If user touches the breaks, then force manual mode
     if (field === 'options.choropleth.breaks') {
       this.datalayer.options.choropleth.mode = 'manual'
-      builder.helpers['options.choropleth.mode'].fetch()
+      if (builder) builder.helpers['options.choropleth.mode'].fetch()
```
yohanboniface, line 262: Ugly but let's go with that for now! ;)

yohanboniface (commented):
:+1:

umap/static/umap/js/umap.forms.js
almet: I don't think we should rely on `makeDirty` as I think it will be out of the picture at some point.

Using a dedicated browser class seem the way to go :-)

umap/static/umap/js/umap.js
almet: 👍

umap/static/umap/js/umap.js
almet: 👍

umap/static/umap/js/umap.layer.js
almet: I would prefer to keep the `on*` methods for events. (`onDataChanged` would imply there is an `data-changed` event that occurred). 

I like the idea of using `render` somewhere, because it's really the intent here. Maybe `rerenderFields`? (it also  makes it more explicit that's it's expecting fields to be passed).

umap/static/umap/js/umap.forms.js
almet: In the end, I did the opposite: this call only happens if `makeDirty` is not explicitely set to `false`.

From where I stand right now, it seems to avoid adding complexity for no almost no benefits. We can take care of doing things differently if/when we remove the `dirty` mechanism.

umap/static/umap/js/modules/utils.js
almet: ✔️

almet:
This is ready to be merged!

umap/static/umap/js/modules/schema.js
```diff
@@ -1,46 +1,61 @@
...
   },
-  scaleControl: {
+  captionBar: {
     type: Boolean,
-    label: translate('Do you want to display the scale control?'),
-    default: true,
+    impacts: ['ui'],
+    label: translate('Do you want to display a caption bar?'),
+    default: false,
+    impacts: [],
```
davidbgk, line 16: Two different `impacts` keys for the captionBar?

almet: 🤫

umap/tests/integration/test_on_edit.py
```diff
@@ -0,0 +1,185 @@
+import re
```
yohanboniface, line 1: Maybe mv those tests in `test_edit_map.py` instead of having a new module ?

umap/static/umap/test/Util.js
```diff
@@ -1,397 +1,4 @@
 describe('L.Util', function () {
```
yohanboniface, line 1: I'm surprised to see all those tests removed here ? Should this have been done in that other PR moving utils to a module ? Or is this just missing a rebase ?

umap/static/umap/js/umap.layer.js
yohanboniface: ok, let's keep `render`!

yohanboniface (approved):

umap/static/umap/test/Util.js
almet: yep, it's broken in the current master, and fixed here! 🙈

PR #1733, opened 2024-04-05
Redirect to the login view when login is required

davidbgk:
When the `UMAP_ALLOW_ANONYMOUS` setting is False, we return the login url through a JSON response. We lost that ability in #1555.

The JS part was not following that link in that particular case and lead to more errors because the map was not saved (hence, no `map_id`).

For now, the current work on the map is lost because of the redirection and we have a confirmation dialog to quit the edited page with unsaved changes.

Maybe we should display a custom message instead of a brutal redirection? Like: you’re not logged in, do it in a separate tab to keep you work? (A bit ugly…)

Sidenote: we might want to use the `redirect` pattern/key in the JSON response that we already use for deletion and clone for consistency.

### DIFF
--- a/umap/decorators.py
+++ b/umap/decorators.py
@@ -19,7 +19,9 @@ def login_required_if_not_anonymous_allowed(view_func):
             not getattr(settings, "UMAP_ALLOW_ANONYMOUS", False)
             and not request.user.is_authenticated
         ):
-            return simple_json_response(login_required=str(LOGIN_URL))
+            response = simple_json_response(login_required=str(LOGIN_URL))
+            response.status_code = 401
+            return response
         return view_func(request, *args, **kwargs)
 
     return wrapper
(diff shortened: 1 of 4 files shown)


almet:
Thanks for the issue. I wonder: if the login is required, maybe we should login first before allowing the creation of a new map? 🤔

yohanboniface:
Humm, but with this proposal, the data is lost, no ?

The tricky part, that was done before #1555, was:
- sent a normal request
- receive a login_required
- to the login
- send again the initial request
- continue the js flow (sending other requests, etc.)


> I wonder: if the login is required, maybe we should login first before allowing the creation of a new map? 🤔

There is an usage where users see a "prebuilt" map with remote data, so they can browse the data AND optionally save it later. Those users are not logged in. This is in use in the ANCT instance.: from the deveco app, people search for companies in their territory, and if they want to see it on a map, they are sent to uMap.

yohanboniface:
Two questions then:
- is "playground mode" the situation where the uMap instance does not accept anonymous map and the user is not logged in ? (We should still keep in mind the "preloaded map" scenario)
- does that mean we gived up on making the login working again during a save process  (as before #1555) ?

davidbgk:
image(1)

To have it here too.

yohanboniface:
Replaced by #2025

Issue #1710, opened 2024-03-25
Text formatting in the description field


Aurelie-Jallut:
**Is your feature request related to a problem? Please describe.**
The text displayed in the help modal gives the impression that it's going to be complicated to create.
the term Iframe, for example, is not understood by non-tech users.

![image](https://github.com/umap-project/umap/assets/131352601/4831607b-3e7d-400d-b0ee-787f46ec5cbf)

**Describe the solution you'd like**
We could add buttons for styling in Markdown

**Describe alternatives you've considered**
We could improve the text in the modal

PeskyPotato:
Related to #813

yohanboniface:
Closing in favor of #1950

Issue #813, opened 2020-06-26
text formatting infobubble is difficult to read


Hugo-Trentesaux:
When entering a description for a marker, a help bubble displays:

```
Text formatting

    *simple star for italic*
    **double star for bold**
    # one hash for main heading
    ## two hashes for second heading
    ### three hashes for third heading
    Simple link: [[http://example.com]]
    Link with text: [[http://example.com|text of the link]]
    Image: {{http://image.url.com}}
    Image with custom width (in px): {{http://image.url.com|width}}
    Iframe: {{{http://iframe.url.com}}}
    Iframe with custom height (in px): {{{http://iframe.url.com|height}}}
    Iframe with custom height and width (in px): {{{http://iframe.url.com|height*width}}}
    --- for an horizontal rule
```

This is difficult to read for inexperienced users not familiar with this kind of markup. Maybe setting the special characters to bold could help identifying them. It has been reported to me that the vertical bar in the "Link with text" is difficult to see.

**[[** http://example.com **|** text of the link **]]** or even color would help to understand.

Spielmops:
Zoom in with the browser like in any webpage: CTRL and + or CTRL and mousewheel

Hugo-Trentesaux:
Zooming in will not highlight the markdown signs as I suggest. Bold font should be applied instead.

Spielmops:
You are right, but if I zoom in, everything is good to read. On the other hand: a little bit of color enhances the readability - as known from editors with syntax-highlighting ...
Spielmops

PeskyPotato:
I've attempted to make the modal description a bit more verbose so unfamiliar users get a better understanding. I've also added more formatting to the text to improve readability. Not sure if this would help at all but a starting point.

![screenshot of modal each category has a header, description and example](https://github.com/umap-project/umap/assets/35942108/f87ac78b-7df3-4340-9a18-55f92f6b5b2c)

This would mean the translations need to be redone in each language again.

PeskyPotato:
> I'm a bit afraid about the size of the box with such details 

This was my concern as well, on smaller screens it can be a bit much to have all that text. I'll try and condense it a little bit without trying to take out any details to see how it looks, later today. Maybe the solution might just bee to have a similar output to what was there before but a link to the documentation with a more verbose description of each.

Spielmops:
Thank you all, looks much better now.

Spielmops

yohanboniface:
[...]

PR #1747, opened 2024-04-15
First step for documentation translations

davidbgk:
The goal is to be able to embed some of the FAQ entries within the app, see #1710 and #813 for instance.


almet:
Super :-)

yohanboniface:
Very cool!

I wonder if we should keep developer and user docs in the same place ? I'd say that the developer doc may appear noise for users, and also I don't think we want to translate the developer docs (translations will never be up to date imho).

PR #1756, opened 2024-04-17
feat: add link in contextmenu to edit map in OSM

### LINKED ISSUE #1651: Open iD or JSOM editor via right click menu
The Open iD/JOSM map menu on uMap is quite handy! However, it could be greatly improved by allowing access to shortcuts through a convenient right-click.

Currently, as I use uMap to address OSM quality gaps in iD, the ability to right-click anywhere on the map and smoothly navigate to the clicked coordinates in iD would significantly enhance the overall user experience. 😊

Big plus if it would jump automatically to the minimum zoom required in iD (16) so the user does not have this extra step to click on "+ Zoom in to edit"

yohanboniface:
fix #1651

image

### DIFF
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -1769,7 +1769,22 @@ U.Map = L.Map.extend({
         callback: this.openExternalRouting,
       })
     }
+    if (this.options.urls.edit_in_osm) {
+      items.push('-', {
+        text: L._('Edit in OpenStreetMap'),
+        callback: this.editInOSM,
+      })
+    }
     this.options.contextmenuItems = items
+  },
+
+  editInOSM: function (e) {
+    const url = this.urls.get('edit_in_osm', {
+      lat: e.latlng.lat,
+      lng: e.latlng.lng,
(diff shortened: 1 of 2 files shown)


umap/static/umap/js/umap.js
```diff
@@ -1769,9 +1769,24 @@ U.Map = L.Map.extend({
...
     this.options.contextmenuItems = items
   },
 
+  editInOSM: function (e) {
+    const url = this.urls.get('edit_in_osm', {
+      lat: e.latlng.lat,
+      lng: e.latlng.lng,
+      zoom: Math.max(this.getZoom(), 16),
+    })
+    if (url) window.open(url)
```
davidbgk, line 1787: Not specific to that PR but I think we should open a new tab/window in these cases (probably applies to `openExternalRouting` too):

```suggestion
    if (url) window.open(url, '_blank')
```

davidbgk (approved):

umap/static/umap/js/umap.js
yohanboniface: That's a good point, but strangely it already opens in a new window (same for routing).

PR #1759, opened 2024-04-17
fix: save back map description in DB

yohanboniface:
Since a few years, we are not saving description in DB anymore. I'm not sure it's a wanted change.

Other option, given we already have it in the `settings` JSONB column, is to remove the `description` field from the model and make it a `@property` instead that will try to read in the `settings` column


yohanboniface:
Closed in favor of #1772

PR #1760, opened 2024-04-17
fix: do not try to close editPanel if not set

yohanboniface:
When in the user dashboard and opening a map preview, when we press escape it will try to close non existing panels (they are not instantiated given there is not edit mode there)

### DIFF
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -524,8 +524,8 @@ U.Map = L.Map.extend({
           this.help.hide()
         } else {
           this.panel.close()
-          this.editPanel.close()
-          this.fullPanel.close()
+          this.editPanel?.close()
+          this.fullPanel?.close()
         }
       }
 


umap/static/umap/js/umap.js
```diff
@@ -524,8 +524,8 @@ U.Map = L.Map.extend({
           this.help.hide()
         } else {
           this.panel.close()
-          this.editPanel.close()
-          this.fullPanel.close()
+          if (this.editPanel) this.editPanel.close()
```
davidbgk, line 527:
```suggestion
          this.editPanel?.close()
```

Would it work?

davidbgk (approved):

umap/static/umap/js/umap.js
yohanboniface: Sounds like it is! Is it valid in the browser we support ?

davidbgk: [I think so](https://caniuse.com/mdn-javascript_operators_optional_chaining)

PR #1762, opened 2024-04-17
chore: allow to set playwright timeout from env var

yohanboniface:
so we can increase it in the CI


.github/workflows/test-docs.yml
```diff
@@ -46,6 +46,7 @@ jobs:
         env:
           DJANGO_SETTINGS_MODULE: 'umap.tests.settings'
           UMAP_SETTINGS: 'umap/tests/settings.py'
+          PLAYWRIGHT_TIMEOUT: '10000'
```
davidbgk, line 49: I think we can bump it safely there, it's so annoying when there is a failure for that…

```suggestion
          PLAYWRIGHT_TIMEOUT: '20000'
```

davidbgk (approved):

PR #1765, opened 2024-04-18
Mock OSM tiles requests in Playwright for speed

davidbgk:
Also, cast the PLAYWRIGHT_TIMEOUT parameter.


davidbgk:
Well, it does fail faster 😅

yohanboniface (approved):

PR #1774, opened 2024-04-25
feat: add browser tabs submenu

yohanboniface:


yohanboniface:
@Aurelie-Jallut I'm working on adding the "tabs style" menu in the browser/facets/about panels.

My first try is to factorize the code with the tabs we made for the icons, which looks like this:

image

So here is the result for the browser:

Screenshot from 2024-04-25 10-34-21

When no filter/facet is active, it looks like this:

image

Thoughts ? :)

Aurelie-Jallut:
something like that : 
image
 
or like this : 
image

Aurelie-Jallut:
image

yohanboniface:
image

Aurelie-Jallut:
@yohanboniface few feedbacks in french

pfbreton-perso:
Wow, this is a huge improvement!  thanks for making this happen!

PR #1776, opened 2024-04-25
feat: very minimal experimental conditional style rules

yohanboniface:
image

### DIFF
--- a/umap/static/umap/js/modules/global.js
+++ b/umap/static/umap/js/modules/global.js
@@ -5,6 +5,7 @@ import Caption from './caption.js'
 import { Panel, EditPanel, FullPanel } from './ui/panel.js'
 import Dialog from './ui/dialog.js'
 import Tooltip from './ui/tooltip.js'
+import Rules from './rules.js'
 import * as Utils from './utils.js'
 import { SCHEMA } from './schema.js'
 import { Request, ServerRequest, RequestError, HTTPError, NOKError } from './request.js'
@@ -43,6 +44,7 @@ window.U = {
   Panel,
   Request,
   RequestError,
+  Rules,
   SCHEMA,
   ServerRequest,
   SyncEngine,
--- a/umap/static/umap/js/modules/rules.js
+++ b/umap/static/umap/js/modules/rules.js
@@ -5,9 +5,7 @@ import { translate } from './i18n.js'
 class Rule {
   constructor(map, condition = '', options = {}) {
     this.map = map
-    this.condition = condition
     this.active = true
-    this.parse()
     this.options = options
     let isDirty = false
     Object.defineProperty(this, 'isDirty', {
@@ -21,6 +19,17 @@ class Rule {
         }
       },
     })
+    let _condition
+    Object.defineProperty(this, 'condition', {
+      get: () => {
+        return _condition
+      },
+      set: (value) => {
+        _condition = value
+        this.parse()
+      },
+    })
+    this.condition = condition
   }
 
   render(fields) {
--- a/umap/static/umap/js/modules/rules.js
+++ b/umap/static/umap/js/modules/rules.js
@@ -3,7 +3,6 @@ import * as Utils from './utils.js'
 import { translate } from './i18n.js'
 
 class Rule {
-  _condition = null
 
   get condition() {
     return this._condition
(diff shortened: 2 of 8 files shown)


yohanboniface:
@Aurelie-Jallut some screens:

image

image

umap/static/umap/js/modules/global.js
```diff
@@ -1,6 +1,7 @@
 import URLs from './urls.js'
 import Browser from './browser.js'
 import Facets from './facets.js'
+import Rules from './rules.js'
```
almet, line 4: Just naming this "rules" might be too broad. Maybe "filtering_rules" ?

umap/static/umap/js/modules/rules.js
```diff
@@ -0,0 +1,240 @@
...
+      },
+      set: (status) => {
+        isDirty = status
+        if (status) {
+          this.map.isDirty = status
+        }
+      },
+    })
+    let _condition
+    Object.defineProperty(this, 'condition', {
```
almet, line 23: Rather than using `Object.defineProperty` you could directly use a [get property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get), which could make the code more readable.

Any reason not to?

umap/static/umap/js/modules/rules.js
```diff
@@ -0,0 +1,240 @@
+import { DomUtil, DomEvent, stamp } from '../../vendors/leaflet/leaflet-src.esm.js'
+import * as Utils from './utils.js'
+import { translate } from './i18n.js'
+
+class Rule {
+  constructor(map, condition = '', options = {}) {
+    this.map = map
```
almet, line 7: If these are private, should we prefix them with `#`?

umap/static/umap/js/umap.js
```diff
@@ -786,7 +787,15 @@ U.Map = L.Map.extend({
     return U.SCHEMA[option] && U.SCHEMA[option].default
   },
 
-  getOption: function (option) {
+  getRuleOption: function (option, feature) {
```
almet, line 790: Can we directly use `this.rules` in the following calls?

umap/static/umap/js/modules/rules.js
almet, line 7: I'm wondering if it would it be possible to not pass the `map` object here? You would need to find ways to handle `isDirty` and the `render` call in a different way, maybe with events?

umap/static/umap/js/modules/rules.js
```diff
@@ -0,0 +1,240 @@
+import { DomUtil, DomEvent, stamp } from '../../vendors/leaflet/leaflet-src.esm.js'
+import * as Utils from './utils.js'
+import { translate } from './i18n.js'
+
+class Rule {
```
almet, line 5: It could be useful to have a comment explaining the API for these two classes.

almet (commented):
Here are just a few comments while reading the code :-) Cool, we'll have filtering rules soon :-)

umap/static/umap/js/modules/rules.js
yohanboniface: Excellent! I've use it :)

umap/static/umap/js/modules/global.js
yohanboniface: I'd keep it simple, until we have another kind of rules (just like we have browser and not "data browser" for eg.).

umap/static/umap/js/modules/rules.js
yohanboniface: Sadly, the browser support is not enough for us for this feature yet :(

umap/static/umap/js/umap.js
yohanboniface: :+1:

umap/static/umap/js/modules/rules.js
yohanboniface: This is a pattern we already have elsewhere in uMap, including in modules. Also I remember removing events call from the `requests` module (done by inheriting Leaflet `Class`), so I'm not exactly sure what is the pattern we'd encourage.

Anyway, I'm ok to challenge this pattern, but I'd prefer to do it in a separate PR and not to charge this one.

almet: The goal would be to move away from having this global `map` superobject. I'm okay to follow the discussion elsewhere.

PR #1782, opened 2024-04-30
fix: ensure tilelayer attribution with smart text is displayed as HMTL

yohanboniface:
We have "smart text" attribution in the DB (which translates to raw html), so I guess we're stuck with innerHTML…


davidbgk (commented):
I would rather have an explicit `L.DomUtil.addHTML()` for special cases we allow.

PR #1785, opened 2024-05-01
fix: deal with the modifier key under macOS

davidbgk:


umap/static/umap/js/umap.controls.js
```diff
@@ -392,7 +392,7 @@ U.EditControl = L.Control.extend({
       'mouseover',
       function () {
         map.ui.tooltip({
-          content: `${L._('Switch to edit mode')} (<kbd>Ctrl+E</kbd>)`,
+          content: map.help.displayLabel('TOGGLE_EDIT'),
```
davidbgk, line 395: ⚠️  This is not exactly the exact same message here, it was `Switch to edit mode` and now `Toggle edit mode`.

PR #1794, opened 2024-05-03
feat: integrate facets into browser filters

yohanboniface:
So now browser has three modes:

Layers (everything is collapsed, but user can show/hide/zoom/edit… layers):

image

Data (filters are collapsed, but layers are expanded):

image

And filters (everything is expanded):

image


yohanboniface:
> how about something like this to differentiate the filter button?

Sounds good! How it like when open ?

Aurelie-Jallut:
image

image

PR #1801, opened 2024-05-06
fix: allow `dir` and `title` attributes

### LINKED ISSUE #1796: Allow html dir attribute
I use dir="auto" in pop-ups so direction of text is based on line content. This is especially useful when someone uses browser translate function to machine-translate content. When text is translated to a rtl or ltr language its direction is fixed automatically.
On previous versions of umap it was possible to use dir attribute. With new updates seems it is disabled.

Please allow using dir attribute in html in pop-ups.

example pop-up that has dir="auto" in its template: [link](https://umap.openstreetmap.fr/en/map/map_943303?feature=%D8%AF%DA%A9%D8%A7%D9%86%20%D9%86%D8%A7%D9%86%D9%88%D8%A7%20%D8%A8%D8%A7%D8%B4%DB%8C#13/35.7889/51.4869)

davidbgk:
Fix #1796

### DIFF
--- a/umap/static/umap/js/modules/utils.js
+++ b/umap/static/umap/js/modules/utils.js
@@ -88,7 +88,7 @@ export function escapeHTML(s) {
       'span',
     ],
     ADD_ATTR: ['target', 'allow', 'allowfullscreen', 'frameborder', 'scrolling'],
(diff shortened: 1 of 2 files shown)


almet (approved):
Sounds good to me :-)

PR #1802, opened 2024-05-07
Expanded docker.md documentation

CampaniaGuy:
Hi,
after having initial issues with `umap.conf` not being found while using docker container i think it could be helpful to add a few more lines in the `docker.md` for a better start off.
- Added a comment below the `umap.conf` path definition asserting it is an optional setting
- Specified how to create an admin user and how to access it, otherwise you will experience problems in saving maps or adding tile-layers

### DIFF
--- a/docs/config/icons.md
+++ b/docs/config/icons.md
@@ -22,7 +22,7 @@ You can import icons manually by going to your uMap admin page: `https://your.se
 
 ## Import icons automatically
 
-To import icons on your uMap server, you will need to use the command `umap import_pictograms` from inside main project folder.
+To import icons on your uMap server, you will need to use the command `umap import_pictograms` (making sure the virtualenv is active).
 
 Note: you can get help with `umap import_pictograms -h`
 
--- a/docs/config/icons.md
+++ b/docs/config/icons.md
@@ -22,7 +22,7 @@ You can import icons manually by going to your uMap admin page: `https://your.se
 
 ## Import icons automatically
 
-To import icons on your uMap server, you will need to use the command `umap import_pictograms`.
+To import icons on your uMap server, you will need to use the command `umap import_pictograms` from inside main project folder.
 
 Note: you can get help with `umap import_pictograms -h`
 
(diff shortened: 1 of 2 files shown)


docs/deploy/docker.md
```diff
@@ -42,3 +43,5 @@ Note that you’ll have to set a [`SECRET_KEY`](https://docs.djangoproject.com/e
 ```sh
 $ python3 -c 'import secrets; print(secrets.token_hex(100))'
 ```
+
+User accounts can be managed via the Django admin page ({SITE_URL}/admin). The required superuser must be created on the container command line with this command: umap createsuperuser.
```
almet, line 47: For commands, it's better to put them on a new line, like this:

````markdown
```bash
umap createsuperuser
```
````

File: docs/deploy/docker.md [hunk 2]
```diff
@@ -27,6 +27,7 @@ services:
     volumes:
       - umap_userdata:/srv/umap/uploads
       # FIX the path on the left, below, to your location 
+      # OPTIONAL, you can comment the line below out
```
almet, line 30: Maybe explain why it's optional, and what would happen if you don't specify it (default values will be used, and then ENV variables will be read)

almet:
Thanks for the proposed changes.
I've added a few comments, hope they make sense.

docs/deploy/docker.md
CampaniaGuy: tried to add more context while keeping the line short. My personal problem with that setting was that while using docker there was an error trying to read the file, if file did not exist then a folder called `umap.conf` was created. Removing the setting was enough for me to make the container run flawlessly

almet (approved):

docs/config/icons.md
```diff
@@ -22,7 +22,7 @@ You can import icons manually by going to your uMap admin page: `https://your.se
 
 ## Import icons automatically
 
-To import icons on your uMap server, you will need to use the command `umap import_pictograms`.
+To import icons on your uMap server, you will need to use the command `umap import_pictograms` from inside main project folder.
```
yohanboniface, line 25:
```suggestion
To import icons on your uMap server, you will need to use the command `umap import_pictograms` (making sure the virtualenv is active).
```

PR #1804, opened 2024-05-09
fix: open edit panel when pressing ESC while drawing, if valid

yohanboniface:


umap/static/umap/js/umap.controls.js
```diff
@@ -1329,6 +1329,7 @@ U.Editable = L.Editable.extend({
       // (eg. line has only one drawn point)
       // So let's check if the layer has no more shape
       if (!e.layer.hasGeom()) e.layer.del()
+      else (e.layer.edit())
```
davidbgk, line 1332: Weird parenthesis syntax to avoid beautiful moustaches 😛

yohanboniface: But, don't we have a deal: if prettier keeps it on one line, I can avoid hipster hair ? :p

yohanboniface: oh, now I see the parenthesis around the line, weird!

PR #1809, opened 2024-05-10
chore: better browser labels

yohanboniface:


davidbgk (approved):
I wonder if we should call it `Panel` 🤔

PR #1822, opened 2024-05-14
feat: use accent color for switch and multichoices on dark background

yohanboniface:
Before:
image

After:
image

cc @Aurelie-Jallut cf discussion on Mattermost, but so you can have an other look at it

### DIFF
--- a/umap/static/umap/base.css
+++ b/umap/static/umap/base.css
@@ -449,9 +449,10 @@ input.switch:checked ~ label:before {
     text-indent: 0.7em;
     text-align: left;
     font-weight: bold;
+}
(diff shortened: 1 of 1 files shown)


davidbgk (approved):

Aurelie-Jallut:
I'm not sure about the constrast of the button. Have you tested the accessibility of the contrast? What is the code for the gray color of the text?

yohanboniface:
> Have you tested the accessibility of the contrast? 

Now yes! :)

Sounds good:

image


> What is the code for the gray color of the text?

#353c3c

PR #1827, opened 2024-05-16
feat: make popup panel expanded by default and add a small panel option

yohanboniface:
cf https://forum.openstreetmap.fr/t/umap-2-3-changement-de-gestion-des-pop-up/23680


yohanboniface:
Replaced by #1829

PR #1832, opened 2024-05-17
fix: do not close blindly editPanel when deleting a layer

### LINKED ISSUE #498: [enhancement] Don't always autoclose sidepanel after layer deletion
In "Manage layer options", you can browse layers and delete them by clicking the trash icon.  If you delete layers this way, the sidepanel automatically closes.

There is no need to close the sidepanel after this action. The user may want to interact with other layers right away (edit, delete, reorder, hide, etc.)... as he is currently using the layer manager.

This isn't the same case that editing a specific layer and deciding to delete it with "Advanced options > Delete". In the latter situation, closing the sidepanel right away makes more sense.

### LINKED ISSUE #1831: in editing : the layers panel automatically closes
In editing, when I perform an action in the Layers panel, the panel automatically closes, even though I want to perform several actions.

I'd like the panel to stay open and close either when I click on another toolbar button, or when I close the panel.

yohanboniface:
But remove from layers list, and make sure browser and caption are up to date.

fix #498
fix #1831

### DIFF
--- a/umap/static/umap/js/modules/caption.js
+++ b/umap/static/umap/js/modules/caption.js
@@ -7,6 +7,15 @@ export default class Caption {
     this.map = map
   }
 
+  isOpen() {
+    return Boolean(document.querySelector('.on .umap-caption'))
+  }
+
+  refresh() {
+    if (!this.isOpen()) return
+    this.open()
+  }
+
   open() {
     const container = DomUtil.create('div', 'umap-caption')
     DomUtil.createTitle(container, this.map.options.name, 'icon-caption')
--- a/umap/static/umap/js/umap.controls.js
+++ b/umap/static/umap/js/umap.controls.js
@@ -607,7 +607,6 @@ U.DataLayer.include({
           if (!this.isVisible()) return
           if (!confirm(L._('Are you sure you want to delete this layer?'))) return
           this._delete()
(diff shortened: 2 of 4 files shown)


umap/static/umap/js/modules/caption.js
```diff
@@ -7,6 +7,15 @@ export default class Caption {
...
+  isOpen() {
+    return !!document.querySelector('.on .umap-caption')
+  }
+
+  redraw() {
+    if (!this.isOpen()) return
+    this.open()
+  }
+
   open() {
```
davidbgk, line 19: Would it make more sense to call it `draw()`?

yohanboniface: You mean because of the `redraw` method ? In this case I'd better change this method name in `reopen` ?

umap/static/umap/js/modules/caption.js
```diff
@@ -7,6 +7,15 @@ export default class Caption {
     this.map = map
   }
 
+  isOpen() {
+    return !!document.querySelector('.on .umap-caption')
```
davidbgk, line 11:
```suggestion
    return document.body.contains('.on .umap-caption')
```

davidbgk (approved):

PR #1834, opened 2024-05-17
fix: do not consider shortcuts if shift is also pressed

### LINKED ISSUE #1452: On mac, ommand+option+i is captured and prevents from opening the dev console.
**Describe the bug**

When using mac OS, doing command + option + i opens the import panel rather than opening the dev tools. I'm not sure that's intentional (?)

 - OS: OS X
 - Browser Firefox 
 - Version 120.0.1

yohanboniface:
fix #1452

### DIFF
--- a/umap/static/umap/js/umap.js
+++ b/umap/static/umap/js/umap.js
@@ -518,11 +518,11 @@ U.Map = L.Map.extend({
 
   initShortcuts: function () {
     const globalShortcuts = function (e) {
-      const key = e.keyCode,
-        modifierKey = e.ctrlKey || e.metaKey
+      const key = e.keyCode
+      const hasModifier = (e.ctrlKey || e.metaKey) && !e.shiftKey
 
       /* Generic shortcuts */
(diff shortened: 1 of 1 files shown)


umap/static/umap/js/umap.js
```diff
@@ -518,11 +518,11 @@ U.Map = L.Map.extend({
 
   initShortcuts: function () {
     const globalShortcuts = function (e) {
-      const key = e.keyCode,
-        modifierKey = e.ctrlKey || e.metaKey
+      const key = e.keyCode
+      const hasModifier = (e.ctrlKey || e.metaKey) && !e.shiftKey
```
davidbgk, line 522: What about shortcuts [with a shift key](https://github.com/umap-project/umap/blob/master/umap/static/umap/js/umap.core.js#L326-L329)?

yohanboniface: Unless I'm blind, which I'm close to, those are not managed here (in the global shortcuts), they are listened on click on some elements, isn't it ?

PR #1846, opened 2024-05-21
chore: move autocomplete to modules/

yohanboniface:


umap/static/umap/js/modules/autocomplete.js
```diff
@@ -0,0 +1,317 @@
+import { DomUtil, DomEvent, setOptions } from '../../vendors/leaflet/leaflet-src.esm.js'
+import { translate } from './i18n.js'
+import { ServerRequest } from './request.js'
+
+export class BaseAutocomplete {
+  constructor(el, options) {
+    this.el = el
+    this.options = {
+      placeholder: translate('Start typing...'),
+      emptyMessage: 'No result',
```
davidbgk, line 10: No translate for that one?

umap/static/umap/js/modules/autocomplete.js
```diff
@@ -0,0 +1,317 @@
...
+  }
+
+  hide() {
+    this.clear()
+    this.container.style.display = 'none'
+    this.input.value = ''
+  }
+
+  setChoice(choice) {
+    console.log(choice)
```
davidbgk, line 145: Oups :)

umap/static/umap/js/modules/autocomplete.js
```diff
@@ -0,0 +1,317 @@
...
+      this.current = result
+      this.highlight()
+    })
+    DomEvent.on(el, 'mousedown', () => this.setChoice())
+    return result
+  }
+
+  resultToIndex(result) {
+    let out = null
+    this.results.forEach((item, index) => {
```
davidbgk, line 178: Using `this.results.filter()` is probably more appropriate to retrieve an element from the list?

umap/static/umap/js/modules/autocomplete.js
```diff
@@ -0,0 +1,317 @@
...
+    }
+  }
+
+  async search() {
+    let val = this.input.value
+    if (val.length < this.options.minChar) {
+      this.clear()
+      return
+    }
+    if (`${val}` === `${this.cache}`) return
```
davidbgk, line 247: Weird comparison, why don't we compare directly the variables?

umap/static/umap/js/modules/autocomplete.js
yohanboniface: `filters` will return an array, but `findIndex` seems to do exactly what I need, thanks for the hint!

umap/static/umap/js/modules/autocomplete.js
yohanboniface: No idea, I removed. I'm not totally why this check if for, actually. So let's fix and see :p

davidbgk (approved):

PR #1847, opened 2024-05-22
chore: move ui to dedicated modules

yohanboniface:
`umap.ui.js` is removed, and replaced by those two modules: `tooltip.js` and `alert.js`
Also, panel.js is moved in the same folder `modules/ui/`
Also, a minimal Dialog class is added, to replace the custom made help box.

Those changes impact a lot of lines, but with very little logic change.
The most notable is certainly that `Request` does not expect anymore a `ui` parameter but an `alert` directly. And thus, it cannot use `ui` anymore to send its event, so it sends them on the body element, which the Map is then listening.

Better to review each commit separately. :)


yohanboniface:
Oops, that PR was totally out of my radar :(

davidbgk (approved):

PR #1857, opened 2024-05-23
Importers for custom external sources

### LINKED ISSUE #112: Ease OSM objects integration through Overpass with a dedicated simple UI
Originally reported by: **BitBucket: [clkao](http://bitbucket.org/clkao), GitHub: [clkao](http://github.com/clkao)**

---

It'd be great if we can simply add way/node/relation from osm as data entries in a layer, and apply styles like current layer nodes.

This allows reusing POI/objects, tracking changes from OSM, and encourages people to edit things on OSM first and simply reference them in umap.

---
- Bitbucket: https://bitbucket.org/yohanboniface/umap/issue/112

### LINKED ISSUE #244: [wishlist] When importing data into a new layer, allow to name it before clicking "Import"
When importing data (Ctrl+I), it could be nice if a "layer's name" field could appear after selecting "Import in a new layer".

My guess is that renaming a layer is the next thing one usually does after creating it.
When importing several files, this would avoid to repeatability: move the mouse to the layer tab, find the layer with the file name, click edit layer + select name's field and rename.

And if one doesn't care about the layer's name, the field would show by default what's currently done: the file's name.

yohanboniface:
fix #112 
fix #244 in the process


umap/static/umap/content.css
```diff
@@ -248,7 +248,7 @@ input[type="submit"],
 ul.umap-autocomplete {
     position: absolute;
     background-color: white;
-    z-index: 1010;
+    z-index: 10100;
```
davidbgk, line 251: Let's talk about a strategy here 😅 

```css
:root {
  --level-base: auto;
  --level-panels: 10;
  --level-autocomplete: 20;
  --level-errors: 30;
  and so on
}
```
and when we need more:

```css
.autocomplete-input {
  z-index: calc(var(--level-autocomplete) + 1);
}
```

These are dumb examples but you hopefully see what I mean.

davidbgk (commented):
Following our recent discussions, I am wondering if we want to store these autocomplete values within the URL to be able to link to a map with one-click territory creation 🤔

umap/static/umap/js/modules/importer.js
```diff
@@ -1,147 +1,165 @@
 import { DomUtil, DomEvent } from '../../vendors/leaflet/leaflet-src.esm.js'
 import { translate } from './i18n.js'
+import Dialog from './ui/dialog.js'
+
+const AVAILABLE_IMPORTERS = ['geodatamine', 'communesfr', 'presets']
```
davidbgk, line 5: I don't get why we need this instead of using `UMAP_EXPERIMENTAL_IMPORTERS`?

yohanboniface: Good point: I've done this as we have a dynamic import so I wanted to make sure we would only load known path. Does that make sense or I'm overthinking that thing ?

davidbgk: I can't see how it can be dangerous with the current implementation.

yohanboniface:
> Following our recent discussions, I am wondering if we want to store these autocomplete values within the URL to be able to link to a map with one-click territory creation

I think this is another reflection and another PR, but yes, having those importers may ease it a lot!

umap/static/umap/content.css
yohanboniface: Let's do that in a dedicated PR! :)

PR #1865, opened 2024-05-29
fix: refactor new line management in popups

### LINKED ISSUE #1854: Caption popup: too much spacing when using the # markup.
**Describe the bug**
There is too much spacing in the caption popup when using the # markup.

**To Reproduce**
1. Go to https://umap.openstreetmap.fr/en/map/reseau-de-benso-jamanu-en-2020_701348
2. Click on any data point
3. in the caption popup, notice the extra spacing below the popup heading

**Expected behavior**
More proportional spacing expected as shown in preferred version at bottom

**Screenshots**
image

 - OS: Win11
 - Browser Edge
 - Version 125.0.2535.51 

**Additional context**
The popup underlying template is as such:
```
# {Type} de {Guichet}
Agence: **{Agence}**
Commune: **{Commune}**
Cercle: **{Cercle}**
Année de Création: **{Annee_Creation}**
Accessibilité: **{Accessibilite}**
```

It appears the code is using `<br>` elements to implement line breaks. However the # markup also calls for `<h3>`, hence the final code has `<h3><br>` which creates a double line break, hence the extra space. Also line height for `<h3>` is set to 3.

I have removed the extra `<br>` below `<h3>` and set line-height to 1.75 in my preferred version below

image

yohanboniface:
Basically, we do not add `<br>` anymore, but instead let the newline and use the css `white-space: pre-line;`. Also remove new lines after title.

While doing that change, I needed to reorder regex so the ones needing a newline before would still have it (bullets points).

I've added tests and removed some to cover those changes.

fix #1854

### DIFF
--- a/umap/static/umap/js/modules/utils.js
+++ b/umap/static/umap/js/modules/utils.js
@@ -112,26 +112,23 @@ export function escapeHTML(s) {
 export function toHTML(r, options) {
   if (!r) return ''
   const target = (options && options.target) || 'blank'
-  let ii
 
-  // detect newline format
-  const newline = r.indexOf('\r\n') != -1 ? '\r\n' : r.indexOf('\n') != -1 ? '\n' : ''
+  // unordered lists
+  r = r.replace(/^\*\* (.*)/gm, '<ul><ul><li>$1</li></ul></ul>')
+  r = r.replace(/^\* (.*)/gm, '<ul><li>$1</li></ul>')
+  for (let ii = 0; ii < 3; ii++) {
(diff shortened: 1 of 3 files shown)


davidbgk (approved):
Why don't we use a regular markdown parser? Newlines should generate HTML paragraphs to be consistent?

yohanboniface:
Moving to markdown is clearly an option at some point, but at this stage our syntax is not compatible with Markdown, it's a Yoyodown thing :p

PR #1867, opened 2024-05-29
fix: show message in case of xml parsing error

### LINKED ISSUE #1866: Show error for ' not well-formed'
Currently if you try and import a kml that isn't a kml file (kmz in my case) and select kml in the file format field.
You'll get a ' not well-formed' error in devtools. which is fair. But the problem is that to the enduser it doesn't show any actual error in the UI for this. For the end user it seems like nothing happens

yohanboniface:
fix #1866


davidbgk (approved):
I guess I'll have to rebase my alert refactor branch :)

PR #1872, opened 2024-06-03
Update fr.json

lmagreault:
Minor correction


yohanboniface:
Pulled from Transifex, thanks!

PR #1876, opened 2024-06-03
Use web components to display alerts + a11y roles

davidbgk:


docs-users/static/tutoriels/create-map-alert.png
davidbgk, line 1: The target goal from @Aurelie-Jallut :

image(1)

umap/static/umap/js/components/alerts/alert.js
```diff
@@ -0,0 +1,177 @@
...
+        this._hide()
+      }
+    })
+    document.addEventListener('umap:alert-creation', (event) => {
+      this._displayCreationAlert(event.detail)
+    })
+  }
+}
+
+class uMapAlertChoice extends uMapAlert {
```
yohanboniface, line 123: I'd name it `uMapAlertConflict`, "choice" seems to generic to me, while it seems to only deals with the 422 case

umap/static/umap/js/components/alerts/alert.js
```diff
@@ -0,0 +1,177 @@
+class uMapAlert extends HTMLElement {
+  static info(message, duration = 5000) {
```
yohanboniface, line 2: I thought we find a way without custom events ?

umap/static/umap/js/components/alerts/alert.js
```diff
@@ -0,0 +1,177 @@
...
+    this._show()
+    if (Number.isFinite(duration)) {
+      setTimeout(() => {
+        this._hide()
+      }, duration)
+    }
+  }
+
+  connectedCallback() {
+    this.addEventListener('click', (event) => {
```
yohanboniface, line 49: can't we call super instead of duplicating this line for each class ?

umap/static/umap/js/components/alerts/alert.js
```diff
@@ -0,0 +1,177 @@
...
+      L.Util.copyToClipboard(editLink)
+      event.target.value = L._('✅ Copied!')
+    })
+    if (sendLink) {
+      this.formWrapper.removeAttribute('hidden')
+      const form = this.formWrapper.querySelector('form')
+      form.addEventListener('submit', async (event) => {
+        event.preventDefault()
+        const formData = new FormData(form)
+        const server = new U.ServerRequest()
```
yohanboniface, line 104: I guess we could import `ServerRequests` directly ?

umap/static/umap/js/modules/importer.js
```diff
@@ -163,16 +163,12 @@ export default class Importer {
         this.map.processFileToImport(file, layer, type)
       }
     } else {
-      if (!type)
-        return this.map.alert.open({
-          content: translate('Please choose a format'),
-          level: 'error',
-        })
+      if (!type) return U.Alert.error(L._('Please choose a format'))
```
yohanboniface, line 166: Import `Alert` directly (and same below) ?

umap/static/umap/js/components/alerts/alert.js
```diff
@@ -0,0 +1,177 @@
...
+  constructor() {
+    super()
+    this.choiceWrapper = this.container.querySelector('#choice-wrapper')
+  }
+
+  _displayChoiceAlert(detail) {
+    const { level = 'info', duration = 5000, message = '' } = detail
+    uMapAlert.prototype._displayAlert.call(this, { level, duration, message })
+    const form = this.choiceWrapper.querySelector('form')
+    form.addEventListener('submit', (event) => {
```
yohanboniface, line 146: Can't we just have have two buttons with their listeners instead of a form we loop on the `submitter.id` ? That would look more readable to me.

umap/static/umap/js/components/alerts/alert.html
```diff
@@ -0,0 +1,69 @@
+{% load i18n static %}
+
+<style type="text/css">
+  @import "{% static 'umap/js/components/alerts/alert.css' %}";
+</style>
+
+<umap-alert hidden>
```
yohanboniface, line 7: No more templates ?

yohanboniface (commented):
:+1:

umap/static/umap/js/components/alerts/alert.js
davidbgk: Yes, we need these but we can make it less verbose when creating events.

umap/static/umap/js/components/alerts/alert.js
davidbgk: Yes, let's split it as a dedicated method.

umap/static/umap/js/components/alerts/alert.js
davidbgk: Let's see if there are circular imports in JS :)

umap/static/umap/js/components/alerts/alert.js
```diff
@@ -0,0 +1,158 @@
...
+      editLink = undefined,
+      sendLink = undefined,
+    } = event.detail
+    uMapAlert.prototype.onAlert.call(this, { detail: { level, duration, message } })
+    this.linkWrapper.querySelector('input[type="url"]').value = editLink
+    const button = this.linkWrapper.querySelector('input[type="button"]')
+    button.addEventListener('click', (event) => {
+      event.preventDefault()
+      L.Util.copyToClipboard(editLink)
+      event.target.value = L._('✅ Copied!')
```
davidbgk, line 99: Replace with `translate()`

PR #1881, opened 2024-06-05
chore: move help to a separate module

yohanboniface:
Keeping the API as is for now!
Small step, but preparing for the next :)


davidbgk (approved):
👍

PR #1900, opened 2024-06-14
Multiple fixes for importers

davidbgk:


umap/static/umap/base.css
```diff
@@ -203,6 +203,7 @@ input[type=checkbox]:checked:after {
 label input[type="radio"] {
     appearance: none;
     margin-right: 10px;
+    background: var(--color-darkGray);
 }
 input[type="radio"]:after {
     display: inline-block;
```
davidbgk, line 209: We might want to switch to [Circled white bullet, white point bullet](https://altcodeunicode.com/alt-codes-bullet-point-symbols/) for consistency when the radio input is not selected.

yohanboniface (approved):

PR #1916, opened 2024-06-18
fix: display an alert if the import is successful

### LINKED ISSUE #1913: Getting feedback when I click on the Import data button
I don't know what happens when I click on the Import data button. I'm not sure if the import was done correctly or if there was a problem.

Feedback confirming that the import was successful or informing me if there was a problem.

davidbgk:
Fix #1913

### DIFF
--- a/umap/static/umap/js/modules/importer.js
+++ b/umap/static/umap/js/modules/importer.js
@@ -234,9 +234,17 @@ export default class Importer {
   }
 
   submit() {
-    if (this.format === 'umap') this.full()
-    else if (!this.url) this.copy()
-    else if (this.action) this[this.action]()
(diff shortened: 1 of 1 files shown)


umap/static/umap/js/modules/importer.js
```diff
@@ -234,9 +234,17 @@ export default class Importer {
   }
 
   submit() {
-    if (this.format === 'umap') this.full()
-    else if (!this.url) this.copy()
-    else if (this.action) this[this.action]()
+    let hasErrors = false
```
davidbgk, line 237: I'm not fully convinced by that approach but I couldn't find a clever way to only display the successful message in case there is no error during the potential imports.

PR #1918, opened 2024-06-19
docs: Major version in case of server change(s)

davidbgk:


docs/release.md
```diff
@@ -58,6 +58,7 @@ We aim to support [Baseline](https://developer.mozilla.org/en-US/blog/baseline-e
 
 * when we bump Django to a major version
 * when we change how we store data (both in database and filesystem)
+* when we make required changes on the server (for people maintaining instances)
```
yohanboniface, line 61: Sounds a bit broad to me, not every change on the server should require a major (eg. updating the settings is a server change). Something like `when we change how uMap can be deployed` ?

PR #1935, opened 2024-06-24
fix: better spacing for the alert box, when activated

almet:
Before:
Capture d’écran du 2024-06-24 11-04-49

After:
Capture d’écran du 2024-06-24 11-07-22

### DIFF
--- a/umap/static/umap/content.css
+++ b/umap/static/umap/content.css
@@ -169,7 +169,7 @@ h2.tabs a:hover {
     background-color: #c0392b;
     color: #efefef;
     margin-top: 0;
-    padding: 1em;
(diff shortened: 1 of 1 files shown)


umap/static/umap/content.css
```diff
@@ -169,7 +169,7 @@ h2.tabs a:hover {
     background-color: #c0392b;
     color: #efefef;
     margin-top: 0;
-    padding: 14px 0;
+    padding: 1em;
```
davidbgk, line 172: We try to use CSS vars now to uniformize gaps and sizes:

```suggestion
    padding: var(--gutter);
```

yohanboniface (approved):

Issue #1931, opened 2024-06-23
Polyline disapears on layer movement


isidroas:
**To Reproduce**
It is easy to replicate. Move a polyline from one layer to another and it will disappear and you will not able to find it either layer.

For example, you can test it here: https://umap.openstreetmap.fr/en/map/untitled-map_1086366#6/24.847/-72.092

 - OS: Ubuntu
 - Browser Firefox and Chromium

yohanboniface:
Thanks for the report!

Seems related to the order of  removing a feature from a layer.

Traceback:

![image](https://github.com/umap-project/umap/assets/146023/e12609c3-b6f5-462e-b0df-d2c852601ac6)

Quick fix:

```diff
diff --git a/umap/static/umap/js/umap.layer.js b/umap/static/umap/js/umap.layer.js
index 2ca9ee62..4ab4462c 100644
--- a/umap/static/umap/js/umap.layer.js
+++ b/umap/static/umap/js/umap.layer.js
@@ -913,10 +913,10 @@ U.DataLayer = L.Evented.extend({
 
   removeLayer: function (feature) {
     const id = L.stamp(feature)
+    this.layer.removeLayer(feature)
     feature.disconnectFromDataLayer(this)
     this._index.splice(this._index.indexOf(id), 1)
     delete this._layers[id]
-    this.layer.removeLayer(feature)
     delete this.map.features_index[feature.getSlug()]
     if (this.hasDataLoaded()) this.fire('datachanged')
   },
```

But needs further investigation, and a unit test.

@almet if you are onboard tomorrow, could you have a look (I'm off) ? Worth a patch release too I'd say.
Otherwise I'll try tomorrow night!

isidroas:
Thank you! 🙏

Just in time for my trip, now I can now move the features to a layer for each day!

PR #1936, opened 2024-06-24
Multiple sync-related fixes

almet:
I found some issues with the current implementation of the sync mechanism, which this PR fixes.

- The "move a feature from one datalayer to another" feature was not thought of, at all, and as a result introduced the bug that was referenced in #1931.
- It turns out that the syncing of this operation wasn't working either, creating shape clones along the way.
- The previous (buggy) implementation was not sending the proper layer id, resulting to features hosted by the wrong layers.
- Operations were triggered when loading data from the original server, which shouldn't happen.

Generally speaking, the current implementation misses some cross-datalayer tests, but I suggest these should come as a separate pull request.

### DIFF
--- a/umap/static/umap/js/umap.features.js
+++ b/umap/static/umap/js/umap.features.js
@@ -23,7 +23,7 @@ U.FeatureMixin = {
     // event triggered to cause a sync event, as it would reintroduce
     // deleted features.
     // The `._marked_for_deletion` private property is here to track this status.
-    if (this._marked_for_deletion == true) {
+    if (this._marked_for_deletion === true) {
       this._marked_for_deletion = false
       return
     }
@@ -37,7 +37,7 @@ U.FeatureMixin = {
   initialize: function (map, latlng, options, id) {
     this.map = map
     this.sync = map.sync_engine.proxy(this)
-    this._mark_for_deletion = false
+    this._marked_for_deletion = false
 
     if (typeof options === 'undefined') {
       options = {}
--- a/umap/static/umap/js/umap.features.js
+++ b/umap/static/umap/js/umap.features.js
@@ -6,7 +6,7 @@ U.FeatureMixin = {
       subject: 'feature',
       metadata: {
         id: this.id,
-        layerId: this.datalayer?.id || null,
+        layerId: this.datalayer?.umap_id || null,
         featureType: this.getClassName(),
       },
     }
--- a/umap/static/umap/js/umap.features.js
+++ b/umap/static/umap/js/umap.features.js
@@ -15,7 +15,18 @@ U.FeatureMixin = {
   onCommit: function () {
     // When the layer is a remote layer, we don't want to sync the creation of the
     // points via the websocket, as the other peers will get them themselves.
-    if (this.datalayer.isRemoteLayer()) return
+    if (this.datalayer?.isRemoteLayer()) return
+
+    // The "endEdit" event is triggered at the end of an edition,
+    // and will trigger the sync.
+    // In the case of a deletion (or a change of layer), we don't want this
+    // event triggered to cause a sync event, as it would reintroduce
+    // deleted features.
+    // The `._marked_for_deletion` private property is here to track this status.
+    if (this._marked_for_deletion == true) {
+      this._marked_for_deletion = false
+      return
+    }
     this.sync.upsert(this.toGeoJSON())
   },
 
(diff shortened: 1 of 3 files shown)


umap/static/umap/js/umap.features.js
```diff
@@ -15,21 +15,29 @@ U.FeatureMixin = {
   onCommit: function () {
     // When the layer is a remote layer, we don't want to sync the creation of the
     // points via the websocket, as the other peers will get them themselves.
-    if (this.datalayer.isRemoteLayer()) return
+    if (this.datalayer && this.datalayer.isRemoteLayer()) return
```
davidbgk, line 18:
```suggestion
    if (this.datalayer?.isRemoteLayer()) return
```

Does it make a difference I missed?

almet: Good hint :+1:

umap/static/umap/js/umap.features.js
```diff
@@ -15,21 +15,29 @@ U.FeatureMixin = {
...
+
+    // The "endEdit" event is triggered at the end of an edition,
+    // and will trigger the sync.
+    // In the case of a deletion (or a change of layer), we don't want this
+    // event triggered to cause a sync event, as it would reintroduce
+    // deleted features.
+    // The `._marked_for_deletion` private property is here to track this status.
+    if (this._marked_for_deletion == true) {
+      this._marked_for_deletion == false
```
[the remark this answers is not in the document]
almet: :upside_down_face: fixed, in [`70a55b6` (#1936)](https://github.com/umap-project/umap/pull/1936/commits/70a55b62d5708ca5f38e2e3dcf2bf0400fc48dbb)

umap/static/umap/js/umap.js
```diff
@@ -190,7 +190,7 @@ U.Map = L.Map.extend({
         dataUrl = decodeURIComponent(dataUrl)
         dataUrl = this.localizeUrl(dataUrl)
         dataUrl = this.proxyUrl(dataUrl)
-        const datalayer = this.createDataLayer()
+        const datalayer = this.createDataLayer({})
```
yohanboniface, line 193: Not sure why this is needed, it seems to me that this parameter already has a default

umap/static/umap/js/umap.layer.js
```diff
@@ -939,11 +939,11 @@ U.DataLayer = L.Evented.extend({
     if (idx !== -1) this._propertiesIndex.splice(idx, 1)
   },
 
-  addData: function (geojson) {
+  addData: function (geojson, sync) {
```
yohanboniface, line 942: Not saying to do it in this PR, but this rings a bell to make sync order out of the internal object process.
In very short, the one who calls `createDataLayer` would also create sync after, or not.

umap/static/umap/js/umap.features.js
```diff
@@ -15,21 +15,29 @@ U.FeatureMixin = {
...
   },
 
-  syncDelete: function () {
-    this.sync.delete()
-  },
-
   initialize: function (map, latlng, options, id) {
     this.map = map
     this.sync = map.sync_engine.proxy(this)
+    this._mark_for_deletion = true
```
yohanboniface, line 40: I don't get why this is initialized to `true`.

umap/static/umap/js/umap.layer.js
almet: Correct, that's something to solve at some point. 👍

umap/static/umap/js/umap.js
almet: You're correct. Fixed.

umap/static/umap/js/umap.features.js
almet: That's an error. Fixed.

almet:
Rebased on latest master, and incorporated the proposed changed from reviews. We should be ready to go.

umap/static/umap/js/umap.js
davidbgk: Is it back again?

umap/static/umap/js/umap.features.js
```diff
@@ -15,21 +15,29 @@ U.FeatureMixin = {
...
-    if (this.datalayer.isRemoteLayer()) return
+    if (this.datalayer?.isRemoteLayer()) return
+
+    // The "endEdit" event is triggered at the end of an edition,
+    // and will trigger the sync.
+    // In the case of a deletion (or a change of layer), we don't want this
+    // event triggered to cause a sync event, as it would reintroduce
+    // deleted features.
+    // The `._marked_for_deletion` private property is here to track this status.
+    if (this._marked_for_deletion == true) {
```
davidbgk, line 26:
```suggestion
    if (this._marked_for_deletion === true) {
```

Is it safer to check for [strict equality](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) here?

almet: yes it is :+1:

yohanboniface (approved):

PR #1942, opened 2024-06-25
chore: rationalize z-indexes in CSS

davidbgk:
Positive side-effect (fix): alerts are now above importers panel.

Before:

Capture d’écran, le 2024-06-25 à 14 25 47

After:

Capture d’écran, le 2024-06-25 à 14 50 17


yohanboniface (approved):
Excellent!

PR #1943, opened 2024-06-25
feat: handle multiple dataUrl parameters on map load

davidbgk:


yohanboniface (approved):
Sounds good, but given all other parameters are unique, in particular `dataFormat` (but `color`, `labelKey` or `popupShape` too), I feel it's a bit asymmetrical. But let's see how it will be used!

davidbgk:
I added users' documentation related to that topic:
Capture d’écran, le 2024-06-26 à 13 40 50

Capture d’écran, le 2024-06-26 à 13 41 03

yohanboniface:
Cool! Do we also mention the `data` parameter ?

davidbgk:
https://media3.giphy.com/media/rvaQRHCzisFeo/giphy.mp4#height=182&width=293&isvideo=true

PR #1944, opened 2024-06-25
chore: apply Biome unsafe changes

davidbgk:
Without `useBlockStatements`

### DIFF
--- a/umap/static/umap/js/modules/rules.js
+++ b/umap/static/umap/js/modules/rules.js
@@ -49,7 +49,7 @@ class Rule {
   }
 
   not_equal(other) {
-    return this.expected != other
+    return this.expected !== other
   }
 
   gt(other) {
@@ -71,10 +71,10 @@ class Rule {
         break
       }
     }
-    if (vars.length != 2) return
+    if (vars.length !== 2) return
     this.key = vars[0]
     this.expected = vars[1]
-    if (!isNaN(this.expected)) this.cast = Number.parseFloat
+    if (!Number.isNaN(this.expected)) this.cast = Number.parseFloat
     else if (['true', 'false'].includes(this.expected)) this.cast = (v) => !!v
     this.expected = this.cast(this.expected)
   }
@@ -161,7 +161,7 @@ class Rule {
   }
 
   _delete() {
-    this.map.rules.rules = this.map.rules.rules.filter((rule) => rule != this)
+    this.map.rules.rules = this.map.rules.rules.filter((rule) => rule !== this)
   }
 }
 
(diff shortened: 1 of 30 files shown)


umap/static/umap/js/modules/sync/updaters.js
```diff
@@ -89,9 +89,10 @@ export class FeatureUpdater extends BaseUpdater {
       console.error(`Unable to find feature with id = ${metadata.id}.`)
     }
     switch (key) {
-      case 'geometry':
+      case 'geometry': {
```
yohanboniface, line 92: Don't get this one

umap/static/umap/js/umap.js
```diff
@@ -1340,7 +1338,6 @@ U.Map = L.Map.extend({
         'options.overlay.url_template',
         {
           handler: 'BlurInput',
-          helpText: `${L._('Supported scheme')}: http://{s}.domain.com/{z}/{x}/{y}.png`,
           placeholder: 'url',
           helpText: L._('Background overlay url'),
```
yohanboniface, line 1342: I think this one should be `label` and the other one kept.

umap/static/umap/js/modules/rules.js
```diff
@@ -71,9 +71,10 @@ class Rule {
         break
       }
     }
-    if (vars.length != 2) return
+    if (vars.length !== 2) return
     this.key = vars[0]
     this.expected = vars[1]
+    // biome-ignore lint/suspicious/noGlobalIsNan: expected might not be a number.
```
yohanboniface, line 77: don't get this exception

yohanboniface: Ok, sorry, found the comment above, `Number.isNaN` excepts a number, that may be `NaN`, while `isNaN` returns false for every non number value.

yohanboniface (approved):

umap/static/umap/js/modules/sync/updaters.js
davidbgk: The error is that one: https://biomejs.dev/linter/rules/no-switch-declarations/ maybe we should use a `if` for that one-case switch?

davidbgk: Something like that:

```suggestion
    if (key === 'geometry') {
      const datalayer = this.getDataLayerFromID(metadata.layerId)
      this.getDataLayerFromID(metadata.layerId).geoJSONToLeaflet({
        geometry: value,
        id: metadata.id,
        feature,
      })
    } else {
      this.updateObjectValue(feature, key, value)
      feature.datalayer.indexProperties(feature)
    }

```

PR #1953, opened 2024-07-01
feat: add new type of layer Categorized

### LINKED ISSUE #1433: Add a Category type for layers
Sorry if this is being implemented, but I did not find it in the roadmap.

Currently, if you want to have categories in your uMap, you have to create a separate layer for each of them. It would be interesting to allow uMap to add this feature for qualitative data, now that it covers quantitative data with the choropleth feature.
Here is how it looks like in QGIS: https://docs.qgis.org/3.28/en/docs/training_manual/vector_classification/classification.html

yohanboniface:
This is like Choropleth, but with categories instead of graduated.

fix #1433

I'd tried to keep only one class (Choropleth), but given Choropleth needs numbers while Categorized use anything and Choropleth needs breaks (range of values) while Categorized only checks for equality, this was ending with more mess.

[Screencast from 2024-07-01 12-44-45.webm](https://github.com/umap-project/umap/assets/146023/6fc25fb2-ea80-4530-80bf-def6210feb2f)


yohanboniface:
cc @severinmenard @ecologeek

davidbgk:
The LEGEND! 🤩

PR #1956, opened 2024-07-01
chore: add custom prompt

yohanboniface:
This also removes the dependency from dialog to map and vice versa.


davidbgk (commented):
I think we should use a Dialog here, but with a support for no-dialog browser given it's baseline2022.

A good inspiration: https://css-tricks.com/replace-javascript-dialogs-html-dialog-element/

davidbgk (approved):

PR #1958, opened 2024-07-01
fix: main help button from edit toolbar was broken

yohanboniface:


umap/static/umap/js/modules/help.js
```diff
@@ -191,7 +191,7 @@ export default class Help {
     const container = DomUtil.add('div')
     DomUtil.createTitle(container, translate('Help'))
     // Special dynamic case. Do we still think this dialog is usefull ?
-    if (entries === 'edit') {
+    if (entries[0] === 'edit') {
```
davidbgk, line 194: That exception looks smelly, can't we have a dedicated `showHelp()`?

davidbgk (approved):

umap/static/umap/js/modules/help.js
yohanboniface: Done! :)

umap/static/umap/js/modules/help.js
```diff
@@ -190,42 +190,48 @@ export default class Help {
...
+    })
     return button
   }
 
-  link(container, entries) {
-    const button = this.button(container, entries, 'umap-help-link')
+  getStartedLink(container) {
+    const button = DomUtil.createButton('umap-help-link', container, translate('Help'))
     button.textContent = translate('Help')
+    DomEvent.on(button, 'click', DomEvent.stop).on(button, 'click', () => {
```
davidbgk, line 232: What are we stopping prior to show the element? Would it be easier to use regular JS event handling?

umap/static/umap/js/modules/help.js
```diff
@@ -190,42 +190,48 @@ export default class Help {
...
-          innerHTML: ENTRIES[name],
-        })
-      }
+        innerHTML: ENTRIES[name],
+      })
     }
     this.map.dialog.open({ content: container, className: 'dark' })
   }
 
+  // Special dynamic case. Do we still think this dialog is usefull ?
```
davidbgk, line 204:
```suggestion
  // Special dynamic case. Do we still think this dialog is useful?
```

Let's do that in another PR.

PR #1964, opened 2024-07-02
fix: add throttle in autocomplete

yohanboniface:


davidbgk (approved):
I wonder if it should be a Util function 🤔

yohanboniface:
> I wonder if it should be a Util function 🤔

Let's do that when we need it twice ?

PR #1979, opened 2024-07-09
fix: BlurInput was not updating the object in some situations

yohanboniface:
Steps to reproduce:
- click in any BlurInput
- change to any value
- click outside the input => the value will be set to the object
- click again in the input
- change value to the initial
- click outside => the value will not be set

This is because the BlurInput in sync on blur, and it was comparing the current field value with an "initial" one, only set at build. This check is made because we don't want to trigger an edit when a user click in a blurinput and click outside without changing the value.


davidbgk (approved):
🙌 👏

PR #1967, opened 2024-07-02
Add a way to delete rows from a layer by a condition

yohanboniface:
Replace #1957

PR #1981, opened 2024-07-09
feat: add AutocompleteDatalist and use it for conditional rules

yohanboniface:
Extracted from #1967


davidbgk (approved):
Let's add some tests for that one

yohanboniface:
Sadly, I didn't make it: playwright does not seem to allow us to expect what's in the autocomplete. We can click on it, but it will only record the final fill of the input :(

Or do you have another plan in mind ?

Issue #1870, opened 2024-05-31
Installation on Synology-NAS


Ball-Pic:
Does anyone have experience installing on a Synlogy NAS?
I just can't do it.
Can someone provide me with instructions?
Maybe even via portainer.

Thx Gerhard

geimist:
These instructions only serve as a start. The following pages serve as a basis and should be observed:
- https://docs.umap-project.org/en/master/deploy/docker/
- https://docs.umap-project.org/en/master/config/settings

## Preparation:
- create a key that will later be used as SECRET_KEY (= "a long and random secret key that must not be shared"
  You can easily generate one with openssl at command line with: `openssl rand -base64 32`)
- Under which web address should uMap be called? It must be assigned to the environment variable SITE_URL and specified in the reverse proxy. In this description https://map.example.com is used

<hr>   

### 1. Paths:   
Create the following folders in the FileStation (there may also be different paths, but these must also be adjusted in the configuration below):
`/volume1/docker/umap/POSTGRESQL_DATA`
`/volume1/docker/umap/STATIC_ROOT`
`/volume1/docker/umap/MEDIA_ROOT`
`/volume1/docker/umap/conf`


<hr>   

### 2. Config:
Create the configuration file and save it here (I recommend the DSM text editor from the package centre):
- /volume1/docker/umap/conf/umap.conf
<details>
  <summary>example config (click) - you still need to customise them:</summary>
  
```  
# ➜ https://docs.umap-project.org/en/master/config/settings
#[general]
# Site name
title = "My uMap"
# Site URL
site_url = "https://map.example.com/"
# Site description
description = "My uMap description"
# Site logo (optional)
logo = "/static/umap.png"
# Site favicon (optional)
# favicon = "/static/favicon.ico"

# Default language
language = "de"
# Default map center
default_latitude = "49.90305"
default_longitude = "15.77054"
default_zoom = "16"
# Map background
background_layer = "osm"
# Map tile server
tile_layer = "https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
# Map attribution
attribution = "© OpenStreetMap contributors"
SECRET_KEY = "1234"
#
# Do you want users to be able to create an account directly on your uMap instance (instead of only using OAuth).
# You need to create first a superuser, on the command line (open Terminal in Container Manager):
# > umap createsuperuser
#   umapadmin
#   umap@map.example.com
#   YOURPASSWORD
#
ENABLE_ACCOUNT_LOGIN=1
[...]

PR #1983, opened 2024-07-09
Create Synology tutorial

geimist:


almet:
Hi and thanks for opening this discussion here. It's a good document to have around for people wanting to follow the same path you took.

I'm not sure it should be included in this documentation, as it's a bit orthogonal to how to deploy umap itself.

Because it seems tied to sinology, which is not required (nor officially supported) by umap, I believe this document would probably get out of date at some point in the future.

To me, it seems more useful to publish this in a blog somehow, and we could eventually link to it from the new https://discover.umap-project.org/ website.

Would that make sense to you?

In any case, I'm pretty sure this will be helpful to other folks!

geimist:
I merely copied my answer in #1870 here at the [request](https://github.com/umap-project/umap/issues/1870#issuecomment-2218110216) of @davidbgk. Please decide for yourself how you would like to proceed. 🙂

almet:
> I merely copied my answer in https://github.com/umap-project/umap/issues/1870 here at the https://github.com/umap-project/umap/issues/1870#issuecomment-2218110216 of @davidbgk. Please decide for yourself how you would like to proceed. 🙂

Sorry, I wasn't aware of the context :-) I'll sync up with David then!

Thanks again.

PR #1987, opened 2024-07-11
try to fix toZoom and fromZoom randomly failing tests

yohanboniface:
Also, define a custom page fixture so we can get console messages in the terminal.

### DIFF
--- a/umap/tests/integration/test_datalayer.py
+++ b/umap/tests/integration/test_datalayer.py
@@ -1,4 +1,5 @@
 import json
+import re
 
 import pytest
 from django.core.files.base import ContentFile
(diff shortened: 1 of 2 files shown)


umap/tests/integration/test_datalayer.py
```diff
@@ -59,13 +62,14 @@ def test_should_honour_toZoom(live_server, map, datalayer, page):
     page.goto(f"{live_server.url}{map.get_absolute_url()}#6/48.55/14.68")
```
davidbgk, line 62:
```suggestion
    page.goto(f"{live_server.url}{map.get_absolute_url()}#6/48.55/14.68")
    expect(page).to_have_url(re.compile(r".*#6/48\..+/14\..+"))
```

Would it help given the current failing test?

davidbgk (approved):
🤞

umap/tests/integration/test_datalayer.py
yohanboniface: We had the same idea! :)

PR #2008, opened 2024-07-22
feat: add {layer} as popup variable (aka extended properties)

### LINKED ISSUE #2007: {layer} in Popup content template does not work
**Describe the bug**

By default the Popup content template has the tag {layer} in it. It seems like this should show the name of the layer an element is in in its popup, but it does not.

**To Reproduce**
1. Just make any element in any layer with a name. Click on it and notice that the popup does not have the name of the layer in it.

**Screenshots**
image

Thank you so much for your work!

yohanboniface:
fix #2007

### DIFF
--- a/docs-users/support/faq.md
+++ b/docs-users/support/faq.md
@@ -46,3 +46,44 @@ With macOS, replace `Ctrl` by `Cmd`.
 * `mycolumn<12.34` → will match features whose column `mycolumn` is lower than `12.34` (as number)
 
 When the condition match, the associated style will be applied to the corresponding feature.
+
+
+## How to use variables ? {: #variables}
+
+In general, using a variable is as simple as `{myvar}`.
+
+It's possible to define another variable as fallback of the first one like this: `{myvar|fallbackvar}`.
+
+To fallback to a string, add it between double quotes: `{myvar|"fallback"}`.
+
+It's possible to combine more variables: `{myvar|othervar|"some string"}`.
+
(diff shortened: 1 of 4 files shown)


davidbgk (approved):
I wonder if we should name it `layerName` instead 🤔 

By the way, is there any documentation about these available variables? Maybe something to add here: https://discover.umap-project.org/support/faq/

yohanboniface:
> I wonder if we should name it layerName instead 🤔

Given it's a user facing variable, I'd keep simple.

> By the way, is there any documentation about these available variables? Maybe something to add here: https://discover.umap-project.org/support/faq/

Ok, let's do that!

docs-users/support/faq.md
```diff
@@ -46,3 +46,42 @@ With macOS, replace `Ctrl` by `Cmd`.
...
+It's possible to use a variable inside an URL, for example: `[[https://domain.org/?locale={locale}|Wikipedia]]`.
+
+Or even as source for an image: `{{{myvar}}}` (note the triple `{}`).
+
+### Available variables in text fields:
+
+Those variables can be used in a feature description, or in popup content templates.
+
+Any property of the feature will be available, plus:
+- `{lat}/{lng}` → the feature position (or centroid in case of line or polygon)
```
davidbgk, line 70: In the test we have `{lon}`, is that an error? (same below)

yohanboniface: No, we support both, but though to only document one, to keep the doc simple. Should I add `lon` to ?

davidbgk: I'm OK with only documenting this one 👍

PR #2010, opened 2024-07-22
Fix variable pointing to null value displayed instead of the fallback

yohanboniface:
cf https://github.com/umap-project/umap/issues/820#issuecomment-2227821746

(Also pass biome on js tests in a separate commit)

### DIFF
--- a/umap/static/umap/js/modules/utils.js
+++ b/umap/static/umap/js/modules/utils.js
@@ -240,8 +240,7 @@ export function greedyTemplate(str, data, ignore) {
       if (staticFallback !== undefined) {
         vars.push(staticFallback)
       }
-      for (let i = 0; i < vars.length; i++) {
-        path = vars[i]
+      for (const path of vars) {
         if (path.startsWith('"') && path.endsWith('"')) {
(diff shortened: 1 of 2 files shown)


umap/static/umap/js/modules/utils.js
```diff
@@ -242,10 +242,12 @@ export function greedyTemplate(str, data, ignore) {
       }
       for (let i = 0; i < vars.length; i++) {
```
davidbgk, line 243:
```suggestion
      for (const path of vars) {
```

davidbgk (approved):

umap/static/umap/js/modules/utils.js
yohanboniface: Can you also remove `path = xxx` in the patch or should I ?

umap/static/umap/js/modules/utils.js
```diff
@@ -242,10 +242,12 @@ export function greedyTemplate(str, data, ignore) {
...
         path = vars[i]
```
davidbgk, line 244:
```suggestion
```

PR #2025, opened 2024-07-29
fix: catch login_required from map page and add a way to login

yohanboniface:
This:
- catch "login_required" from the map page, when hitting the save button
- proceed the save and close the login page at the end of the login process
- make sure to update the map user after a save

It's not ideal because when the save process includes a magic link (which can be the case with MonComptePro eg.), the new window will never be closed, and the save will not proceed, so the user will have to close the new window and click save again. Then the map will save and the user name will be shown in the edit toolbar as normal.

But that's the better compromise I can get to…

### DIFF
--- a/umap/templates/umap/login_popup_end.html
+++ b/umap/templates/umap/login_popup_end.html
@@ -5,10 +5,13 @@
 </h3>
 <script type="text/javascript">
   function proceed() {
-    if (window.opener && window.opener.umap_proceed) {
-      window.opener.umap_proceed()
+    if (window.opener?.onLogin) {
+      // We are in the normal process login
+      window.opener.onLogin()
+      window.close()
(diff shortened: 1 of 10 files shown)


umap/templates/umap/login_popup_end.html
```diff
@@ -2,10 +2,13 @@
 <h3>{% trans "You are logged in. Continuing..." %}</h3>
 <script type="text/javascript">
   function proceed() {
-    if (window.opener && window.opener.umap_proceed) {
-      window.opener.umap_proceed()
+    if (window.opener && window.opener.onLogin) {
```
davidbgk, line 5:
```suggestion
    if (window.?opener.?onLogin) {
```

todo: run biome against html files 😇

PR #2032, opened 2024-07-30
feat: allow to star a map from query string

yohanboniface:
Some users hide the star button from their map (to keep the UI clean I guess), but as a uMap instance administrator I may want to star it anyway, to make it appear on the home, when using the "starred" feed, or to save a good showcase.

That's an hidden feature, but we may use it to star a map from a list at some point.


yohanboniface:
Not needed anymore since #2442

PR #2035, opened 2024-07-31
fix: fix georss import

yohanboniface:
With a very basic test.


umap/static/umap/js/modules/formatter.js
```diff
@@ -106,7 +106,7 @@ export class Formatter {
   }
 
   async fromGeoRSS(str) {
-    return GeoRSSToGeoJSON(this.toDom(c))
+    return GeoRSSToGeoJSON(this.toDom(str))
```
davidbgk, line 109: 🙈

davidbgk (approved):

PR #2041, opened 2024-08-02
feat: allow to display a polygon "negative"

### LINKED ISSUE #1125: Display a polygon/multipolygon "negative"
From time to time I see this kind of map:

image

Would be great if one could draw a polygon or multi, and then ask uMap to display it reverse, i.e. clear inside and coloured outside (from the world border).

yohanboniface:
fix #1125

Refactor a bit the API and separation of concerns of our Feature and UI elements.

[Screencast from 2024-08-02 12-08-41.webm](https://github.com/user-attachments/assets/2d10a073-bac1-4524-a117-59f58bb211a3)

### DIFF
--- a/umap/static/umap/js/modules/data/features.js
+++ b/umap/static/umap/js/modules/data/features.js
@@ -114,8 +114,7 @@ class Feature {
   }
 
   makeUI() {
-    const klass = this.getUIClass()
-    this._ui = new klass(this, this.toLatLngs())
+    this._ui = new this.uiClass(this, this.toLatLngs())
   }
 
   getClassName() {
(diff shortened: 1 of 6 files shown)


umap/static/umap/js/modules/data/features.js
```diff
@@ -75,13 +80,41 @@ class Feature {
...
+      this.sync.update('geometry', this.geometry)
+    }
+  }
+
+  fromLatLngs(latlngs) {
+    this._geometry = this.convertLatLngs(latlngs)
+  }
+
+  makeUI() {
+    const klass = this.getUIClass()
```
davidbgk, line 116: Why not a getter for UI class?

davidbgk (approved):

Issue #2038, opened 2024-07-31
Misalignment of map icons with right-to-left user interface of umap


iriman:
**Describe the bug**
In RTL environment map icons and tooltips are out of correct position. I found 3 issues with this map:

![image](https://github.com/user-attachments/assets/f4d1f76a-b7b0-4da5-a184-d6f11819d456)

**To Reproduce**
1. Go to 'https://umap.openstreetmap.fr/fa-ir/map/map_943303#8/33.891/51.117' 

**Desktop:**
 - OS: Windows 11
 - Browser Firefox
 - Version 128

fim100:
Label placement also has problem in RTL mode that I couldn't fix:

![image](https://github.com/user-attachments/assets/f3371343-0f07-4d85-967c-34e759e48d79)

https://umap.openstreetmap.fr/fa-ir/map/test-map_1100282
https://umap.openstreetmap.fr/en/map/test-map_1100282

fim100:
issues here:
- [x] misalignment in cluster icon 
- [x] misalignment in parts of icon shapes: default, drop and ball icons 
- [ ] 'edit tooltip' out of the position 
- [ ] label out of the position

fim100:
All should be solved by #2046

yohanboniface:
This is now deployed in umap.openstreetmap.fr, should all be good. Thanks a lot @iriman for spotting the issues, and @fim100 for actually fixing them! :+1:

PR #2044, opened 2024-08-02
right to left support

fim100:
This would fix almost all RTL issues except two I know:
- 'Edit tooltip' that is mentioned in #2038 (number 3)
- alignment of object labels (picture below)
image

### DIFF
--- a/umap/static/umap/base.css
+++ b/umap/static/umap/base.css
@@ -121,11 +121,18 @@ dt {
 .col {
     float: left;
 }
+html[dir="rtl"] .col {
+  float: right;
+}
 .right {
     float: right;
 }
 .col + .col {
     padding-left: 20px;
+}
+html[dir="rtl"] .col + .col {
+  padding-right: 20px;
+  padding-left: 0;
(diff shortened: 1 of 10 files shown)


davidbgk:
Thanks a lot for your PR @fim100 ! Really appreciated 🙇 

I wonder if we could make the base CSS more “RTL-friendly” using [CSS logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_logical_properties_and_values) ([support](https://caniuse.com/css-logical-props)) to avoid part of the duplication here. A [recent example](https://frontendmasters.com/blog/what-you-need-to-know-about-modern-css-spring-2024-edition/#what-are-logical-properties) amongst others.

Do you have any experience in these?

fim100:
You're welcome! 

Not that much experience, but I can do a text replacement for these rules and discard redundant ones.

Yet we would have some html[dir=] rules, mostly "bakcground-position"s.

If there is no worries about browser support, this replacement should be a simple task for another PR.

PR #2046, opened 2024-08-05
Rtl support using css logical properties

fim100:
follow up on #2044

### DIFF
--- a/umap/static/umap/vendors/leaflet/leaflet.css
+++ b/umap/static/umap/vendors/leaflet/leaflet.css
@@ -481,7 +481,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	}
 .leaflet-popup-content-wrapper {
 	padding: 1px;
-	text-align: left;
+	text-align: start;
 	border-radius: 12px;
 	}
 .leaflet-popup-content {
--- a/umap/static/umap/vendors/leaflet/leaflet.css
+++ b/umap/static/umap/vendors/leaflet/leaflet.css
@@ -481,7 +481,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	}
 .leaflet-popup-content-wrapper {
 	padding: 1px;
-	text-align: start;
+	text-align: left;
 	border-radius: 12px;
 	}
 .leaflet-popup-content {
--- a/umap/static/umap/vendors/markercluster/MarkerCluster.Default.css
+++ b/umap/static/umap/vendors/markercluster/MarkerCluster.Default.css
@@ -48,7 +48,7 @@
 .marker-cluster div {
 	width: 30px;
 	height: 30px;
-	margin-inline-start: 5px;
+	margin-left: 5px;
 	margin-top: 5px;
 
 	text-align: center;
--- a/umap/static/umap/vendors/markercluster/MarkerCluster.Default.css
+++ b/umap/static/umap/vendors/markercluster/MarkerCluster.Default.css
@@ -48,7 +48,7 @@
 .marker-cluster div {
 	width: 30px;
 	height: 30px;
-	margin-left: 5px;
+	margin-inline-start: 5px;
 	margin-top: 5px;
 
 	text-align: center;
--- a/umap/static/umap/base.css
+++ b/umap/static/umap/base.css
@@ -119,13 +119,13 @@ dt {
     margin-bottom: 2rem;
 }
 .col {
-    float: left;
+    float: inline-start;
 }
 .right {
     float: right;
 }
 .col + .col {
-    padding-left: 20px;
+    padding-inline-start: 20px;
 }
 .half {
     width: 50%;
@@ -143,7 +143,7 @@ dt {
     width: 100%;
 }
 .col + .wide {
-    padding-left: inherit;
+    padding-inline-start: inherit;
 }
 .mshow, .tshow {
     display: none;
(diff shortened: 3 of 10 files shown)


davidbgk:
WOW. Awesome 💚 Thank you so much @fim100 I'll review it tomorrow!

umap/static/umap/vendors/leaflet/leaflet.css
```diff
@@ -481,7 +481,7 @@ svg.leaflet-image-layer.leaflet-interactive path {
 	}
 .leaflet-popup-content-wrapper {
 	padding: 1px;
-	text-align: left;
```
davidbgk, line 484: This is not possible to modify CSS from the `vendors` folder, is that critical to left it as is?

umap/static/umap/vendors/markercluster/MarkerCluster.Default.css
```diff
@@ -48,7 +48,7 @@
 .marker-cluster div {
 	width: 30px;
 	height: 30px;
-	margin-left: 5px;
```
davidbgk, line 51: This is not possible to modify CSS from the `vendors` folder, is that critical to left it as is?

davidbgk:
There is currently an issue when clicking on `↩ Cancel edits` and then `Cancel` (yes, the double naming is confusing), it should not make the current work be lost but return to the map with edits unchanged. I'm trying to figure that out, it *should* not be affected by CSS changes.

davidbgk:
> There is currently an issue when clicking on `↩ Cancel edits` and then `Cancel` (yes, the double naming is confusing), it should not make the current work be lost but return to the map with edits unchanged. I'm trying to figure that out, it _should_ not be affected by CSS changes.

Nevermind, it was a local cache issue.

davidbgk:
There is still an issue with tooltip editing:
Capture d’écran, le 2024-08-08 à 15 55 56

It might well be related to https://github.com/Leaflet/Leaflet/issues/7201

davidbgk:
There is an issue in the dashboard for some icons:

Capture d’écran, le 2024-08-08 à 15 58 53

davidbgk:
So far, terrific job @fim100 👏

umap/static/umap/vendors/leaflet/leaflet.css
fim100: 'text-align:left' makes content of popup always left-aligned. But it actually depends on user content not the website language. Writing system of user content can be same as UI or different from it.

'text-align:start' also isn't always correct, since with this text-alignment is obtained using direction of UI (dir=rtl/ltr). 

So I suggest to ignore this change and instead we have a hint for users about description field of popup, so they know that it's possible to control direction and alignment of popup content by wrapping it in `<div dir="rtl/ltr" style="text-align:right/left"></div>`. It can also be a button to apply this.

umap/static/umap/vendors/markercluster/MarkerCluster.Default.css
fim100: This fixes cluster icon issue described in #2038

fim100:
> There is an issue in the dashboard for some icons:
> Capture d’écran, le 2024-08-08 à 15 58 53

thanks for this!
https://github.com/umap-project/umap/pull/2046/commits/b9bea5eddebc52e3493074f0a891305ac7dd9fe7 also fixes navigation links that are very close together.

fim100:
> There is still an issue with tooltip editing:
> Capture d’écran, le 2024-08-08 à 15 55 56
> 
> It might well be related to [Leaflet/Leaflet#7201](https://github.com/Leaflet/Leaflet/issues/7201)

Similar to https://github.com/Leaflet/Leaflet/issues/7201#issuecomment-713112492 we can temporary fix this 'tooltip position' and also 'label position' by adding these in a separate css stylesheet leaflet-rtlfix.css until leaflet is updated:
```
html[dir="rtl"] .leaflet-inplace-toolbar {
    direction: ltr;
}
html[dir="rtl"] .leaflet-inplace-toolbar > * {
    direction: rtl;
}
html[dir="rtl"] .leaflet-tooltip-pane {
    direction: ltr;
}
html[dir="rtl"] .leaflet-tooltip-pane > * {
    direction: rtl;
}
```

davidbgk:
> Similar to [Leaflet/Leaflet#7201 (comment)](https://github.com/Leaflet/Leaflet/issues/7201#issuecomment-713112492) we can temporary fix this 'tooltip position' and also 'label position' by adding these in a separate css stylesheet leaflet-rtlfix.css until leaflet is updated:

It looks to be a good option, can you add that fix at the top of `map.css`? Thanks!

umap/static/umap/vendors/markercluster/MarkerCluster.Default.css
davidbgk: I get it but the `vendors` folder contains files we copy/overwrite from `node_modules` installations. If you want to override that CSS rule, you need to do it within one of our under-control CSS file, I would say `map.css`.

(`vendors` are only commited to ease the deployment/release process.)

umap/static/umap/vendors/leaflet/leaflet.css
fim100: I'll create an issue for this, so for now we leave it as is.

davidbgk (approved):

PR #2063, opened 2024-08-15
Use preferred tile.openstreetmap.org URL

Dimitar5555:
For reference: https://github.com/openstreetmap/operations/issues/737


yohanboniface:
Thanks! I'd say the line 79 of the same file needs to be adapted too :)

Dimitar5555:
Nice catch! I've updated it.

yohanboniface (approved):

yohanboniface:
:+1:

Issue #1335, opened 2023-09-23
Separate datalayers settings and data POST


yohanboniface:
At this stage, the client makes a single call to save both datalayer's settings and data, it'd be cleaner to have two, because:

- now settings live in the DB
- that would make using the future API much easier (instead of dealing with a private `_umap_options` in a whole json file)

cf #1333 #175

jvmatl:
Agree this sounds like a good step toward an exportable, documented API and its use cases. 

(Again, I don't currently know much about your application's structure, I'm just thinking out loud about how I would structure my data to accomplish what umap does. Feel free to ignore anything here that is not helpful)

I watched the import process in in Chrome developer tools, and I see what you're talking about now: for purposes of discussion, it looks like the process is roughly:

* Create a map, get a MAPID **xxx** (POST to /map/create)
* Create a layer, get a LayerID **yyy** (POST to /map/xxx/datalayer/create)
* On the client:
   * user uploads/imports the point data (how is irrelevant here)
   * user tweaks settings, and then clicks save
* save the map-level settings (POST to /map/xxx/update/settings)
* save the layer (POST to /map/xxx/datalayer/update/yyy)
  * and this post includes json `settings` and a binary blob of `geojson`)

And so, as I understand it, the gist of this ticket is to split that last POST into two separate operations; one that updates settings, and another that updates the geojson feature data?

If I could make a suggestion, the url structure for the layer update is a little surprising; if you want to make it a bit more intuitive (and RESTy) for future API users while you're looking at this, consider something like

* PUT /map/xxx/datalayer/yyy/settings   {payload is settings object}
* PUT /map/xxx/datalayer/yyy/geojson   {payload is your geojson features}

(The main point being the URL structure -- I mean, to me, PUT makes more sense than POST in this context, since you are overwriting the whole object, but that's not a huge deal either way)

jvmatl:
(I had another question/thought, but it's not directly relevant to this so I thought it made sense to split it out: I noticed that you also have an option for a remote data url in the import controls -- how does that work? Does the map client dynamically load data directly from the URL? Or does the server pull from that URL and populate the geojson data on the server side? I'd try it, but I don't have a handy network data source.)

jvmatl:
@yohanboniface  - I was looking through the issue tracker, and it occurs to me that  this separation should also help you with what you are trying to accomplish with #1307 -- since you would be separating the resources, it would be easier to, say, allow somebody to edit mapdata but not, other settings..

[...]

PR #2072, opened 2024-08-26
WIP: rename DataLayer/Map.settings to DataLayer/Map.metadata and more

### LINKED ISSUE #1636: Uniformize the way we store metadata in the geojson
- map metadata are stored in `properties` of the top level
- datalayer metadata are stored in `_umap_options` of the `FeatureCollection` level
- feature metadata are stored in the `_umap_options` key inside `properties` at the `Feature` level

This should be more consistent.
We may have a separate key for all (eg. `metadata`, `style`, `umap`) at the same level of `properties`. But we may also decide that everything goes inside `properties`, and any key that is recognized by uMap would be used as a metadata (and exposed to user through specific form fields), and any key which is not part of its format would be considered as data, (and exposed to used through the non typed data form).
Maybe we should also separate the styling metadata (which may be used by other tools) from behaviour metadata, which are very tied to uMap itself.

cf #316 
cf https://github.com/mapbox/simplestyle-spec

cf #1635

yohanboniface:
The main goal is to clarify the use of options/settings/properties…

- `settings` should be reserved to Django
- `options` should be reserved to pure Leaflet context
- `properties` used in the geojson context, for user data

Main changes:
- `DataLayer.settings` is renamed to `DataLayer.metadata`
- `DataLayer.geojson` is renamed to `Datalayer.data`
- `DataLayer.metadata` are not saved anymore in the geojson data file
- `Map.settings` is renamed to `Map.metadata`
- `Map.metadata` is now a flat key/value object, not a geojson Feature anymore
- `Map.zoom` is now populated and reused

Note: U.Map is still inheriting from Leaflet Map, so it mixes `options` and `metadata`.

This changes is very intrusive, it changes

- the DB schema
- the way we store data in the FS (now without metadata)
- the way we backup map data

fix #1636
cf #1335


yohanboniface:
Closing as this is not going to be merged, but will serve as a direction from (hopefully) smaller PRs.

PR #2082, opened 2024-08-29
Send remote operations to peers when they join

almet:
Operations are now stored locally in memory. All operations are tied to an HLC (Hybrid Logical Clock), making it possible to order them consistently.

When a new peer join, it asks a random peer for the list of operations, and re-apply them locally.

A new type of message has been introduced, making it possible to relay messages to other peers via the server. Messages types names have been updated to follow CamelCase, and to be similar across the client and the server.

### DIFF
--- a/umap/static/umap/js/modules/sync/engine.js
+++ b/umap/static/umap/js/modules/sync/engine.js
@@ -1,6 +1,46 @@
 import { DataLayerUpdater, FeatureUpdater, MapUpdater } from './updaters.js'
 import { WebSocketTransport } from './websocket.js'
+import { HybridLogicalClock } from './hlc.js'
+import * as Utils from '../utils.js'
 
+/**
+ * The syncEngine exposes an API to sync messages between peers over the network.
+ *
+ * It's taking care of initializing the `transport` layer (sending and receiving
+ * messages over websocket), the `operations` list (to store them locally),
+ * and the `updaters` to apply messages to the map.
+ *
+ * You can use the `update`, `upsert` and `delete` methods.
+ *
+ * @example
+ *
+ * ```
+ * const sync = new SyncEngine(map)
+ *
+ * // Get the authentication token from the umap server
+ * sync.authenticate(tokenURI, webSocketURI, server)
+ *
+ * // Alternatively, start the engine manually with
+ * sync.start(webSocketURI, authToken)
+ *
+ * // Then use the `upsert`, `update` and `delete` methods.
+ * let {metadata, subject} = object.getSyncMetadata()
+ * sync.upsert(subject, metadata, "value")
+ * sync.update(subject, metadata, "key", "value")
+ * sync.delete(subject, metadata, "key")
+ * ```
+ *
+ * A `proxy()` method is also exposed, making it easier to use without having
+ * to specify `subject` and `metadata` fields on each call:
+ *
+ * @example
+ * ```
+ * // Or using the `proxy()` method:
+ * let syncProxy = sync.proxy(object)
+ * syncProxy.upsert("value")
+ * syncProxy.update("key", "value")
+ * ```
+ */
 export class SyncEngine {
   constructor(map) {
     this.updaters = {
@@ -9,6 +49,7 @@ export class SyncEngine {
       datalayer: new DataLayerUpdater(map),
     }
     this.transport = undefined
+    this._operations = new Operations()
   }
 
   async authenticate(tokenURI, webSocketURI, server) {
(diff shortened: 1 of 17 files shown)


almet:
> will such idea give us ability to locate each other in the map instantly ?

It is currently not on scope, but might come at a later time if needed.

ahspence53:
9 Vorlich Place Kilmarnock KA1 4NE
Thanks

almet:
I've just updated this branch with the latest changes, and I believe this is ready for a first round of review @yohanboniface. 

The current work has been tested with two sessions, but not more. It is able to bring together peers that weren't here before the editing session begin, and before the map has been saved.

> [!NOTE]
> - The map is currently not saved automatically
> - Scenarii where peers go offline for sometime (doing edits in the meantime) is not taken care of (yet).
> - I'm currently fallbacking on `Utils.randomId()` for the identifying part of the HLC; While this might be enough in practice, we could also reuse the `UUIDs` generated by the websocket server: once we are assigned the UUID, change this part on the HLC. I've not yet had the time to do this, but I believe we should do it before merging.

umap/static/umap/js/modules/sync/engine.js
```diff
@@ -27,39 +69,201 @@ export class SyncEngine {
...
 
-  // This method is called by the transport layer on new messages
+  _applyOperation(operation) {
+    const updater = this._getUpdater(operation.subject, operation.metadata)
+    updater.applyMessage(operation)
+  }
+
+  _updateLastKnownHLC(remoteHLC) {
+    if (!this._lastKnownHLC || remoteHLC > this._lastKnownHLC) {
+      this._lastKnownHLC = remoteHLC
```
yohanboniface, line 107: I don't get why we have both `operation._hlc` and `engine._lastKnownHLC`, are those the same value at the end ?

umap/static/umap/js/modules/sync/engine.js
```diff
@@ -27,39 +69,201 @@ export class SyncEngine {
...
+    const lastOperation = remoteOperations[remoteOperations.length - 1]
+    this._updateLastKnownHLC(lastOperation.hlc)
+
+    // Sort the local operations only once, see below.
+    const sortedLocalOperations = this._operations.sorted()
+    for (const remote of remoteOperations) {
+      if (!Operations.isLocalOperationNewer(remote, sortedLocalOperations)) {
+        this._applyOperation(remote)
+      } else {
+        console.log(
```
yohanboniface, line 217:
```suggestion
        debug(
```

umap/websocket_server.py
```diff
@@ -1,70 +1,170 @@
...
+
 async def join_and_listen(
     map_id: int, permissions: list, user: str | int, websocket: WebSocketClientProtocol
 ):
-    """Join a "room" whith other connected peers.
-
-    New messages will be broadcasted to other connected peers.
-    """
+    """Join a "room" with other connected peers, and wait for messages."""
     print(f"{user} joined room #{map_id}")
```
yohanboniface, line 109: I guess we want to log instead ?

yohanboniface (commented):
Sounds good to me!
I can't say I've been in all details, as it break my brain quickly, but I get the big picture.
Can't wait to release and test that for real! :)

umap/static/umap/js/modules/sync/engine.js
```diff
@@ -91,3 +295,140 @@ export class SyncEngine {
...
+   * applied locally, or if newer values already exist.
+   *
+   * @param {Object} remote the remote operation
+   * @param {Object[]} sortedLocalOperations the sorted list of local operations
+   * @returns bool
+   */
+  static isLocalOperationNewer(remote, sortedLocalOperations) {
+    // No operations are stored, no need to check
+    if (sortedLocalOperations.length <= 0) {
+      console.log('No operations are stored, no need to check')
```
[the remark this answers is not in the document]
almet: Updated all the calls to use `debug` instead. Thanks!

umap/static/umap/js/modules/sync/engine.js
```diff
@@ -91,3 +295,140 @@ export class SyncEngine {
...
+    )
+  }
+
+  /**
+   * Apply a list of remote operations locally
+   *
+   * - Update the clock to the
+   * @param {Array} remoteOperations
+   */
+  storeRemoteOperations(remoteOperations) {
```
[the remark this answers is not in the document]
almet: Thanks for the comment. Operations are not applied as part of this method. I've clarified the documentation in commit b2fb4e56a83dfc8c3896e9f3fbbd9b8117c86fc4

umap/static/umap/js/modules/sync/engine.js
almet: I've refactored a bit the code in commit 3089da160a9d303f8526cb4ea05908271d071e97 to be clearer and easier to follow, effectively removing this from the main `SyncEngine` class, and putting it in the `Operations` class instead.

umap/websocket_server.py
almet: Done :heavy_check_mark:

PR #2085, opened 2024-08-30
Fix warning in worflows

Jean-BaptisteC:
This PR updates checkout and setup-python actions, because versions used in repository use a deprecated version of node js
More information here:
- https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/
- [Checkout actions releases](https://github.com/actions/checkout/releases)
- Setup-[python actions releases](https://github.com/actions/setup-python/releases)

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -28,9 +28,9 @@ jobs:
         database: [postgresql]
 
     steps:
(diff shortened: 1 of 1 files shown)


davidbgk (approved):

davidbgk:
Thank you!

PR #2108, opened 2024-09-05
feat: add a "Copy as GeoJSON" entry in the contextmenu

yohanboniface:
I often need this and end up using the console to do so, so why not make this simpler and available for everyone ? :)

Screenshot from 2024-09-05 17-33-45


davidbgk (approved):
That's a big menu 🙈

yohanboniface:
Yeah, we may remove some items

PR #2109, opened 2024-09-05
chore: use our own contextmenu

yohanboniface:
Let's start from here to clean the contextmenu items.

Current status:

### Some entries are generic, and always present:

In preview mode:

image

In edit mode:

image

### Some entries are specific to a clicked feature

In preview mode (Marker):

image

In edit mode (Marker):

image

In preview mode (Polygon):

image

In edit mode (Polygon):

image

In preview mode (Line):

image

In edit mode (Line):

image

@Aurelie-Jallut Do you want to make suggestion on what to show or not in this menu, depending on the context (feature or not, preview or edit mode) ?


Aurelie-Jallut:
yes I can but I need a little time : )

davidbgk (approved):

yohanboniface:
@Aurelie-Jallut we merged, but this is just a code change, the need of cleaning is still valid :)

Aurelie-Jallut:
if we keep in mind that the right-click menu is a contextual menu, we can
- restrict actions to the selected item

Following this logic, in the case of the marker, we could reduce it like that : 
image

PR #2131, opened 2024-09-12
feat: very minimal CSV export of users and maps in admin

### LINKED ISSUE #2020: Add an export button in the admin panel
Currently it's tedious for non technical admins to export the list of users / list of maps. 
Same for maps.

It'd be great to have an export button that downloads a csv when clicked.
In order to avoid downloading the full db, being able to aply filters and export the selection would be interesting to have but not mandatory.

I've been copy pasting a paginated list into a table but I'm missing info.
Having a queryable api would be an interesting approach as it would open the data to non admins. 

**Additional context**
I've been doing this for a small instance.

yohanboniface:
fix #2020

### DIFF
--- a/umap/admin.py
+++ b/umap/admin.py
@@ -1,6 +1,29 @@
+import csv
+
+from django.contrib.auth.admin import UserAdmin as UserAdminBase
+from django.contrib.auth.models import User
 from django.contrib.gis import admin
+from django.http import HttpResponse
+from django.utils.translation import gettext_lazy as _
 
 from .models import DataLayer, Licence, Map, Pictogram, Team, TileLayer
+
+
+class CSVExportMixin:
+    actions = ["as_csv"]
+
+    @admin.action(description=_("CSV Export"))
+    def as_csv(self, request, queryset):
+        response = HttpResponse(
+            content_type="text/csv",
+            headers={"Content-Disposition": 'attachment; filename="export.csv"'},
+        )
+
+        writer = csv.writer(response)
+        writer.writerow(self.csv_fields)
+        for user in queryset:
+            writer.writerow(getattr(user, field) for field in self.csv_fields)
+        return response
 
 
 class TileLayerAdmin(admin.ModelAdmin):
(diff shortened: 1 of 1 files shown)


umap/admin.py
```diff
@@ -1,8 +1,31 @@
...
 
 
+class CSVExportMixin:
+    actions = ["as_csv"]
+
+    @admin.action(description=_("CSV Export"))
+    def as_csv(self, request, queryset):
+        response = HttpResponse(
+            content_type="text/csv",
+            headers={"Content-Disposition": 'attachment; filename="export.csv"'},
```
davidbgk, line 19: I would but a `csv_filename` on the class to get a comprehensible one (my download folder is a mess 🙈), like `umap_export_users.csv`? Even better with the date of the export?

yohanboniface: Or could I use the model name in lower case ?

umap/admin.py
```diff
@@ -11,10 +37,22 @@ class TileLayerAdmin(admin.ModelAdmin):
     list_editable = ("rank",)
 
 
-class MapAdmin(admin.GISModelAdmin):
+class MapAdmin(CSVExportMixin, admin.GISModelAdmin):
     search_fields = ("name",)
     autocomplete_fields = ("owner", "editors")
     list_filter = ("share_status",)
+    csv_fields = (
```
virgile-dev, line 44: Can you add the number of features the map contains @yohanboniface  ?

umap/admin.py
```diff
@@ -30,9 +68,23 @@ class TeamAdmin(admin.ModelAdmin):
     filter_horizontal = ("users",)
 
 
+class UserAdmin(CSVExportMixin, UserAdminBase):
+    csv_fields = [
```
virgile-dev, line 72: Can you add a user map count @yohanboniface ?

virgile-dev: Is there anything like last_edit containing the date of the last edit ?

umap/admin.py
yohanboniface: Nope, too costly to compute :/

PR #2147, opened 2024-09-17
fix: editing coordinates manually would not be saved

### LINKED ISSUE #2144: Coordinate updates not taken
**Describe the bug**
When I place a new marker on the map and **edit the lat and lon values in the Coordinates section** of the Feature properties pane, the marker will immediately jump to the right place on the map, but irrespective whether I edit or not edit also other feature properties and whether I close or not close the Feature properties pane before saving, the **lat and lon values are not taken upon hitting Save**. The same happens if I edit an already existing marker.

To work around the bug, I need to grab the correctly placed marker with the hand tool and move it a little on the map. Only then the lat and lon values will be stored at Save. This is especially frustrating, if people enter many marker locations this way and everything looks fine until they first time close the editing session after hours and reopen it to see that hours worth of precise placement work is gone.

**To Reproduce**
uMap instance: https://umap.openstreetmap.fr/ as of 15 Sep 2024
Map used: https://umap.openstreetmap.fr/en/map/milke-danda-trek_1112959
Adding many markers to the "Further attractions" layer and some to the "Trek details" layer

**Desktop:**
OS: Windows 10
Edge, 128.0.2739.67 (Official build) (64-bit)

yohanboniface:
The primary coordinates is a two-elements Array in Point.coordinates, but FormBuilder does not know how to edit an Array, so we edit the `ui._latlngs` instead.

The proper fix would have been to teach FormBuilder how to edit an Array, but that path is way longer, and given we plan to refactor it totally, let's wait for that.

fix #2144

### DIFF
--- a/umap/static/umap/js/modules/data/features.js
+++ b/umap/static/umap/js/modules/data/features.js
@@ -655,6 +655,7 @@ export class Point extends Feature {
           builder.restoreField('ui._latlng.lat')
           builder.restoreField('ui._latlng.lng')
         }
+        this.pullGeometry()
         this.zoomTo({ easing: false })
       },
     })


davidbgk (approved):
Alright, let's keep it in mind for the refactor!

PR #2151, opened 2024-09-19
fix: polygon with interactive=false was still interactive

yohanboniface:
This bug has been introduced when spliting features, as we now only set the Leaflet Polygon options with setStyle, will the `interactive` option is only used in the init by Leaflet.

See https://github.com/Leaflet/Leaflet/pull/9475

### DIFF
--- a/umap/static/umap/js/modules/rendering/ui.js
+++ b/umap/static/umap/js/modules/rendering/ui.js
@@ -344,6 +344,10 @@ const PathMixin = {
     }
     options.pointerEvents = options.interactive ? 'visiblePainted' : 'stroke'
     this.parentClass.prototype.setStyle.call(this, options)
+    // TODO remove me when this gets merged and released:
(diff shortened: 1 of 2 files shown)


umap/static/umap/js/modules/rendering/ui.js
```diff
@@ -344,6 +344,14 @@ const PathMixin = {
     }
     options.pointerEvents = options.interactive ? 'visiblePainted' : 'stroke'
     this.parentClass.prototype.setStyle.call(this, options)
+    // TODO remove me when this gets merged and released:
+    // https://github.com/Leaflet/Leaflet/pull/9475
+
+    if (options.interactive) {
```
davidbgk, line 350: `this._path.classList.toggle('leaflet-interactive', options.interactive)` ?

davidbgk (approved):

PR #2152, opened 2024-09-19
Update 2-first-map.md

C-Sophie:
Add texts and images for public officials to properly use uMap ANCT


davidbgk:
Aggregated in #2156

PR #2158, opened 2024-09-21
feat: delete datalayer's files on delete

yohanboniface:
Until now, uMap was not deleting files on delete, which can increase file storage a lot after some time.

The files are not deleted, but moved to a "purgatory" folder, from where they can be deleted after some time.


davidbgk (commented):
Is the plan to temporary set `UMAP_PURGATORY_ROOT = MEDIA_ROOT` for the OSMfr instance, then run `purge_purgatory` and finally set a dedicated `UMAP_PURGATORY_ROOT` for future purges?

yohanboniface:
> Is the plan to temporary set UMAP_PURGATORY_ROOT = MEDIA_ROOT for the OSMfr instance, then run purge_purgatory and finally set a dedicated UMAP_PURGATORY_ROOT for future purges?

humm, seems to me that this would not work, apart from the management command that only consider root files (while MEDIA_ROOT has subfolders), this would delete a lot a files for datalayers that are not deleted, or am I missing something ?

Issue #1500, opened 2024-01-04
Download results of a facet search


virgile-dev:
**Is your feature request related to a problem? Please describe.**
I was working with a geojson file which contained properties. I wanted to filter it acording to some values and then upload the result to different layers. 
I'm no developer, but to get my filtered file was a bit of hassle. I ended up making a python script with the aid of chatGPT to create my filtered files.
It was frustrating because thanks to facet search I could easily filter my geojson within uMap.

**Describe the solution you'd like**
For noobs like me it would be great to be able to download the result of a facet search directly from uMap UI.
Adding a button inside the facet search pane "Download the result" when some checkboxes are clicked would be great.

**Describe alternatives you've considered**
Using python to filter my geojson.

yohanboniface:
Maybe in this particular case, the action you've needed is something like "Move result in a separate layer" ?

virgile-dev:
In that case yes. But maybe being able to download allows for more use cases

virgile-dev:
> Maybe in this particular case, the action you've needed is something like "Move result in a separate layer" ?
Facet search is available only in preview mode. So wouldn't there be some kind of issue trying to edit the layers ?

yohanboniface:
@Aurelie-Jallut let's have a look at this while we are hot on the browser! :)

PR #2160, opened 2024-09-23
feat: add a global toolbox in browser

### LINKED ISSUE #2097: Ability to isolate a layer
I have a map where I collect many GPX files uploaded from various sources.  After a few years, I ended up with over 100 layers.  I would like to be able to quickly isolate a layer for inspection, download etc.

A series of contextual commands that let me manage the visibility of multiple layers at a time would get ideal:

minimally, 2 commands that let me hide or show all layers would work:

- show all layers
- hide all alyers
(then I can unhide the layer of my choice)

another approach could be to have a command entitled:

- toggle visibility of selected layer(s) , assuming that layer selection is something possible

**Additional context**
Here is the type of map that I use where I would find this feature useful:

image

yohanboniface:
fix #2097
cf #1500

image

This is an alternative proposal for #2097:

- adding the button near the "Filter" toggle was a bit ugly, CSS/HTML talking, because this "Filters" box is a details/summary
- I thought that other needs could live in the same place: a link to download the filtered/visible data, and a button to show the map data extent

The "toggle all" button works this way:
- if one or more layers are shown, it will hide those layers
- otherwise it will show all layers

I've preferred not to change the button icon according to map state, because I didn't find a reliable way to do so, thus I'm afraid to have inconsistency, but I can't try harder if needed.

Related: I'm thinking of adding a "Download" button for each layers, as it's a request that comes quite often.

@Aurelie-Jallut could you have a look ? :)

### DIFF
--- a/umap/static/umap/js/modules/browser.js
+++ b/umap/static/umap/js/modules/browser.js
@@ -1,6 +1,7 @@
 import { DomEvent, DomUtil, stamp } from '../../vendors/leaflet/leaflet-src.esm.js'
 import { translate } from './i18n.js'
 import * as Icon from './rendering/icon.js'
+import * as Utils from './utils.js'
 
 export default class Browser {
   constructor(map) {
@@ -164,6 +165,7 @@ export default class Browser {
     })
     this.filtersTitle = container.querySelector('summary')
     this.toggleBadge()
+    this.addMainToolbox(container)
     this.dataContainer = DomUtil.create('div', '', container)
 
     let fields = [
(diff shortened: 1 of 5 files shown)


Aurelie-Jallut:
We could also differentiate the hide all / show all action from the 2 others, which are more related to the layers displayed.

image

yohanboniface:
Thanks!
Dark background seems reserved for the edit area to me, wdyt ?
Also, I don't really get why separate the toggle button from the others ?

Aurelie-Jallut:
_The dark background seems to be reserved for the editing area, doesn't it?_

um is that a problem? not sure. it could also be justified by the fact that there are additional editing functions when we edit.
The aim here too is to contrast the different elements with each other to make them easier to read.


Also, I don't really understand why separate the toggle button from the others?
I wanted to try something here :  differentiate the 2 buttons linked to the visible layers : 
- zoom for the visible layers
- Export for the visible layers

with the one that does something different: Hide all layers / Show all layers

yohanboniface:
Discussed just now: let's add a drop down menu for download options instead of opening another panel and switch context.

Aurelie-Jallut:
Test: no grey background, just a line

image


To compare : 
image

yohanboniface:
I prefer the simple line, as you may expect, but… I'll let you decide :)

Just implemented the dropdown:

image

image

Should I make it open above the button (now it's on top of it) ?

Aurelie-Jallut:
below (au dessous) :)

yohanboniface:
Oops, yeah, I meant below :p 

Here it is:

image

davidbgk (approved):

PR #2169, opened 2024-09-24
chore: display user maps in admin and add in csv export

yohanboniface:
maps are owned maps + maps as editor

cf https://github.com/umap-project/umap/pull/2131#discussion_r1772775039


umap/admin.py
```diff
@@ -77,7 +82,13 @@ class UserAdmin(CSVExportMixin, UserAdminBase):
         "last_name",
         "last_login",
         "date_joined",
+        "maps",
     ]
+    list_display = list(UserAdminBase.list_display) + ["maps"]
+
+    def maps(self, obj):
```
davidbgk, line 89: `maps_count` would be more explicit (in case we want to return `ids` at some point)

davidbgk (approved):

PR #2184, opened 2024-10-01
7-Where to find data to make a map on uMap

C-Sophie:
Where to find data, how to clone a map from another instance


davidbgk:
Done in https://github.com/umap-project/umap/commit/61de72ce85b9e00c5d0d4bc515e6ab186097bb2d

PR #2187, opened 2024-10-03
feat: allow to define tooltip size

### LINKED ISSUE #1884: labels : give more options to let read better
IDEA :
In some maps whe may need to display different kind of labels. In the same map same labels are more important.
image
https://umap.incubateur.anct.gouv.fr/fr/map/fr84-donnees-pdh-2024-04_319#11/44.0197/4.9871

SOLUTION :
Some would be bold and others normal, or a little smaller.

yohanboniface:
image

Or do we want a slider to change the font size more gradually ?

fix #1884

### DIFF
--- a/umap/static/umap/js/modules/data/features.js
+++ b/umap/static/umap/js/modules/data/features.js
@@ -312,6 +312,7 @@ class Feature {
       'properties._umap_options.showLabel',
       'properties._umap_options.labelDirection',
       'properties._umap_options.labelInteractive',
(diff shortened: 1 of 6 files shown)


yohanboniface:
Suggestion for improvement when there are many points with tooltips:
- only show markers actually on the map view box
- if there are more than 20 (?) points with tooltips, only render medium and high ones
- if there are more than 100 (?) points, only high are rendered
- if more than 1000 (?) display none

PR #2188, opened 2024-10-03
feat: allow to search by code INSEE in communes importer

yohanboniface:
Otherwise when searching for a very commun name (like "Saint-Paul"), one may not find the right one among the x returned results.

Screenshot from 2024-10-03 16-59-39

Screenshot from 2024-10-03 16-53-49

Screenshot from 2024-10-03 17-00-55


umap/static/umap/js/modules/importers/communesfr.js
```diff
@@ -25,7 +37,7 @@ export class Importer {
       textContent: "Importer les contours d'une commune française.",
     })
     const options = {
-      placeholder: 'Commune…',
+      placeholder: 'Nom ou code INSEE…',
```
davidbgk, line 40: We will have to replace all these placeholders soon #a11y

davidbgk (approved):

PR #2196, opened 2024-10-04
tests: use `at.least` to compare dates on the hlc tests.

almet:
Should fix [this error](https://github.com/umap-project/umap/actions/runs/11165503827/job/31037336947?pr=2188)


davidbgk (approved):
There is another one failing though https://github.com/umap-project/umap/actions/runs/11186769625/job/31102475717?pr=2196

almet:
I updated the test to account for the slow CI, let's see how it goes.

almet:
Thanks for the merge. Quick thing to note: "FIXUP" commits were meant to be squashed before being integrated the `master` branch, so they don't appear in the history, sorry they went trough.

PR #2205, opened 2024-10-10
chore: use pytest-rerunfailed

yohanboniface:
I first thought about marking test as flaky one by one, but my impression is mainly all playwright test can become flacky, and we actually only see some as their appear in the list.


davidbgk (approved):
🤞

Issue #2174, opened 2024-09-26
Sync: should we integrate with Django Channels?


almet:
As part of our work on real-time collaboration, we're now shipping a WebSocket server as part of uMap. @yohanboniface suggested to take another look at [Django Channels](https://channels.readthedocs.io/), hence this issue for tracking our thinking on the matter.

Pros:  

- Django channels integrate with Django concepts (URL routing, authentication), and can ease the addition of websocket-related 
features in the future (even if not related to the sync).
- Different channel layers r (the name they use for the pub/sub backend) are possible, and this is configurable via the settings. Note that only Redis is officially supported, but there is a third party PostgreSQL channel layer.

Cons:

- uMap currenlty relies on WSGI, and Django Channels requires ASGI instead; This would result in a change in the NGINX configuration, even for instances that do not want to enable the real-time collaboration feature.
- Django channels adds a layer of complexity¹ over doing things ourselves as we've done so far. Of course, this "added complexity" can also help us in some cases, but it's adding another layer on top of WebSocket, as we are using right now.

It's worth noting that some [work has been done](https://github.com/umap-project/umap/pull/1701) in the past to move to Django Channels.

yohanboniface:
I'm in favor of going that way. I've revived your PR here: #2209 

Here is a suggestion:
- create an asgi.py file, alongside the current wsgi.py, so we can run uMap either with asgi or wsgi (which is what the #2209 PR does, maybe we should add a bit of documentation about running asgi, with "experimental" in bold and upper case)
- allow to run the sync websockets with channels
- deploy ourselves with channels in our dev server and start testing
- for next major release, document the asgi deployment, and warn about wsgi behing deprecated, but users can still deploy with wsgi (without the sync feature)
- switch Docker image to asgi
- next major release +1 (or +2), properly remove wsgi support

WDYT ? :)

almet:
Let's go for a switch to Django Channels then :-) As mentioned in the chat, since we are reaching the end of the work for me on real-time collaboration, I think we should add support for Django Channels once all the rest is working, to avoid generating conflicts with the current design.

> create an `asgi.py` file, alongside the current `wsgi.py`, so we can run uMap **either with asgi or wsgi**

Nice to be able to have the best of both worlds! That opens for a nice transition, I like it.

After #2209 is merged, here is a list of remaining things to do (but not right now if possible):

- Rethinking the way we handle authentication (right now it's done via a specific URL endpoint)
[...]

PR #2209, opened 2024-10-14
Allow to run umap with asgi

yohanboniface:
Follow up of #1701 

cf #2174

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ install: ## Install the dependencies
 
 .PHONY: develop
 develop: ## Install the test and dev dependencies
-	python3 -m pip install -e .[test,dev]
+	python3 -m pip install -e .[test,dev,sync]
 	playwright install
 
 .PHONY: format
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,13 +34,11 @@ dependencies = [
   "django-probes==1.7.0",
   "Pillow==10.4.0",
   "psycopg==3.2.3",
-  "pydantic==2.9.2",
   "requests==2.32.3",
   "rcssmin==1.1.2",
   "rjsmin==1.2.2",
   "social-auth-core==4.5.4",
   "social-auth-app-django==5.4.2",
-  "websockets==13.1",
 ]
 
 [project.optional-dependencies]
@@ -70,6 +68,8 @@ docker = [
 sync = [
   "channels==4.1.0",
   "daphne==4.1.2",
+  "pydantic==2.9.2",
+  "websockets==13.1",
 ]
 
 [project.scripts]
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -67,6 +67,10 @@ test = [
 docker = [
   "uwsgi==2.0.27",
 ]
+sync = [
+  "channels==4.1.0",
+  "daphne==4.1.2",
```
almet, line 72: Good idea to add a `sync` target here! I think we could also add a few other dependencies in here, such as `websockets` and `pydantic`.

I wonder how this will work with the django management commands, though.

umap/asgi.py
```diff
@@ -0,0 +1,16 @@
...
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "umap.settings")
+# Initialize Django ASGI application early to ensure the AppRegistry
+# is populated before importing code that may import ORM models.
+django_asgi_app = get_asgi_application()
+
+application = ProtocolTypeRouter(
+    {
+        "http": django_asgi_app,
+        # Just HTTP for now. (We can add other protocols later.)
```
almet, line 14: nit: I didn't mean to commit this in the end, should be removed :-)

almet (changes requested):
Nice. A few remarks on the sync target, and we should be good to go!

almet:
Also noting that this doesn't really fix #2174, because it would require us to move the websocket handling to django channels directly, and a few other things mentioned there.

umap/asgi.py
yohanboniface: Done :)

pyproject.toml
almet: > Not sure what you wonder here ? You mean imports that could fail because the package are not installed but still imported automatically ?

That's exactly what I have in mind. We should ensure that these doesn't make the command fail when the `sync` group is not installed.

Makefile
```diff
@@ -7,7 +7,7 @@ install: ## Install the dependencies
 
 .PHONY: develop
 develop: ## Install the test and dev dependencies
-	python3 -m pip install -e .[test,dev]
+	python3 -m pip install -e .[test,dev,sync]
```
almet, line 10: Seeing this makes me also think we should add this group in the tests, when running the CI (not sure if it uses `make develop`).

yohanboniface: It does use `make develop`, I've checked :)

pyproject.toml
almet: I was thinking about just checking manually, but doing so in the CI wouldn't hurt. It's not a requirement though.

yohanboniface: Not sure how to do this through the CI, because this would mean changing the deps during the process, or having another dedicated recipe, may be overkill, no ? Or do you have another idea ?

almet (approved):

almet:
LGTM, Let's merge!

davidbgk (approved):

PR #2217, opened 2024-10-15
fix: do not show browser counter when layer is empty or not loaded

### LINKED ISSUE #2215: Hidden layers show '0' elements on load
**Describe the bug**
Due to lazy loading of hidden layers the initial number of features is shown as zero.

**To Reproduce**
1. Set properties of a layer with several features to hide on load
2. Map settings: Show databrowser on load
3. Open the map - invisible layers are shown as having zero features

**Expected behavior**
Either the number `0` indicating there wouldn't be any features should be hidden until the geojson is loaded an the correct number can be shown. Or the number of features belongs into the database so that the correct number can be shown when the map is loaded (in gray as if you hide an already visible layer).

**Screenshots**
Image

yohanboniface:
fix #2215

### DIFF
--- a/umap/static/umap/js/modules/browser.js
+++ b/umap/static/umap/js/modules/browser.js
@@ -98,6 +98,7 @@ export default class Browser {
     datalayer.eachFeature((feature) => this.addFeature(feature, container))
 
     const total = datalayer.count()
+    if (!total) return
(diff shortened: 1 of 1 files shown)


davidbgk (approved):
An early return would work too given that's a pattern already in use at line 85.

PR #2221, opened 2024-10-18
fix: use correct icon for layer download advanced action

yohanboniface:
Before:
image

After:
image


umap/static/umap/js/modules/data/layer.js
```diff
@@ -802,13 +802,11 @@ export class DataLayer {
...
-        'button umap-download',
-        advancedButtons,
-        translate('Download'),
-        this._dataUrl(),
-        '_blank'
-      )
+      const download = Utils.loadTemplate(`
+        <a class="button" href="${this._dataUrl()}" target="_blank">
+          <i class="icon icon-24 icon-download"></i>${translate('Download')}
+        </button>`)
```
davidbgk, line 808:
```suggestion
        </a>`)
```

umap/static/umap/js/modules/data/layer.js
```diff
@@ -802,13 +802,11 @@ export class DataLayer {
...
     if (this.umap_id) {
-      const download = DomUtil.createLink(
-        'button umap-download',
-        advancedButtons,
-        translate('Download'),
-        this._dataUrl(),
-        '_blank'
-      )
+      const download = Utils.loadTemplate(`
+        <a class="button" href="${this._dataUrl()}" target="_blank">
```
davidbgk, line 806: Instead of the `target` attribute, I think a `download` one with an explicit name would be more interesting for the user.

davidbgk (approved):

PR #2223, opened 2024-10-18
feat: add importer for French cadastre

yohanboniface:
Screenshot From 2024-10-18 12-48-23

### DIFF
--- a/umap/static/umap/js/modules/importers/cadastrefr.js
+++ b/umap/static/umap/js/modules/importers/cadastrefr.js
@@ -1,0 +1,62 @@
+import { DomUtil } from '../../../vendors/leaflet/leaflet-src.esm.js'
+import { BaseAjax, SingleMixin } from '../autocomplete.js'
+import * as Util from '../utils.js'
+import { AutocompleteCommunes } from './communesfr.js'
+
+const TEMPLATE = `
+  <h3>Cadastre</h3>
+  <p>Importer les données cadastrales d’une commune française.</p>
+  <select name="theme">
+    <option value="batiments">Bâtiments</option>
+    <option value="communes">Communes</option>
+    <option value="feuilles">Feuilles</option>
+    <option value="lieux_dits">Lieux dits</option>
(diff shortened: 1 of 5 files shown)


yohanboniface:
@Aurelie-Jallut do you want to create a dedicated icon ?

Aurelie-Jallut:
let's try !
image

image

umap/static/umap/js/modules/importers/cadastrefr.js
```diff
@@ -0,0 +1,62 @@
+import { DomUtil } from '../../../vendors/leaflet/leaflet-src.esm.js'
+import { BaseAjax, SingleMixin } from '../autocomplete.js'
+import * as Util from '../utils.js'
+import { AutocompleteCommunes } from './communesfr.js'
+
+const TEMPLATE = `
+  <h3>Cadastre</h3>
+  <p>Importer le cadastre d'une commune française..</p>
```
davidbgk, line 8:
```suggestion
  <p>Importer le cadastre d’une commune française…</p>
```

umap/static/umap/js/modules/importers/cadastrefr.js
```diff
@@ -0,0 +1,62 @@
...
+  <select name="theme">
+    <option value="batiments">Bâtiments</option>
+    <option value="communes">Communes</option>
+    <option value="feuilles">Feuilles</option>
+    <option value="lieux_dits">Lieux dits</option>
+    <option value="parcelles" selected>Parcelles</option>
```
davidbgk, line 14: I'm tempted to put the selected choice at the top of the list, it is a more classic pattern (and I get that it's sorted alphabetically for now but still).

davidbgk (approved):
So good 🙌

umap/static/umap/js/modules/importers/cadastrefr.js
yohanboniface: I'm tempted to think that given the list is small, the alphabetical order is more natural and convenient once open. Let's try that way and see ?

PR #2224, opened 2024-10-21
fix: use correct icon for layer download advanced action

yohanboniface:
Replace #2221


umap/static/umap/js/modules/data/layer.js
```diff
@@ -802,13 +802,12 @@ export class DataLayer {
...
     )
     if (this.umap_id) {
-      const download = DomUtil.createLink(
-        'button umap-download',
-        advancedButtons,
-        translate('Download'),
-        this._dataUrl(),
-        '_blank'
-      )
+      const filename = Utils.slugify(this.options.name)
```
davidbgk, line 805: Probably miss an extension?

yohanboniface: Meeh, good point!

yohanboniface: humm, actually, the downloaded file does have an extension. I guess because we are calling the back office, which set one. So does the `download` attribute still make sense ?

yohanboniface: No, it's not the backend that sets this extension. Can it be firefox adding it, as the backend sets the content-type to `application/geo+json ?

yohanboniface: The name of the layer ;)

image

PR #2233, opened 2024-10-22
feat(sync): Let the clients set layers UUID

almet:
This make it possible to synchronize datalayers before their creation on the server, allowing at the same time to solve issues related to them not being saved (e.g. duplication of geometries)


umap/migrations/0023_alter_datalayer_uuid.py
```diff
@@ -0,0 +1,21 @@
...
+
+    operations = [
+        migrations.AlterField(
+            model_name="datalayer",
+            name="uuid",
+            field=models.UUIDField(
+                default=uuid.uuid4, primary_key=True, serialize=False, unique=True
+            ),
+        ),
+    ]
```
almet, line 21: I'm actually unsure this migration is needed, but this is what I got after updating the model. I'm curious about your feedback on this @yohanboniface ?

yohanboniface: I guess because we've set it to editable ? But I'm not sure yet why we needed this.

umap/static/umap/js/modules/data/layer.js
```diff
@@ -61,6 +61,13 @@ export class DataLayer {
     this._isDirty = false
     this._isDeleted = false
     this.setUmapId(data.id)
+
+    if (!this.umap_id) {
```
davidbgk, line 65: Why don't we reuse `createdOnServer` here?

umap/static/umap/js/modules/data/layer.js
```diff
@@ -36,7 +36,7 @@ const LAYER_MAP = LAYER_TYPES.reduce((acc, klass) => {
 }, {})
 
 export class DataLayer {
-  constructor(map, data) {
+  constructor(map, data, sync, future_uuid) {
```
yohanboniface, line 39: I don't get why `sync` is needed here.

yohanboniface: Also, at this time, I did not understand the need of `future_uuid`, can you elaborate a bit on why we can't just have an id/uuid ?

umap/static/umap/js/modules/data/layer.js
```diff
@@ -115,11 +129,28 @@ export class DataLayer {
...
   }
 
+  /**
+   * When receiving a reference version, discard the future uuid
+   * because the layer is now created on the server.
+   */
+  set _reference_version(version){
+    console.debug("set _reference_version", version)
+    if (version !== undefined) {
+      this.__reference_version = version
```
yohanboniface, line 139: Is there any reason we use double underscore here ?

umap/static/umap/js/modules/data/layer.js
```diff
@@ -1040,14 +1073,17 @@ export class DataLayer {
     // Filename support is shaky, don't do it for now.
     const blob = new Blob([JSON.stringify(geojson)], { type: 'application/json' })
     formData.append('geojson', blob)
-    const saveUrl = this.map.urls.get('datalayer_save', {
+    console.log("map_id", this.map.options.umap_id, "pk", this.umap_id || this._future_uuid, "created", this.createdOnServer)
+    const saveURL = this.map.urls.get('datalayer_save', {
       map_id: this.map.options.umap_id,
-      pk: this.umap_id,
+      pk: this.umap_id || this._future_uuid,
+      created: this.createdOnServer,
```
yohanboniface, line 1080: Instead of passing `created`, I'd better make here a condition and ask for `datalayer_update` or `datalayer_create`  depending on it

umap/static/umap/js/modules/schema.js
```diff
@@ -563,4 +563,9 @@ export const SCHEMA = {
     type: Object,
     impacts: ['data'],
   },
+
+  _reference_version: {
```
yohanboniface, line 567: On my understanding, this property does not stand in the json metadata, and I understand the schema as a (simple to be refined) representation of it.

umap/static/umap/js/modules/sync/engine.js
```diff
@@ -1,7 +1,7 @@
+import * as Utils from '../utils.js'
+import { HybridLogicalClock } from './hlc.js'
```
yohanboniface, line 2: Comment unrelated to this PR, but this "biome noise" is a bit of a pity, while the main scope of biome in my understanding is to remove all formatting noise from PR. How is it that this passed the CI ? @davidbgk any idea ?

umap/static/umap/js/modules/sync/updaters.js
```diff
@@ -31,8 +31,8 @@ class BaseUpdater {
     }
   }
 
-  getDataLayerFromID(layerId) {
-    if (layerId) return this.map.getDataLayerByUmapId(layerId)
+  getDataLayerFromID(layerId, future_uuid) {
```
yohanboniface, line 34: Can't we totally replace the layerId by the uuid ?

umap/static/umap/js/modules/sync/updaters.js
```diff
@@ -53,18 +53,22 @@ export class MapUpdater extends BaseUpdater {
 }
 
 export class DataLayerUpdater extends BaseUpdater {
-  upsert({ value }) {
-    // Inserts does not happen (we use multiple updates instead).
-    this.map.createDataLayer(value, false)
+  upsert({ value, metadata }) {
+    console.log('upsert', value, metadata)
+    // Upsert only happens when a new datalayer is created.
+    this.map.createDataLayer(value, false, metadata.future_uuid)
```
yohanboniface, line 59: Here I would expect the uuid to be in the value, not separate.

umap/migrations/0023_alter_datalayer_uuid.py
almet: Why we needed to make it editable? Or why the migration is required?

On the former: My understanding is that we need to make it editable, otherwise it's not possible (to my knowledge) to have it provided by the caller on creation. I'll actually try another time to do it without `editable=True`, as it seems better indeed.

yohanboniface: https://docs.djangoproject.com/en/5.1/ref/models/fields/#editable :)

umap/static/umap/js/modules/data/layer.js
almet: >  I did not understand the need of future_uuid / why we can't just have an id/uuid ?

We need a way to distinguish between three different scenarii here:

1. Creation happened locally, but is not yet synced on the server ;
2. Creation is coming from a remote peer ;
3. Creation is coming from an already saved map, with existing layers.

`future_uuid` is meant to solve the first two, when a layer is created locally on the map, but the server isn't yet aware of it. In this case, we need to decide on the uuid locally, but we need a way to know that the server doesn't know yet about it, so we can ask it to save. That's the purpose of this field.

In an earlier iteration, I tried to track the state of the server by using the `version` info provided by the server, but I believe this design is better because it's more explicit that the uuid isn't yet saved.

>  don't get why sync is needed here.

Ah :-) ` sync` is a flag to tell you if you need to replicate the creation on other peers via the sync protocol. It is not used at the moment, but it is passed by `map.createDataLayer()`. I can remove it!

umap/static/umap/js/modules/data/layer.js
almet: We now can, this is actually a leftover from a previous iteration, where I couldn't actually rely on `createdOnServer`. Let's change this!

umap/static/umap/js/modules/data/layer.js
almet: There is a need to react to the change of the `_reference_version`, to be able to unset the `_future_uuid` property.

Specifically, this is useful when receiving a (potentially new) reference version from another peer (this is a signal that a save has been triggered). 

That's why I'm using a setter here. the double underscore is to avoid infinite recursions!

umap/static/umap/js/modules/schema.js
almet: Because we now check that the properties are part of the schema before applying them in the updaters, the replication of the `_reference_version` field wasn't replicated.

As a solution, I added it to the schema. My way of thinking about the schema isn't only a representation of the JSON metadata, it's a way to tell what needs to be updated when a property is updated on the map / datalayer.

As such, some properties might not belong to the JSON files, but still need to trigger changes. I believe we'll need to set properties specific to the sync.

I see different ways to solve this:

- Let them in the schema here, and mark them with a specific thing, to tell they are sync-related.
- Create another schema, with these specific fields
- My favorite: special-case these in the updaters. After-all, it's part of the sync protocol, so it could be done there.

umap/static/umap/js/modules/sync/updaters.js
almet: Actually, the layerId **is** an UUID already. Here, we have potentially two different UUIDs. These are from calls in different times:

- A layerId is provided when the updated property of the layer is from a layer that's already saved on the server
- A futureUUID is provided when we're dealing with yet-to-be-saved layers.

I believe we need a way to tell them apart, and we also won't look for them the same way (see `getDataLayerByUmapId`, which is looking using `dl._future_uuid` if it's provided, otherwise `dl.umap_id`)

umap/static/umap/js/modules/schema.js
yohanboniface: Or we may actually put the reference_version in the json instead of in the headers ?

umap/static/umap/js/modules/sync/updaters.js
almet: It *could* also be in the value, but since it's not yet saved on the server I believe it's better to keep it this way. After all, metadata is that: a way to identify the data. Once it will be saved, then it will be part of the data (and also be sent in the metadata)

umap/migrations/0023_alter_datalayer_uuid.py
almet: That link says:

> If False, the field will not be displayed in the admin or any other [ModelForm](https://docs.djangoproject.com/en/5.1/topics/forms/modelforms/#django.forms.ModelForm). They are also skipped during [model validation](https://docs.djangoproject.com/en/5.1/ref/models/instances/#validating-objects). Default is True.

(My goal was to actually not skip the validation)

umap/static/umap/js/modules/schema.js
almet: Why not ! As this might cause some other problems down the line, is it okay to ship this as-is, and iterate on the subject in a later PR?

umap/static/umap/js/modules/data/layer.js
yohanboniface: >  I can remove it!

Thanks! I'd prefer the `sync` attribute to stay outside data classes as much as possible

> 2.  Creation is coming from a remote peer

Isn't actually what the `sync` attribute is for ? Or otherwise I don't get why we need to distinguish this case from the others (what do we need to do/prevent for this case ?).

I'm in favor of the goal of this PR, which is to move the uuid creation from the backend to the frontend. Doing so, I'm expecting that the front client that creates a new datalayer is the source of truth, and both other clients and server should follow.

And even for the other cases, it's not clear to me why we need to distinguish them. For me, always doing a PUT to the server would be good (as we are the source of truth when we save now), and then when creating a datalayer with data coming from the server, I'd expect that the server only act as a storage, so as soon as the datalayer is created on the frontend, it behave like any datalayer.

I feel like I'm missing something obvious, but I can't find it ;)

umap/migrations/0023_alter_datalayer_uuid.py
yohanboniface: Can't we let the DB validate it ? As it will, at the end, no ?

umap/static/umap/js/modules/data/layer.js
almet: I think what I did in this PR is one way to do it. It's not perfect, but it works without adding complexity to the overall design.

> Isn't actually what the `sync` attribute is for ? Or otherwise I don't get why we need to distinguish this case from the others (what do we need to do/prevent for this case ?).

Sort of, but `sync` is missing part of the info I need: the fact that it's not saved yet on the server. This is not provided by the `sync` property, as it's just saying that it's coming from another peer. It doesn't say if that peer has saved the data yet (or not).

I think that `_future_uuid` is an elegant way to solve this: it stores the uuid so it's possible to pass it to the server on save, and and the same time it's telling us that this datalayer hasn't been saved yet on the server (info that was provided before by looking at `this.umap_id !== undefined`.

> And even for the other cases, it's not clear to me why we need to distinguish them. For me, always doing a PUT to the server would be good (as we are the source of truth when we save now), and then when creating a datalayer with data coming from the server, I'd expect that the server only act as a storage, so as soon as the datalayer is created on the frontend, it behave like any datalayer.

I agree that it would be the ultimate goal to merge the two `/create/` and `/update` endpoints.

I will have another look at this tomorrow, to see if it's possible to merge the two `DataLayerCreate` and `DataLayerUpdate` classes together. The reason I didn't do it yet, is because it feels to be a bigger step.

With that being said, having these two endpoints merged in one might not remove the requirement of the `future_uuid` field, as we will still need to communicate about the datalayer (with other peers) before it's saved to the server, **and** convey the fact that it's not saved yet.

> it's not clear to me why we need to distinguish them.

Why do we need to know that it's not saved yet? For the same reasons we check if `umap_id` is defined or not: because if it's not saved we don't make a request to the server asking for it. Instead, we create it.

> I'd expect that the server only act as a storage, so as soon as the datalayer is created on the frontend, it behave like any datalayer.

I agree, and I believe that's what's happening right now in this PR 🙌

umap/migrations/0023_alter_datalayer_uuid.py
almet: I believe it's safer to check the inputs before passing them to the DB 💇

umap/static/umap/js/modules/data/layer.js
yohanboniface: Sorry, I still don't get the `_future_uuid` pattern :(

But I'm certainly missing the problem we are trying to solve.

For what I understand, I'd:

- make any peer always PUT (when dirty, of course)
- on reset/delete, always send a DELETE to the server, and don't care if it's a 404 (or rely on `reference_version` if we want to be explicit, but I'm not sure it's worth it)

I'm also tempted to not sync to other peers the datalayer unless it's persisted (if client A creates a datalayer without saving it, then sync, then client B adds data in this datalayer without saving, then client A cancel the datalayer creation, what happen to client B data ?).

almet: > I'm also tempted to not sync to other peers the datalayer unless it's persisted (if client A creates a datalayer without saving it, then sync, then client B adds data in this datalayer without saving, then client A cancel the datalayer creation, what happen to client B data ?).

The delete of the datalayer will also be synced, so we're good here :+1:

almet: > But I'm certainly missing the problem we are trying to solve.

The problem is: we want the ability to sync layers creation and property update before server-save. 

It's not at odds to what you're proposing here:

> For what I understand, I'd:
> 
> - make any peer always PUT (when dirty, of course)
> - on reset/delete, always send a DELETE to the server, and don't care if it's a 404 (or rely on reference_version if we want to be explicit, but I'm not sure it's worth it)

As I said yesterday I want to have another look at how to merge the two views together, to always do a PUT.

The only bit that stays in the way is the `_future_uuid` field, and merging the two views won't make the `_future_uuid` **concept** go away.

The current code looks at `dl.umap_id == undefined` in a few places and uses this information to act differently pre-save. I believe these cases to be here for a reason, and `_future_uuid` is just a way to not get in the way of it. We could do this with another field like `not_saved_on_server` or similar, but that will stay conceptually the same.

almet:
Just so things are clear in terms of who does what: @yohanboniface will be taking over the work on this PR, and merge it once it's ready.

Just noting a few tasks that still need to be done:

- [ ] Check if the `editable=True` is required on the model uuid field
- [ ] Try to merge Django views for datalayer create and update
- [ ] Squash "fixup" commits together (with `git rebase -i master`)

PR #2235, opened 2024-10-24
feat(sync): Reconnect the websocket on failures.

almet:
image

(I didn't manage to capture it, but you can see when the next try will happen on hover)

When the websocket is disconnected, try to redo an authentication roundtrip. This commit does the following changes:

- Change the way the SyncEngine is instanciated, passing it a server object and the URLs.
- Add a ping/pong mechanism. This is required because otherwise we have no certainty that the connection is still alive.
- Offer a "reconnect" mechanism, increasing the wait time a bit more each time.
- Trigger the reconnect on disconnections and on failure to connect to the websocket (after a specified time)
- When disconnected the "number of connected peers" now indicates "Disconnected".
- Display a dialog to the users saying that they are disconnected.

Todo:

- [ ] When offline: let the number of connected peers and add disconnected.
- [ ] Changer la couleur de la fenêtre 
- [ ] When offline: change the background color (not red, but something else)
- [ ]  [Here be dragons](https://en.wikipedia.org/wiki/Here_be_dragons). This map has enabled real-time synchronization with other users, but you are currently disconnected. We will try to reconnect in the background and reconcile with other users, but this feature is still experimental, and you might lose data. Have fun!

### DIFF
--- a/umap/static/umap/js/modules/sync/engine.js
+++ b/umap/static/umap/js/modules/sync/engine.js
@@ -48,7 +48,7 @@ const MAX_RECONNECT_DELAY = 32000;
  * ```
  */
 export class SyncEngine {
-  constructor(map, urls, server) {
+  constructor(map, websocketTokenURI, websocketURI, server) {
     this.updaters = {
       map: new MapUpdater(map),
       feature: new FeatureUpdater(map),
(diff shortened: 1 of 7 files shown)


almet:
I've just added a screen that warns the user when it's disconnected. It looks like this:

image

The template is currently defined in the `umap.js` file, but maybe it should live in the `sync` folder instead? I'm curious about where you think this should be?

yohanboniface:
Replaced by #2389

PR #2238, opened 2024-10-25
feat(sync): Only send sync events after one second of inactivity

almet:
This is a first take, and probably doesn't really cut it as events  from different sources will overwrite each other (if somebody is fast enough to do two actions on different objects in less than 1 second, but still).

### DIFF
--- a/umap/static/umap/js/modules/utils.js
+++ b/umap/static/umap/js/modules/utils.js
@@ -430,3 +430,15 @@ export function deepEqual(object1, object2) {
 export function slugify(str) {
   return (str || 'data').replace(/[^a-z0-9]/gi, '_').toLowerCase()
 }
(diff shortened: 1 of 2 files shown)


umap/static/umap/js/modules/utils.js
```diff
@@ -430,3 +430,15 @@ export function deepEqual(object1, object2) {
 export function slugify(str) {
   return (str || 'data').replace(/[^a-z0-9]/gi, '_').toLowerCase()
 }
+
+export function debounce(callback, wait) {
```
almet, line 434: Coming from https://www.joshwcomeau.com/snippets/javascript/debounce/

yohanboniface:
Replaced by #2415

PR #2240, opened 2024-10-28
chore: introduce SaveManager class

yohanboniface:
The last refactor of the save process (0fdb70ce669838f9fd66eb5de570ffde78e3653d) has introduced a bug: the save flow was not more stopped when a layer fail to save. This ends with a broken UI, as the map is not dirty anymore, so the save button is not active anymore, while at least one layer still needs to be saved.

Save can fail in two scenarios:
  - there is a conflict (status 412)
  - the server is down or as an issue (eg. disk is full)

I tried a more modest scenario (listening for status in map.save and recallign the map save after a conflict), but this ended with calling map.save uselessly even more.

So I decided to try this refactor, which also totally remove the useless map.save we were sending at each save until now.

### DIFF
--- a/umap/static/umap/js/modules/data/layer.js
+++ b/umap/static/umap/js/modules/data/layer.js
@@ -20,6 +20,7 @@ import { translate } from '../i18n.js'
 import { DataLayerPermissions } from '../permissions.js'
 import { Point, LineString, Polygon } from './features.js'
 import TableEditor from '../tableeditor.js'
+import { ServerStored } from '../saving.js'
 
 export const LAYER_TYPES = [
   DefaultLayer,
@@ -35,8 +36,9 @@ const LAYER_MAP = LAYER_TYPES.reduce((acc, klass) => {
   return acc
 }, {})
 
-export class DataLayer {
+export class DataLayer extends ServerStored {
   constructor(map, data) {
+    super()
     this.map = map
     this.sync = map.sync_engine.proxy(this)
     this._index = Array()
@@ -58,7 +60,6 @@ export class DataLayer {
       editMode: 'advanced',
     }
 
-    this._isDirty = false
     this._isDeleted = false
     this.setUmapId(data.id)
     this.setOptions(data)
@@ -89,21 +90,14 @@ export class DataLayer {
     if (this.isVisible()) this.propagateShow()
   }
 
-  set isDirty(status) {
-    this._isDirty = status
+  onDirty(status) {
     if (status) {
-      this.map.isDirty = true
       // A layer can be made dirty by indirect action (like dragging layers)
       // we need to have it loaded before saving it.
       if (!this.isLoaded()) this.fetchData()
     } else {
-      this.map.checkDirty()
       this.isDeleted = false
     }
-  }
-
-  get isDirty() {
-    return this._isDirty
   }
 
   set isDeleted(status) {
@@ -582,7 +576,6 @@ export class DataLayer {
 
   reset() {
     if (!this.umap_id) this.erase()
-
     this.resetOptions()
     this.parentPane.appendChild(this.pane)
     if (this._leaflet_events_bk && !this._leaflet_events) {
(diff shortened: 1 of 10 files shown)


umap/static/umap/js/modules/saving.js
```diff
@@ -0,0 +1,54 @@
...
+  delete(obj) {
+    this._queue.delete(obj)
+    this.checkStatus()
+  }
+
+  has(obj) {
+    return this._queue.has(obj)
+  }
+
+  checkStatus() {
```
davidbgk, line 32: I'm not sure I get the difference between `checkStatus` and `isDirty` concepts, can't we merge them both?

umap/static/umap/js/modules/saving.js
```diff
@@ -0,0 +1,54 @@
...
+    document.body.classList.toggle('umap-is-dirty', this._queue.size)
+  }
+}
+
+export const SAVEMANAGER = new SaveManager()
+
+export class ServerStored {
```
davidbgk, line 39: I'm not sure of that pattern, a proxy to a constant makes me think we are missing a JS way of handling it 😅

umap/static/umap/js/modules/data/layer.js
```diff
@@ -1057,8 +1050,9 @@ export class DataLayer {
     const headers = this._reference_version
       ? { 'X-Datalayer-Reference': this._reference_version }
       : {}
-    await this._trySave(saveUrl, headers, formData)
+    const status = await this._trySave(saveUrl, headers, formData)
```
davidbgk, line 1053: `_trySave` doesn't always return a status, is that a problem?

umap/static/umap/js/modules/data/layer.js
```diff
@@ -1071,7 +1065,13 @@ export class DataLayer {
               'This situation is tricky, you have to choose carefully which version is pertinent.'
           ),
           async () => {
+            // Save again this layer
             await this._trySave(url, {}, formData)
```
davidbgk, line 1069: What if there is still an error, shouldn't we check that `true` is returned here?

umap/static/umap/js/modules/saving.js
yohanboniface: Bad naming then I guess :)

`isDirty` is about a given object to need to be saved to the server or no.
`checkStatus` is the SaveManager that update the UI to reflect that there is at least one object to be saved to the server.

umap/static/umap/js/modules/saving.js
yohanboniface: Are you sure we are not using PHP ? :thinking: 

I'll look for a better pattern, but if you have a suggestion, please go… :)

umap/static/umap/js/modules/data/layer.js
yohanboniface: Indeed! Good point :)

umap/static/umap/js/modules/data/layer.js
yohanboniface: Here is how I think it works:
- when a layer conflicts, the whole save flow will stop (as we do not return a truthy status, maybe this should be more explicit?)
- user can click on the alert actions (override/cancel my edits), but AFAIK it's too late to proceed with the save flow (it's already stopped), we have to start it again

But I'll have a better look a it. I thought we cannot `await` for the user to click on the alert, but I did not test it for real, so I'll do!

umap/static/umap/js/modules/saving.js
yohanboniface: Would `updateDOM` be a better naming here (instead of `checkStatus`, which says about `how` but not about `what`) ?

umap/static/umap/js/modules/saving.js
yohanboniface: Note that I had a look at this bible verset, and stopped reading here https://stackoverflow.com/a/59646297 ;)

umap/static/umap/js/modules/data/layer.js
yohanboniface: Fixed!

umap/static/umap/js/modules/saving.js
yohanboniface: Also, one of my goals was to move that responsability out of the main `Map` object, to make it lighter and easier to refactor, but this ends with the needs to another object, so maybe you'd prefer keeping it in the `Map` object (once separated from the `LeafletMap`) ?

davidbgk (approved):

PR #2272, opened 2024-11-18
chore: move editToolBar and captionBar to modules

yohanboniface:
This also tries to rework a bit the "reflow" process, which how to edit the UI when something changes in the data.

The idea is to build the whole HTML, then:
- the `propagate` method tries to change the UI with targetted CSS selector, but anywhere
- while `render` hide/show some elements

This is not the destination, but just another step. We certainly need to refactor the SCHEMA to make the next step (as we want to associate some data change to UI change).


davidbgk (approved):
If you have time for a few more CSS vars, that would be great!

Thanks for `WithTemplate` :)

PR #2294, opened 2024-11-23
chore: attempt to fix CI tests related to `/hot/`

davidbgk:


davidbgk:
We need to decide first if we want to target the `/hot/` theme AND if we want to keep the `.fr` or switch to `.org`.

yohanboniface:
In fact we don't really care about the used, what we should try to do is to fulfill the request with a dump image (or it can even be a path to a local image), so we do not depend on the network anymore. We only want to check that the import has changed the background, after all.

yohanboniface:
fixed in #2317

PR #2310, opened 2024-12-01
fix: compute length of all shapes for MultiLineString (not only first)

yohanboniface:


umap/static/umap/js/modules/rendering/ui.js
```diff
@@ -380,8 +385,16 @@ export const LeafletPolyline = Polyline.extend({
   },
 
   getMeasure: function (shape) {
+    const shapes = shape
```
davidbgk, line 388: That section is not trivial to readerstand

yohanboniface: Agreed! I've changed it :)

Issue #2302, opened 2024-11-26
Improving the import process (wizard)


Aurelie-Jallut:
**Is your feature request related to a problem? Please describe.**
U tests have shown that people don't understand that once they have chosen the perimeter of a commune, for example, they have to continue with the steps in the panel. 

**Describe the solution you'd like**
- Move the link to open the wizard closer to the URL field to show that they are linked.
- Remove the data type selection buttons from the panel to make it easier for people to see the Import button.
- after selecting the data in the modal: Add a luminous outline to the URL field and a picto to guide users 
- Change the style of the button with a deactivated and activated (illuminated) state

[Figma prototype](https://www.figma.com/proto/7ytmJizPCtdncPBD7BeRlj/UI-uMap?node-id=2695-1663&node-type=frame&t=WlLrq4fjTPQHP77T-1&scaling=min-zoom&content-scaling=fixed&page-id=2077%3A1370&starting-point-node-id=2637%3A457)

Aurelie-Jallut:
One argument in favour of leaving a marker is that, for accessibility reasons, you don't want to give information using colour alone.

Perhaps we could replace it with the one used in front of the Assistant link?

![Image](https://github.com/user-attachments/assets/27eb5c9b-c2a1-44db-8463-7600bc64cc2d)

yohanboniface:
This highlight will describe the input state (fulfilled), not the origin of the value (import helper or user), so I'd say we need something generic :)

yohanboniface:
@Aurelie-Jallut can you extract the icon from figma ? I don't know how to do that. Thanks :)

Aurelie-Jallut:
![Image](https://github.com/user-attachments/assets/3c539fb5-a518-4e2a-93d6-2e7dd7431cf5)

PR #2323, opened 2024-12-02
feat: highlight importer URL field when it is fulfilled

yohanboniface:
Very small step of #2302 

image


davidbgk (approved):
Thanks for the distinct commit!

PR #2328, opened 2024-12-03
feat: add a disabled/active mode to the submit button of import panel

yohanboniface:
cf #2302

Warning: I removed the historic `width: 100%` on all buttons, who knows where this will break? :p

Disabled:

image

Form valid:

image


davidbgk:
It might be interesting to use the JavaScript native API for that? https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#validating_forms_using_javascript

The disabled state doesn't look disabled enough (visually) to me. @Aurelie-Jallut some advices/good practices about it?

yohanboniface:
> It might be interesting to use the JavaScript native API for that? https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#validating_forms_using_javascript

I'm not sure, given that in reality what we check here is the "validity" (consistency?) of the whole form, not the validity of each field/value. Maybe a naming issue of the `isValid` method ?

yohanboniface:
Closed in favor of #2341

PR #2340, opened 2024-12-05
fix: Avoid map-panning on mobile using two fingers navigation

fttriquet:
If scrollWheelZoom option is false we deactivate leaflet dragging options only on mobile


yohanboniface:
Thanks!

I'm not sure to understand: while in mobile and scrollWheelZoom if off, this will disabled dragging the map with one finger, but it will still be possible with two fingers ?

yohanboniface:
I guess answer is no (for two fingers scrolling), but let's go like this and see how it goes!

PR #2343, opened 2024-12-06
feat: add elevation gain and loss in extended properties

yohanboniface:
image


davidbgk (commented):
Which tools provide a third altitude coordinate when generating geojson files?

yohanboniface:
Usually any GPX import in uMap will generate a geojson with altitudes (technically some GPX may not have the altitude, but it's rare now I think).

PR #2349, opened 2024-12-09
feat: allow to edit datalayer name in datalayers list

### LINKED ISSUE #1995: Support renaming layers directly in the Manage layers list
Currently to rename a layer I need to open its editing form and change the title. This is a bit cumbersome and requires more steps than would be ideal.

In many layered editing programs (e.g. GIMP, Inkscape) it's possible to simply double-click the layer name in the list of layers and the text changes to edit mode, so that one can edit the title inline. I'd like uMap to also have this functionality.

If it's easier to implement, a Rename icon next to the layer name, that when clicked shows a pop up to edit new layer name, would also be more convenient than the current workflow. But inline editing would of course be preferable.

yohanboniface:
fix #1995

https://github.com/user-attachments/assets/b16d7d1b-eee1-4da2-93b9-2aac127051ba

### DIFF
--- a/umap/static/umap/css/form.css
+++ b/umap/static/umap/css/form.css
@@ -1,3 +1,6 @@
+.umap-form-inline {
+    display: inline;
+}
 input[type="text"], input[type="password"], input[type="date"],
 input[type="datetime-local"], input[type="email"], input[type="number"],
 input[type="search"], input[type="tel"], input[type="time"], input[type="file"],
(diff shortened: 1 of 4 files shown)


davidbgk:
@Aurelie-Jallut WDYT?

davidbgk (approved):

waldyrious:
This is amazing! Thank you for implementing this feature! :heart:

PR #2357, opened 2024-12-10
feat: introduce Map.share_status=DRAFT and DELETED

### LINKED ISSUE #2207: Add a Draft status
Suggestion from @Binnette: add a way to distinguish quality maps from the mass.

One option may be to have a draft status, and each map would first saved in this status, and editors would explicitly "publish" a map.

That could be a new boolean property `draft=true/false`, or a new value for `share_status=draft`, and in this case the map would react as for `share_status=public`: it would not be listed in the home/search, and would not be indexable by bots.

yohanboniface:
This PR introduce two new share_status: `DRAFT` and `DELETED`.

So all status are now:

        (DRAFT, _("Draft (private)")),
        (PUBLIC, _("Everyone (public)")),
        (OPEN, _("Anyone with link")),
        (PRIVATE, _("Editors and team only")),
        (BLOCKED, _("Blocked")),
        (DELETED, _("Deleted")),

Here are the impact of such introduction, on the draft side:
- by default maps are now create in draft status, and they are not visible from others than owner (or collaborators if any); this can be changed for a given instance with the setting `UMAP_DEFAULT_SHARE_STATUS`
- now even anonymous owned maps have a share status, given one of the goals is to make a better distinction between maps ready to be shared and other, this also apply to maps without logged in owners
  image
- when the map in in draft mode, the "Save" button on the frontend says "Save draft", so to make the state clear
  image
- now the visibility of the map is immediately visible on the top bar (while before this was only displayed after first save)
  image

Note: add the end `DRAFT` and `PRIVATE` are very similar, but I made the choice to keep the two, so one can still distinguish their draft maps from their ready map they want to keep private.

On the delete side:
- when deleting a map, it's now set as `share_status=DELETED`, which act as a sort of trash; so it become easier to recover a map, with it's full datalayers, etc. (the only thing which will not be restored is the previous share_status, which should be draft again after a restore; the restore function itself is not implemented)
- there is a new command `empty_trash` which delete for real maps in DELETED status and with last_modified > 30 days (by default, can be changed with a command line argument)
- deleted maps disappear from all views: home, search, dashboard…
- in the future, we could create a new view "My Trash", where one could see their deleted map not yet deleted for real (and that should be the opportunity to also add the `restore` function, which for now can be done by simply changing the share_status from the shell or the admin)
- all the purgatory related code has been removed

fix #2207


umap/management/commands/empty_trash.py
```diff
@@ -0,0 +1,32 @@
+from datetime import datetime, timedelta
+
+from django.core.management.base import BaseCommand
+
+from umap.models import Map
+
+
+class Command(BaseCommand):
+    help = "Remove maps in trash. Eg.: umap purge_deleted --days 7"
```
davidbgk, line 9:
```suggestion
    help = "Remove maps in trash. Eg.: umap empty_trash --days 7"
```

umap/static/umap/js/modules/ui/bar.js
```diff
@@ -31,7 +31,8 @@ const TOP_BAR_TEMPLATE = `
         <button class="edit-save button round" type="button" data-ref="save">
             <i class="icon icon-16 icon-save"></i>
             <i class="icon icon-16 icon-save-disabled"></i>
-            <span class="">${translate('Save')}</span>
+            <span data-ref="saveLabel">${translate('Save')}</span>
+            <span data-ref="saveDraftLabel">${translate('Save draft')}</span>
```
davidbgk, line 35: Maybe putting the `hidden` attribute by default would avoid flashing buttons? Or is that rendered after the `redraw()` anyway?

davidbgk (approved):
Looks great!

umap/static/umap/js/modules/ui/bar.js
yohanboniface: I'm not sure it would blink as `redraw` will always change both `hidden` in the same time with opposite values, but it does not hurt to make both hidden by default, so I've done that!

Issue #564, opened 2018-03-01
Better feedback during the import process


jidanni:
When Importing a .geojson file with 3000 points:
1. No spinner so one can tell when the upload is finished.
1. No documentation on how many points is the max allowed.
2. Maybe not all my points got uploaded... The browser became unresponsive.

jidanni:
Now putting the file on a website, and entering the URL so as to upload from URL... same thing happens: nothing.

Aurelie-Jallut:
I'm opening up the subject of user feedback.
Here are 2 observations linked to 2 user tests
- on the OSM instance: After importing the datamine water point data for a commune: it's slow and buggy
=> no error message

- remote import of geodatamine library data for Lyon / cadastre import (in layer and remotely) => ‘Problem in server response’ error message following import 

=> It's better to have a message even if you don't really understand whether it's temporary, whether you need to relay the problem somewhere or whether it will work later.

yohanboniface:
If I summarize, we need:

- to start the loader when starting the export (and make the loader more visible ? => #148), and stop it at the end
- zoom to the imported data after import (imported data != datalayer data, as it may already have some preexisting data)
- to have a message when the import does not found any data

@Aurelie-Jallut thoughts ? :)

yohanboniface:
A bit of technical considerations:
- what we need here is to be able to plug in the end of a whole data import, after which we could decide:
  - whether to create a datalayer or not
  - where to zoom
  - alert if not data has been found

The fact is that all the import process for now is blind about the imported content, and only the already created datalayer knows about the content, and when it knows it it is already mixed with preexisting content if any. Also  we need to keep in mind that it's possible to import multiple files at once (so it's like multiple import one after the other).

We think that a clean approach would be to totally extract the data management from the datalayer to a dedicated "importer" or whatever class. But this change is quite committing, and given we are about to integrate vector tiles and thus more concepts around layers, it may not be the best moment to refactor this part (yet). Let's keep this in mind when we work on integrating vector tiles.

In the mean time, we could have a temporary solution with either a promise returned by the import chain process (a bit intrusive), an event (but now uMap classes do not know about events, only Leaflet ones do), or maybe a very targeted layer function that would be called after each import whatever the caller of this import (eg. event when using a query string url).

Aurelie-Jallut:
another question : )
What types of error can be detected when importing data?
(the idea would be to see if we could make the messages clearer depending on the problem encountered or see how to avoid them beforehand).

[...]

PR #2363, opened 2024-12-12
feat: refactor importer feedback

yohanboniface:
We basically make the all import chain return the results as promise, so the importer can act at the end of the process and:
- zoom only to the imported data (in case the layer already as some)
- display a counter of imported data when import is successfull
- display an alert when nothing has been imported

cf #564

### DIFF
--- a/umap/static/umap/js/modules/data/layer.js
+++ b/umap/static/umap/js/modules/data/layer.js
@@ -252,10 +252,11 @@ export class DataLayer extends ServerStored {
   }
 
   fromGeoJSON(geojson, sync = true) {
-    this.addData(geojson, sync)
+    const features = this.addData(geojson, sync)
     this._geojson = geojson
     this.onDataLoaded()
     this.dataChanged()
+    return features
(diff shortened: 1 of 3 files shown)


umap/static/umap/js/modules/data/layer.js
```diff
@@ -503,31 +508,47 @@ export class DataLayer extends ServerStored {
...
+      const reader = new FileReader()
+      reader.onloadend = () => resolve(reader.result)
+      reader.readAsText(f)
+    })
+  }
+
+  async importFromFiles(files, type) {
+    let all = []
+    for (const file of files) {
+      const features = await this.importFromFile(file, type)
```
davidbgk, line 531: I wonder if the [`promise.all()` pattern](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) could be in use here 🤔

davidbgk (approved):
Tests are OK 😮

PR #2370, opened 2024-12-16
feat: load all datalayers in parallel

yohanboniface:


davidbgk (approved):
👍 Could it be in use [here too](https://github.com/umap-project/umap/pull/2363#discussion_r1882966275)?

PR #2386, opened 2024-12-19
fix: honour carriage returns in layer description (in caption panel)

yohanboniface:
This could have been a four letters fix (adding class "text"), but I've seen that `Caption` was still using `DomUtil` instead of `WithTemplate`, so I made this little effort more…


davidbgk (approved):
Yay! 🎉

PR #2396, opened 2024-12-23
wip: add a 403.html template

### LINKED ISSUE #2390: Add a "Non authorized" page
When someone loads a map they cannot access to, they have a blank page, without any explanation.

Now that draft is default mode, this situation will happen more often, so I think we need a message in this page. Something like "This map is private" or "You are not authorized to see this map".

@Aurelie-Jallut what do you think ? :)

yohanboniface:
fix #2390

### DIFF
--- a/umap/templates/404.html
+++ b/umap/templates/404.html
@@ -1,19 +1,10 @@
-{% extends "base.html" %}
+{% extends "40x.html" %}
 
 {% load i18n static %}
 
 {% block content %}
-  <div class="content404">
-    <a href="{% url "home" %}"
-       title="{% trans "Take me to the home page" %}">
-      <h1>
-        4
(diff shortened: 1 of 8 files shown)


umap/templates/404.html
```diff
@@ -1,19 +1,10 @@
...
-        4
-        <img alt="0" width="128px" height="128px" src="{% static "umap/img/logo.svg" %}">
-        4
-      </h1>
-      <h2>
-        Not Found
-      </h2>
-    </a>
+  <div class="content-40x content-404">
+    <h1>404 Not Found</h2>
```
davidbgk, line 7: Translation too?

davidbgk:
Capture d’écran, le 2024-12-23 à 10 27 17

Hopefully I didn't miss a case!

PR #2406, opened 2024-12-29
fix(helm): reference secret-env by fullname instead release-name

swarnat:
Related to helm Charts:
Currently the reference to secret is done by
```          
envFrom:
- secretRef:
  name: {{ .Release.Name }}-env
```
[Link](https://github.com/umap-project/umap/blob/master/charts/umap/templates/deployment.yaml#L67C1-L69C44)

but secret object is created like this:
```
metadata:
  name: {{ include "umap.fullname" . }}-env
```
[Link](https://github.com/umap-project/umap/blob/master/charts/umap/templates/secret-env.yaml#L4C3-L4C4)

Works as long as no Sub-Chart feature is used, when individual configuration is stored within a git repo.
Then the subchart name is added to fullname. (And I think if custom fullName is set)

To be safe, `{{ include "umap.fullname" . }}` is correct. 

Should be no breaking change, when it is working at the moment. (Because then both return same value)
It is only a breaking change, if it is not working at the moment.

### DIFF
--- a/charts/umap/templates/deployment.yaml
+++ b/charts/umap/templates/deployment.yaml
@@ -80,7 +80,7 @@ spec:
       volumes:
         - name: config
           secret:
-            secretName: {{ .Release.Name }}-config
+            secretName: {{ include "umap.fullname" . }}-config
         - name: statics
           emptyDir: {}
       {{- if .Values.persistence.enabled }}
--- a/charts/umap/templates/deployment.yaml
+++ b/charts/umap/templates/deployment.yaml
@@ -66,7 +66,7 @@ spec:
           {{- end }}
           envFrom:
           - secretRef:
-              name: {{ .Release.Name }}-env
+              name: {{ include "umap.fullname" . }}-env
           volumeMounts:
             - name: config
               mountPath: /etc/umap/


davidbgk:
@swarnat thank you for your contribution!

There are two other places where `{{ .Release.Name }}` is referenced, do you think it's better to be consistent with a `{{ include "umap.fullname" . }}` call too?

swarnat:
I just validate that. For configMaps yes. 

For the postgis/cnpg cluster I'm not sure, but I think yes, how second level of subchart is handled. Never worked with cnpg
I do not get it running, because the CRD **postgresql.cnpg.io/Cluster** is missing, but expected.
I currently prepare an environment to test it within an empty k8s system and will update the PR.

yohanboniface:
Thanks!

cc @NaPs :)

ohemelaar:
I just reviewed this by running the template with a few different options (release name or in a subchart) and it looks like it works as expected. The secret name generated by the CNPG dependency should match the template. @davidbgk If you don't have a strict versionning policy for now (I see it's still in 0.1.0) I think this can be merged.

davidbgk:
Thank you @swarnat !

PR #2420, opened 2025-01-06
chore: internalize FormBuilder

### LINKED ISSUE #2280: Internalize and modularize FormBuilder
We want to use modules, but not really to spend the energy of migrating this plugin (sparsely used) with retrocompat.

yohanboniface:
fix #2280 

That's a first step, which:
- internalize Formbuilder as a bunch of modules
- use Javascript classes instead of Leaflet ones
- remove dependencies to Leaflet (L.DomUtil…)
- replaces `L.FormBuilder` by `Form` (in theory generic, but not quite) and `U.FormBuilder` by `MutatingForm` (knows about isDirty, `inheritable` and such)

There is much more room for refactor, but let's do it step by step!


umap/static/umap/js/modules/data/layer.js
```diff
@@ -668,10 +667,11 @@ export class DataLayer extends ServerStored {
       ],
     ]
     DomUtil.createTitle(container, translate('Layer properties'), 'icon-layers')
-    let builder = new U.FormBuilder(this, metadataFields, {
-      callback(e) {
+    let builder = new MutatingForm(this, metadataFields, {
+      callback: (helper) => {
+        console.log(helper)
```
davidbgk, line 672: 👀

umap/static/umap/js/modules/form/builder.js
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  getTemplate(helper) {
+    let template
+    if (helper.properties.inheritable) {
+      const extraClassName = helper.get(true) === undefined ? ' undefined' : ''
+      template = `
+        <div class="umap-field-${helper.name} formbox inheritable${extraClassName}">
+          <div class="header" data-ref=header>
+            <a href="#" class="button undefine" data-ref=undefine>${translate('clear')}</a>
```
davidbgk, line 208: a11y: these links should be at least `role="button"` and even better `<button>` elements 😉

umap/static/umap/js/modules/form/fields.js
```diff
@@ -0,0 +1,1366 @@
...
+}
+
+Fields.TernaryChoices = class extends Fields.MultiChoice {
+  getDefault() {
+    return 'null'
+  }
+
+  toJS() {
+    let value = this.value()
+    switch (value) {
```
davidbgk, line 1202: Third time, can we move it to a dedicated util?

umap/static/umap/js/modules/utils.js
```diff
@@ -416,9 +416,11 @@ export function loadTemplate(html) {
 }
 
 export function loadTemplateWithRefs(html) {
-  const element = loadTemplate(html)
+  const template = document.createElement('template')
```
davidbgk, line 419: I don't get the difference with `loadTemplate`?

umap/static/umap/js/modules/utils.js
```diff
@@ -416,9 +416,11 @@ export function loadTemplate(html) {
...
+  template.innerHTML = html
+  const element = template.content.firstElementChild
   const elements = {}
-  for (const node of element.querySelectorAll('[data-ref]')) {
+  for (const node of template.content.querySelectorAll('[data-ref]')) {
```
davidbgk, line 423: Or maybe `loadTemplate` should (also) return the `template.content`?

davidbgk (commented):
This is huge! And it paves the way to improve forms' accessibility 👍 

I wonder if we could replace the callback mechanism, let's discuss that in particular before merging.

umap/static/umap/js/modules/utils.js
yohanboniface: We need to run the querySelectorAll on the content element, so if the root element itself has a `data-ref` it will be taken into account.

umap/static/umap/js/modules/utils.js
yohanboniface: IMHO, that would pollute the API, as we would have to do this each time:

```
const myElement = Utils.loadTemplate('xxx').firstElementChild
```

instead of what we do now:

```
const myElement = Utils.loadTemplate('xxx')
```

yohanboniface:
> I wonder if we could replace the callback mechanism, let's discuss that in particular before merging.

For sure! What kind API do you have in mind ?

davidbgk:
> For sure! What kind API do you have in mind ?

Not sure how we can deal with a Promise or dedicated events here (and maybe that's a bad idea).

umap/static/umap/js/modules/form/builder.js
yohanboniface: Done! :)

umap/static/umap/js/modules/form/fields.js
yohanboniface: Oops, good catch, NullableBoolean was duplicated (I guess we had one in FormBuilder and one in our custom fields, and I merged all in one file without noticing)! So I remove one, so we only have two switch remaining, and they are not exactly the same (handling `null` or not).

umap/static/umap/js/modules/utils.js
```diff
@@ -446,3 +448,176 @@ export function eachElement(selector, callback) {
     callback(el)
   }
 }
+
+export class WithEvents {
```
davidbgk, line 452: Why not reusing the standard `addEventListener`/ `CustomEvent`+`dispatchEvent` while maintaining the `on`/`fire` API if you prefer?

davidbgk (approved):
All 🟢

PR #2442, opened 2025-01-23
feat: move star button to caption

### LINKED ISSUE #2282: Show map stars on the caption panel
And add the "star" button there too ?

The idea is to make the star feature more visible, so people will use it more, and we'll then be able to showcase some "liked" maps.

yohanboniface:
fix #2282

image
Screenshot From 2025-01-23 10-42-23


umap/views.py
```diff
@@ -780,6 +784,9 @@ def is_starred(self):
             return False
         return Star.objects.filter(by=user, map=self.object).exists()
 
+    def stars(self):
+        return Star.objects.filter(map=self.object).count()
```
yohanboniface, line 788: I wonder if we should store this count in the map itself, updating it at each star/unstar, instead of requesting this at each map view.

davidbgk: I'm not sure it's necessary, a SQL count is relatively light.

umap/static/umap/js/modules/caption.js
```diff
@@ -1,14 +1,16 @@
...
 
 const TEMPLATE = `
 <div class="umap-caption">
   <div class="header">
     <i class="icon icon-16 icon-caption icon-block"></i>
     <hgroup>
       <h3><span class="map-name" data-ref="name"></span></h3>
-      <h4 data-ref="author"></h4>
       <h5 class="dates" data-ref="dates"></h5>
+      <h4 data-ref="author"></h4>
```
davidbgk, line 12: ⚠️ A11Y: the order of titles on the page should be consecutive. Here we end up with h3 => h5 => h4.

I also wonder if everything is really part of the title here 🤔

yohanboniface: > ⚠️ A11Y: the order of titles on the page should be consecutive. Here we end up with h3 => h5 => h4.

Sounds like a very weird rule to me, isn't the "x" in "hx" is semantic ?

> I also wonder if everything is really part of the title here 🤔

For me, `heading` is an "en-tête", so "anything that is not the body" could more or less be considered as heading in my comprehension.

davidbgk: It is semantic indeed, that's why it's probably not a title in the first place. You can put a `p` in a [`hgroup`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup) for instance.

For the official recommendation relative to order of titles:
https://www.w3.org/WAI/tutorials/page-structure/headings/#heading-ranks

yohanboniface: humm, ok! This is what I used to know, and this was making more sense to me: https://web.archive.org/web/20211231200635/https://html.spec.whatwg.org/multipage/sections.html#the-hgroup-element

Using a `p` or any generic tags for subheadings seems semantically poor to me, but ok, let's follow that rules, I'll update the PR.

yohanboniface: Done!

It's quite ugly as the button and the p does not have the same font-size, and I don't why, and I don't have the energy for magic CSS…

image

(Also, I'm a bit sad each time I use a `p` for anything else than text, but I can live with that.)

davidbgk (approved):

Aurelie-Jallut:
Can you reduce the space inside the button? and increase the space between the text above and the button?

What I see :

image

What I would lik :)

image

PR #2444, opened 2025-01-23
Sync show usernames

### LINKED ISSUE #2267: Sync: show the list of connected peers
Currently, only the number of connected peers is shown. We could also list the connected peers and show their username if we have one.

Attention: it's possible for a user to have multiple sessions, so just showing their names might not be enough.

yohanboniface:
fix #2267

### DIFF
--- a/umap/static/umap/css/bar.css
+++ b/umap/static/umap/css/bar.css
@@ -32,6 +32,11 @@
     background-color: var(--color-lightCyan);
     color: var(--color-dark);
 }
+.dark [type=button].connected-peers:hover
+{
+    text-decoration: none;
+    border: 0.5px solid var(--color-brightCyan);
+}
 .dark .off.connected-peers {
     background-color: var(--color-lightGray);
     color: var(--color-darkGray);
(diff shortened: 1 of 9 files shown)


yohanboniface:
Current rendering:

image

@Aurelie-Jallut here is where we are! We do not put the colored circle, as we do not show the other peers position in the map nor do anything with this color. Should we anyway ?

Aurelie-Jallut:
um, why did we already add a colour? 😬 
 It's probably to do with being able to see where users are on the map. 

Are there any plans to do this?

yohanboniface:
>  It's probably to do with being able to see where users are on the map.

I'd say so!

> Are there any plans to do this?

Not soon at least ;)

Aurelie-Jallut:
Can you remove the hyphen and the underline from the number ? 

image

Aurelie-Jallut:
On hover, a border could be displayed in the same way as for the Undo and Save buttons. 

border: 0.5px solid #42ECE6

Aurelie-Jallut:
we could also : 
- add to the frame a border-radius: 5px;
- reduce the size of the arrow and add a radius of 1px
- balance the top and bottom margins: bottom margin slightly larger than the top margin

in image :)
image

davidbgk:
Capture d’écran, le 2025-01-30 à 14 13 20

@Aurelie-Jallut better? (the radius on the arrow tip is too complex with the current implementation)

davidbgk (approved):

PR #2445, opened 2025-01-24
feat(forms): add a debounce for Input and Textarea fields

### LINKED ISSUE #2415: Sync: add debounce for FormBuilder input
As all the changes made from FormBuilder are immediately synced, we need to have a debounce, to only write after say 500 ms of inactivity or something like that.

fix #2238

yohanboniface:
fix #2415

I'm a bit afraid this will add more hiccup to the playwright tests, tough :(

### DIFF
--- a/umap/static/umap/js/modules/form/fields.js
+++ b/umap/static/umap/js/modules/form/fields.js
@@ -132,7 +132,10 @@ Fields.Textarea = class extends BaseElement {
     super.build()
     this.textarea = this.elements.textarea
     this.fetch()
-    this.textarea.addEventListener('input', () => this.sync())
+    this.textarea.addEventListener(
+      'input',
+      Utils.debounce(() => this.sync(), 300)
+    )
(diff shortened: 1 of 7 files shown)


umap/static/umap/js/modules/form/fields.js
```diff
@@ -212,8 +218,8 @@ Fields.Input = class extends BaseElement {
 }
 
 Fields.BlurInput = class extends Fields.Input {
-  getSyncEvent() {
-    return 'blur'
+  listenForSync() {
+    this.input.addEventListener('blur', () => this.sync())
```
davidbgk, line 222: No debounce for blur inputs?

davidbgk (approved):

umap/static/umap/js/modules/form/fields.js
yohanboniface: I don't think it's needed, given for blur input, the sync is only done once (at… "blur" :) ), while for normal input (and textarea) it was done at each character change.

PR #2455, opened 2025-01-27
a11y: include site description within page titles

davidbgk:
This is not the same as the short `SITE_NAME` which is displayed as the title of the (home)page for instance.

The plan is to set `SITE_DESCRIPTION` as `uMap OpenStreetMap` for the OSM instance and `uMap agents publics` for the ANCT one.

### DIFF
--- a/umap/templates/umap/team_form.html
+++ b/umap/templates/umap/team_form.html
@@ -2,6 +2,10 @@
 
 {% load i18n %}
 
+{% block head_title %}
+  {% translate "Create or edit a team" %} - {{ SITE_DESCRIPTION }}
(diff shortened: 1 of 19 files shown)


umap/templates/umap/team_form.html
```diff
@@ -2,6 +2,10 @@
 
 {% load i18n %}
 
+{% block head_title %}
+  {% translate "Create or edit a teamp" %} - {{ SITE_DESCRIPTION }}
```
yohanboniface, line 6: teamp :p

yohanboniface (approved):

PR #2459, opened 2025-01-29
feat: soft delete datalayers

yohanboniface:
When deleting a datalayer, it will now be moved to a state "deleted", and it will only be deleted for real when running the command `umap empty_trash`.

This is what we already do for the map itself, but until now if a user deleted a only a datalayer by mistake (not the map itself) it could not retrieve it.


davidbgk (approved):
We might use a dedicated Django QuerySet Manager but maybe it's overkill? I'm afraid we forget to use `.datalayers` at some point…

Also, should we initialize `modified_at` with today's date for all datalayers? (Best was 10 years ago but second best… 😉)

yohanboniface:
> We might use a dedicated Django QuerySet Manager but maybe it's overkill? I'm afraid we forget to use .datalayers at some point…

I thought about that, but given it's a reverse relation manager, I find it would add complexity, but it also may be me missing the right way to use a custom manager in reverse relation (we still need to have access to the unfiltered set).

> Also, should we initialize modified_at with today's date for all datalayers? (Best was 10 years ago but second best… 😉)

This is already done by the migration (Django does it for an autonow field, I've checked) :)

PR #2487, opened 2025-02-07
chore: sync save state

yohanboniface:
When a peer save the map, other peers in dirty state should not need to save the map anymore.

That implementation uses the lastKnownHLC as a reference, but it changes all dirty states at once. Another impementation could be to have each object sync its dirty state, but in this case we do not have a HLC per object as reference, and it also creates more messages.

ping @almet if you're around and wanna have a fresh look. :)

### DIFF
--- a/umap/models.py
+++ b/umap/models.py
@@ -523,17 +523,27 @@ class DataLayer(NamedModel):
     def metadata(self, request=None):
         # Retrocompat: minimal settings for maps not saved after settings property
         # has been introduced
-        obj = self.settings or {
-            "name": self.name,
-            "displayOnLoad": self.display_on_load,
-        }
+        metadata = self.settings
+        if not metadata:
+            # Fallback to file for old datalayers.
+            data = json.loads(self.geojson.read().decode())
+            metadata = data.get("_umap_options")
+            if not metadata:
+                metadata = {
+                    "name": self.name,
+                    "displayOnLoad": self.display_on_load,
+                }
+            # Save it to prevent file reading at each map load.
+            self.settings = metadata
+            # Do not update the modified_at.
+            self.save(update_fields=["settings"])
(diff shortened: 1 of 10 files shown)


almet (commented):
Thanks! Two thoughts on the matter:

A/ About the type of the message:

I believe we're facing the same conceptual challenge we've hit in the past with some other properties that are useful to the uMap UI, but which aren't useful otherwise.

 If we have other messages like this and we don't want to merge them with `OperationMessage` (which are also sent to other connected peers), we might want to add a new type for them.
 
If the goal is to broadcast to all other peers but not have this listed in the operations, should we have a new type of message?

But... for this specific case, this is just one part of the story here. These "save" messages might trigger specific behavior, on the server and on the other peers.

When a new version is saved on the server, we might want to do something specific with it, for instance using it as a base for the future reference, rather than storing all the operations indefinitely in memory. e.g. the goal is not only to grey the "save" button for other peers, it might have another meaning for the protocol.

In this case, keeping the specific "MapSavedMessage" would be useful, as we could want to react on it.

yohanboniface:
> In this case, keeping the specific "MapSavedMessage" would be useful, as we could want to react on it.

We more or less made the same "cheminement de pensée" (oops), and arrived at the same conclusion that this "message/event" could be used later.

As you say, we are facing an ambiguity with the map/map concept: are we talking about the map as a whole (like a umap page) or are we talking about the Umap object. So about moving the code dealing with the save manager state, we could move it to an update (I guess the MapUpdater?), but it would still not be very accurate, as this state does not target only the map object, but also the datalayer or permissions. We could also create a dedicated updater, but that sounds a bit overkill.
I think I am not mature enough to make a better design decision here, so I'd keep like this and wait for the next opportunity to better organize/name it. WDYT ?

PR #2493, opened 2025-02-10
Remove docs ci

yohanboniface:


almet:
I don' thave all the context, but wanted to add that the goal of having the docs build in the CI was to find out errors that might occur in there, like missing links, before merging docs-related changes, and also because sometimes the docs are tied to some files in the repo.

Could you explain what's the reason for removal?

yohanboniface:
Well, we actually wanted to do the other commit (run tests event if the target branch is not master), but when having a look at the file we just thought "humm, but why are we running the tests (twice) in the CI ? let's remove" without thinking more than this. :p

Let's add this back! :)

Issue #673, opened 2019-02-25
Line features difficult to select on mobile


TZornG:

yohanboniface:
And not only on mobile, when the line is 1px width even in a browser one can have a hard time clicking on it.

samwilson:
Is there any workaround for this? It's also tricky after you've selected a line, e.g. it'd be great to be able to make the little vertex boxes a bit bigger to make them easier to click on.

yohanboniface:
@samwilson are you on mobile or desktop ?

About the line hard to click, maybe we should increase the default weight ?

PR #2506, opened 2025-02-11
feat: make vertex icons bigger (and round)

yohanboniface:
Before:
Screenshot From 2025-02-11 13-59-40

After:
Screenshot From 2025-02-11 13-59-12

(Size unchanged on mobile)

cf #673


yohanboniface:
On mobile it's already bigger (20px, while it was 8 on desktop, and now it's 12).

PR #2510, opened 2025-02-12
chore: use our contexmenu class for inplace toolbar

yohanboniface:
And remove Leaflet.Toolbar dependency.
This also teach ContextMenu how to display icons instead of text and how to render in horizontal orientation instead of vertical.

Before:
image

After:

image

We've lost the bottom tip in the process, should I add this to the ContextMenu class ?

Also, the automatic border for the focused button is a bit noisy imho, not sure how to deal with that.


yohanboniface:
Let's merge and iterate on this!

PR #2523, opened 2025-02-24
chore: use template for map advanced buttons (and fix icons)

yohanboniface:
Before:

image

After:

image


davidbgk (approved):
We're not far from the end of that tunnel! 👏

PR #2532, opened 2025-02-28
a11y: switch to better contrasted links

davidbgk:
And underline for external links otherwise the new green color is too close to the text one to clearly identify links.

### DIFF
--- a/umap/settings/base.py
+++ b/umap/settings/base.py
@@ -274,7 +274,7 @@ UMAP_MAPS_PER_PAGE = 5
 UMAP_MAPS_PER_SEARCH = 25
 UMAP_MAPS_PER_PAGE_OWNER = 10
 UMAP_SEARCH_CONFIGURATION = "simple"
-UMAP_HELP_URL = "https://wiki.openstreetmap.org/wiki/UMap#Feedback_and_help"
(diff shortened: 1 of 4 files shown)


umap/settings/base.py
```diff
@@ -274,7 +274,7 @@
 UMAP_MAPS_PER_SEARCH = 25
 UMAP_MAPS_PER_PAGE_OWNER = 10
 UMAP_SEARCH_CONFIGURATION = "simple"
-UMAP_HELP_URL = "https://wiki.openstreetmap.org/wiki/UMap#Feedback_and_help"
+UMAP_HELP_URL = "https://umap-project.org/"
```
yohanboniface, line 277: Why not https://discover.umap-project.org/ ?

yohanboniface (approved):
A before/after screenshot would help understand the changes :)

umap/settings/base.py
davidbgk: You're right 👍

davidbgk:
Before:

Capture d’écran, le 2025-03-03 à 09 11 50

After:

Capture d’écran, le 2025-03-03 à 09 11 08

PR #2533, opened 2025-02-28
a11y: turn embeded maps into inert elements

davidbgk:
One of the side-effects is to make the content of the map un-focusable but we also loose the ability to interact with the map itself.


davidbgk:
That one requires discussion, I'm not sure if it's a feature or not to make the map thumbnail fully inert.

Issue #419, opened 2016-12-10
[wishlist] highlight selected search results on map


druxstr:
When you use the "Search a place" functionality and select a search result, the map zooms to the selected place (can be a street, a building, etc.).

But the selected place isn't highlight in any way. When there are many elements, what you're looking for can be pretty hard to find. Especially as the default zoom level for search is 16 and covers a wide area.

For instance, the image below is centered on the "86 rue Saint Maur" in Paris (16/48.8648/2.3781). But with no visual element to find where the street is, finding it is nearly impossible. 

A kind of red dot would help.

![umap-highlightresults](https://cloud.githubusercontent.com/assets/17235663/21073412/96be7dda-bedd-11e6-925c-847accf4165c.png)

svalo:
This would be really useful, which would be the best approach to implement it? I would like to look into this or support somone doing it.

mutipg:
Great idea, there should be a pin to highlight search result! Please, do this :) 👍

Aurelie-Jallut:
a target ?
![Image](https://github.com/user-attachments/assets/879c111c-a01a-4c2e-80b7-9e0bda9260e2)

yohanboniface:
Thanks!

The target itself may be misleading, because of the locate control:

![Image](https://github.com/user-attachments/assets/0ad87d3f-3de6-49d9-a401-d740a1e8929d)

So I'm more on the two arrows myself. What about just one arrow/triangle ?

yohanboniface:
Should this icon appear:
- on click on the "lens" button
- Or on hover 
- Or on click on the result item title (but that would need to change the current UX, as this closes the panel) 
- Or on search results (so for each result)

davidbgk:
Maybe something more like an actual target?

![Image](https://github.com/user-attachments/assets/84ef53f7-297d-44dc-af03-7008a2e1665b)
![Image](https://github.com/user-attachments/assets/01be45c1-f52f-4dcb-9c0b-03de43f03f26)

Aurelie-Jallut:
an arrow in a circle, but as we're looking at the left-hand panel, a little animation of the picto might catch our attention.

![Image](https://github.com/user-attachments/assets/506c5ff4-85f4-46aa-a312-db8ff0ab2ef3)

Aurelie-Jallut:
target 
![Image](https://github.com/user-attachments/assets/7a11a5c5-8005-413b-8dc6-7d682c8d3000)

[...]

PR #2545, opened 2025-03-06
chore: refactor search UX

yohanboniface:
This is just the UI part of #419

image


yohanboniface:
We now also allow to import full geometries from the search results:

Screenshot From 2025-03-06 16-52-54

yohanboniface:
And now with the target on hover:

image

(Not sure how it will work on mobile/touch though)

davidbgk (approved):

PR #2576, opened 2025-03-14
Update the Dockerfile to expose websockets

almet:
The Dockerfile now uses ASGI by default (via uvicorn) and embeds a
redis server that is used for the synchronization.


davidbgk (approved):

Dockerfile
```diff
@@ -1,20 +1,20 @@
...
-FROM python:3.11-slim AS common
+FROM python:3.11-slim AS runtime
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         tini \
-        uwsgi \
         sqlite3 \
         libpq-dev \
+        redis-server \
```
yohanboniface, line 9: So after discussing with Oscar and Jacques, they both prefer to have Redis out of the Dockerfile, just as we did for PostgreSQL (basically one container == one process).

Could you remove it from here or should I ?

almet: Hey, thanks for the feedback. I don't have the time for it right now, so please do not block on me. I will have a look later when I find some time, if it's not done already :-)

PR #2579, opened 2025-03-17
feat: layers selector in bottom bar

yohanboniface:


davidbgk:
Current status:

Capture d’écran, le 2025-03-26 à 14 10 20

PR #2583, opened 2025-03-18
fix: make sure umap.properties.slideshow is defined

yohanboniface:


yohanboniface:
Not sure why it fails while it is said as "widely available" here:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment

Screenshot From 2025-03-19 09-35-07

davidbgk:
We need to bump to[ 2021 here](https://github.com/umap-project/umap/blob/master/.eslintrc.json#L8) (cf https://stackoverflow.com/questions/76092858/how-can-i-get-eslint-to-accept-the-nullish-coalescing-assignment)

davidbgk:
I check and `npx eslint umap/static/umap/js/` doesn't propose anything better with the new baseline, I guess we'll have to figure it out by ourselves 😬

PR #2584, opened 2025-03-19
feat: add titles in the text formatting dialog

### LINKED ISSUE #813: text formatting infobubble is difficult to read

yohanboniface:
image

fix #813

### DIFF
--- a/umap/static/umap/css/dialog.css
+++ b/umap/static/umap/css/dialog.css
@@ -14,6 +14,9 @@
     height: fit-content;
     max-height: 90vh;
 }
+.umap-dialog ul + h4 {
+  margin-top: var(--box-margin);
+}
 :where([data-component="no-dialog"]:not([hidden])) {
   display: block;
   position: relative;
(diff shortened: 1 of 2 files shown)


davidbgk:
Still not convinced by the final size of that help box, I think we should display the bare minimum and redirect to https://discover.umap-project.org/support/faq/#text-formatting but that requires a helper for the langage, let's do that tomorrow.

yohanboniface:
I don't think an help link that open another page is the same as a popup inside the application. Formatting text seems an important point to me. And I think adding those titles really make it more browsable by a human.

PR #2628, opened 2025-04-09
fix: fix text overflow on search results

yohanboniface:
Before:
Screenshot From 2025-04-09 09-41-21

After:
Screenshot From 2025-04-09 09-40-58

(Please don't ask :p )


davidbgk (approved):
(🤣)

Let's say we don't want to wrap the content.

PR #2636, opened 2025-04-11
document that nginx needs to be added in Docker stack to server statics

yohanboniface:


docker/nginx.conf
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    include       mime.types;
+    default_type  application/octet-stream;
+    sendfile        on;
+    keepalive_timeout  65;
+
+    # Server block
+    server {
+        listen       80;
```
davidbgk, line 27: How does Docker handle https/443 port?

Issue #2625, opened 2025-04-09
Polyline is not saved and shown


rschwebel:
I made a map with multiple (long and with many nodes) polylines between markers. When I save it and reload it, or even when looking at the link-to-view or looking at share->gpx-map, one segment of the polyline is missing. 

Reproducing is difficult, as the segment is missing on save+load. I used [https://umap.openstreetmap.fr](https://umap.openstreetmap.fr), running umap 3.0.2.

This is in the original editor:
![Image](https://github.com/user-attachments/assets/4730c549-90a5-4260-83ba-0cf791702b6a)

And this is the link-to-view:
![Image](https://github.com/user-attachments/assets/6b746a72-bce0-4ef4-abb6-45f85b389e13)

System: Debian stable with Firefox 128.9.0esr.

rschwebel:
Here is a full backup. Unfortunately, github doesn't let me add the .umap file, but with .umap.txt it works 🙈:

[umap_backup_island-tour-2024.umap.txt](https://github.com/user-attachments/files/19661459/umap_backup_island-tour-2024.umap.txt)

rschwebel:
If I select the "missing" polyline Hannover-Hamburg and delete it, then save, then go to the tab showing the "view" and reload, the Hannover-Hamburg polyline is still missing, but in addition the Hamburg-Flensburg polyline does also disappear!

yohanboniface:
Thanks for the report, I'm having a look!

yohanboniface:
The backup is (indeed) already missing the polyline. How did you add those lines ? Drawing by hand or importing ? I'm trying to figure out how to reproduce.

yohanboniface:
> If I select the "missing" polyline Hannover-Hamburg and delete it, then save, then go to the tab showing the "view" and reload, the Hannover-Hamburg polyline is still missing, but in addition the Hamburg-Flensburg polyline does also disappear!

Just to make sure I understand: the originally missing path is "Hamburg-Flensburg" ? This is what I understand from your screenshots. So if I'm wrong here, it means I'm missing a point in issue report.

yohanboniface:
Also, if you still have the original page opened, can you check the console logs ?

rschwebel:
> The backup is (indeed) already missing the polyline. How did you add those lines ? Drawing by hand or importing ? I'm trying to figure out how to reproduce.

Drawing by hand.

rschwebel:
[...]

PR #2637, opened 2025-04-11
fix: do not try to remove a feature not yet added

yohanboniface:
cf #2625

This happens while drawing a line or polygon and then hitting escape while the path is not yet valid.

### DIFF
--- a/umap/static/umap/js/modules/data/layer.js
+++ b/umap/static/umap/js/modules/data/layer.js
@@ -418,6 +418,10 @@ export class DataLayer {
 
   removeFeature(feature, sync) {
     const id = stamp(feature)
+    // This feature was not yet added, may be after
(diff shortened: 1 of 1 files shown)


davidbgk (approved):

umap/static/umap/js/modules/data/layer.js
```diff
@@ -418,6 +418,10 @@ export class DataLayer {
 
   removeFeature(feature, sync) {
     const id = stamp(feature)
+    // This feature was not yet added, may be after
+    // hitting Escape while drawing a new line or
+    // polygon, not yet valid (not enough points)
+    if (!this._index[id]) return
```
davidbgk, line 424: Maybe we should remove it from `this._index` completely?

PR #2646, opened 2025-04-14
Adapt chart to match La Suite's expectations

ohemelaar:
- env vars and config files can be loaded from existing secrets
- chart is deployed on a helm repository

to make the helm repository work with github pages, create an empty orphaned branch called gh-pages, and under settings>pages>source select deploy from branch and use gh-pages

### DIFF
--- a/.github/workflows/release-helm.yml
+++ b/.github/workflows/release-helm.yml
@@ -3,7 +3,7 @@ name: Release Charts
 on:
   push:
     branches:
-      - main
+      - master
 
 jobs:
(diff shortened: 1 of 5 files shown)


.github/workflows/release-helm.yml
```diff
@@ -0,0 +1,26 @@
+name: Release Charts
+
+on:
+  push:
+    branches:
+      - main
```
davidbgk, line 6:
```suggestion
      - master
```

PR #2660, opened 2025-04-17
wip: first naive version of map templates

yohanboniface:


davidbgk (approved):
Even the design system! 🫶

PR #2673, opened 2025-04-23
feat: add a back button in rules form

### LINKED ISSUE #2631: Make it easier to add a conditional property
When a conditional property is added, once the form has been completed, the path requires 3 clicks to add a new one : 
Form completed : 
Image

Close the window (1), click on advanced map properties (2), and choose conditional property (3).

If possible, add a button on form 1 and specify ‘Add another conditional property’.

See [Enhancing the conditional rules #2334 ](https://github.com/umap-project/umap/issues/2334)

yohanboniface:
To go back in rules list.

fix #2631

### DIFF
--- a/umap/static/umap/js/modules/umap.js
+++ b/umap/static/umap/js/modules/umap.js
@@ -1181,7 +1181,7 @@ export default class Umap {
     }
     this._advancedActions(container)
 
-    this.editPanel.open({
+    return this.editPanel.open({
       content: container,
       className: 'dark',
       highlight: 'settings',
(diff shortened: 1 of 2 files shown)


umap/static/umap/js/modules/umap.js
```diff
@@ -1181,7 +1181,7 @@ export default class Umap {
     }
     this._advancedActions(container)
 
-    this.editPanel.open({
+    return this.editPanel.open({
```
davidbgk, line 1184: Is it related to that PR?

yohanboniface: Yes, it return the promise so someone calling open() can use this promise to run some code when the panel is opened (in this case we'll open a given fieldset) :)

PR #2733, opened 2025-05-29
feat: add minimal Open Data portals importer

yohanboniface:
https://github.com/user-attachments/assets/2c8a1080-d532-48de-8dfc-849cc0dd5631


yohanboniface:
I'm tempted to call this importer "Open Data" or "Open Data Portal", as I think users do not know which plateform is used by which city/country, and so to add the underlying plateform in the list:

```python
UMAP_IMPORTERS = {
    "opendatasoft": {
        "choices": [
            {
                "name": "Aix-Marseille Métropole",
                "url": 'https://data.ampmetropole.fr' , 
                "plateform": 'opendatasoft'
            },
         }
     ]
}
```

PR #2737, opened 2025-05-30
wip: add a "My Templates" page

yohanboniface:
Just to draft so to have your views @davidbgk 

Basically, `user_dashboard.html` template has become `user_maps.html`, and is extended by `user_dashboard.html` (new) and `user_templates.html`.

I thought about doing an include instead, but given that the template defines two blocks (`main_content` and `bottom_js`), I'm not sure how to do that.

But I feel like doing a hack in extending a template with variables…


davidbgk:
Superseeded by #2742

PR #2740, opened 2025-06-02
feat: add management command to purge old versions

yohanboniface:
Old versions are usually deleted on save of a datalayer, but we still keep UMAP_KEEP_VERSION. Let's say after some time even those versions could be deleted, to save some space.


davidbgk (approved):
I would print the number of actually deleted versions at the end of the loop (and if it was dry!).

PR #2771, opened 2025-06-17
Fields management

yohanboniface:
TODO:
- [x] compute fields on the fly for datalayer created prior to fields management
- [x] set default fields when no field has been set (name/description)
- [ ] allow to manage "global" fields from map itself
- [x] adapt fields autocompletes
- [ ] fields with remote data ?
- [x] refresh the UI when fields have changed

### DIFF
--- a/umap/static/umap/css/icon.css
+++ b/umap/static/umap/css/icon.css
@@ -7,229 +7,298 @@
     border: none;
     background-color: initial;
 }
+
 .icon-16 {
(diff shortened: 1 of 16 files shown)


yohanboniface:
>  allow to manage "global" fields from map itself

Let's do this in another iteration!

yohanboniface:
>  fields with remote data ?

For now we hide the "manage field" form

PR #2794, opened 2025-06-30
fix: update brewer color palettes when layer properties change

yohanboniface:
The color palettes we display depend on the amount of classes, thus the number of categories, thus the number of different values on the selected column of the data.

So when selecting another column, we need to update the color palettes list.

The fix itself consists in allowing the `selectOptions` property to be a callable, so be dynamically updated each time we build the options (on fetch call).

### DIFF
--- a/umap/static/umap/js/modules/form/fields.js
+++ b/umap/static/umap/js/modules/form/fields.js
@@ -364,6 +364,9 @@ Fields.Select = class extends BaseElement {
   }
 
   getOptions() {
+    if (this.properties.getOptions) {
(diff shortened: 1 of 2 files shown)


umap/static/umap/js/modules/form/fields.js
```diff
@@ -364,6 +364,9 @@ Fields.Select = class extends BaseElement {
   }
 
   getOptions() {
+    if (this.properties.selectOptions instanceof Function) {
```
davidbgk, line 367: I don't really like the smell of it, let's talk about it.

PR #2800, opened 2025-07-01
wip: add markers in datalayer's pane instead of in their own dedicated pane

yohanboniface:
By default, Leaflet has two panes:
- one for markers
- one for paths (polygon and polylines), below the markers one

In uMap, we have made that each datalayer create its own pane, so to follow the datalayer order, but we only added polyline and polygons in that pane. Markers were still added to their pane, above datalayers panes. In this case, the orders or markers between them was "random", in the sense that even if the datalayers are still created in the right order, the markers are added in the order the ajax requests return.

This proposal changes that, and markers are now added in the same pane as polygons and polylines.

This will make that markers will now follow the datalayer order.

But the downside is that, before, it was impossible for a polygon or path to cover a marker, which will now be possible.

So maybe we want to be more ambitious, and instead of puting paths and markers in same pane, to make that datalayers will create two panes, one for paths (as currently) and one for markers, which would always be above all paths dedicated panes.

Before:

Screenshot From 2025-07-01 11-45-49

After:

Screenshot From 2025-07-01 11-46-13

### DIFF
--- a/umap/static/umap/js/modules/data/features.js
+++ b/umap/static/umap/js/modules/data/features.js
@@ -374,7 +374,7 @@ class Feature {
     this.datalayer = datalayer
     // FIXME should be in layer/ui
     this.ui.options.renderer = this.datalayer.renderer
-    this.ui.options.pane = this.datalayer.pane
+    this.ui.options.pane = this.datalayer.markersPane
   }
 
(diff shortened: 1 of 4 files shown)


yohanboniface:
Replaced by #2812

PR #2812, opened 2025-07-09
feat: use DataLayer order to set marker z-index

yohanboniface:
We use to respect the datalayers order for path but not for markers, which can be frustrating sometimes, now it's the case.


davidbgk (approved):
🔥

PR #2819, opened 2025-07-10
fix: paginate open data results

yohanboniface:
Results were limited to 100 currently.
In the same time, order them by title.


davidbgk (commented):
Maybe we should put a hard limit ourselves? I mean, a dropdown of thousands of entries is barely usable 😬

yohanboniface:
Makes sense!

PR #2831, opened 2025-07-21
feat: open edit buttons on right click instead of left click

yohanboniface:
We want to keep the "click experience" (!) the same whether the user is in view mode or in edit mode.

This is going to be a huge UX change for users, so let's prepare for complaints! :p 

Screenshot From 2025-07-21 17-01-36


umap/static/umap/js/modules/data/features.js
```diff
@@ -637,35 +633,32 @@ class Feature {
...
-        label: `${translate('Edit this feature')} (⇧+Click)`,
-        action: () => this.edit(),
-      })
-    }
-    items = items.concat(
+    return [
       {
-        label: this._umap.help.displayLabel('EDIT_FEATURE_LAYER'),
+        action: () => this.toggleEditing(),
+        title: translate('Toggle edit mode (⇧+Click)'),
```
davidbgk, line 643: I don't think that's the right title for that icon. We are already in edit mode and the keyboard shortcut is not really useful at that point:
Capture d’écran, le 2025-07-22 à 09 56 54

I would just let “Edit” (or the more complex “Edit this …”)

umap/static/umap/js/modules/data/features.js
```diff
@@ -637,35 +633,32 @@ class Feature {
...
+        icon: 'icon-edit',
+      },
+      {
+        title: this._umap.help.displayLabel('EDIT_FEATURE_LAYER', false),
+        icon: 'icon-layers',
         action: () => this.datalayer.edit(),
       },
       {
-        label: translate('Delete this feature'),
+        title: translate('Delete this feature'),
```
davidbgk, line 652: Always keep the delete icon at the end? /cc @Aurelie-Jallut

davidbgk:
Overall, icons are quite small clicking targets, I think we are below recommended sizes for accessibility. Should we bump these to 24px ones @Aurelie-Jallut ?

Issue #2330, opened 2024-12-03
Get rid of Leaflet.Toolbar


yohanboniface:
It's unmaintained, and we want to remove the dependency on Leaflet, and we want to use modules.

Issue #1885, opened 2024-06-06
PICTURE in POI : let display a bigger picture for specific POI (outside the map)


JLZIMMERMANN:
**IDEA :**
In some maps several contextual informations are needed, this is why we may add some picture outside the map to let understand who can be involved (or any other kind of explanation).
![image](https://github.com/umap-project/umap/assets/4929752/6a2e7a85-5832-476e-b2af-7c6398321651)
https://umap.incubateur.anct.gouv.fr/fr/map/sare-en-vaucluse-2024_410#10/44.0585/5.1787

**SOLUTION :**
When a POI is used with such choice an option would be welcome to let add like in the left panel 
{{https://toulourenc-horizons.org/wp-content/uploads/2019/12/logo-chambre-metiers-artisanat-paca.jpg.webp|200}}
but this option is in the POI : **bigget text** ( https://github.com/umap-project/umap/issues/1884 ) and **bigger image**

yohanboniface:
This needs custom icons (#125) and changing icon size (#859)

Aurelie-Jallut:
Here's some feedback from a user: 
‘For example our bus icons are too small in the marker or we can't find the right marker to display them. the default marker shape is the one that shows our icon but we don't think it's great and the bus icon is too small.’

![Image](https://github.com/user-attachments/assets/dd4eafe0-2df0-4bab-83a8-35b4139178ff)

AZWCI:
And resizable beyond 64 pixels ? Until the choice of the desired number of pixels by the fine "uMap goldsmiths" ?
Et redimensionnable au-delà de 64 pixels ? Jusqu’àu choix du nombre de pixels souhaité par "les orfèvres d’uMap" ?

Issue #859, opened 2020-12-08
Modify size icon shape?


robibrazze:
Hi,
is it possible modify the size the "icon shape"?
note: i'm triying whit umap install in local.

Thanks a lot
Roberto

sylvain-m:
This does not seem possible at the moment. I confirm the interest of this feature.

mary-loulou:
Would be great.

yohanboniface:
Related to #125

yohanboniface:
If it's a custom instance, it's possible to have a [custom css](https://umap-project.readthedocs.io/en/master/custom/) to increase the size

PR #2846, opened 2025-08-01
feat: introduce LargeCircle icon class and make it resizable

yohanboniface:
cf #2330
cf #1885
cf #859 

https://github.com/user-attachments/assets/4219fe45-add5-4359-9c7f-24a92d37edd6

### DIFF
--- a/umap/static/umap/js/modules/data/features.js
+++ b/umap/static/umap/js/modules/data/features.js
@@ -736,6 +736,7 @@ export class Point extends Feature {
     return [
       'properties._umap_options.color',
       'properties._umap_options.iconClass',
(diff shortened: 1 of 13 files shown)


ttooff:
Hi @yohanboniface, this is a great new feature, especially for node networks where numbers must be displayed in circles.

Would it be possible to ensure the text is centered horizontally and vertically in the circle?

And to allow changing the background color of the circle (the color parameter only seems to impact the circle edge).

Many thanks!

Christophe.

clwacker:
+1 yes. The LargeCircle is a very nice addition. And yes please add centered text and background color for the full circle. Thanks !

Issue #2675, opened 2025-04-23
Cluster points inside pre-defined polygons


hellpe:
**Is your feature request related to a problem? Please describe.**
Currently in uMap, setting a layer as "Cluster" group its points into dynamic clusters. However, I'd like to see how many points are in a subdivision 

**Describe the solution you'd like**
When zooming out, I'd like to see points grouped inside polygons from another layer.

**Additional context**
See current behavior vs. mockup:

<img width="1008" alt="Image" src="https://github.com/user-attachments/assets/0d942a74-b760-48bb-9e70-c41340d41e2c" /> 

![Image](https://github.com/user-attachments/assets/5c77f3e4-d747-481a-be98-d4a9174cf18a)

Issue #508, opened 2017-08-19
Cluster with multiple layers in umap


AdrianG19:
Hi! My problem is that I use cluster on each layers of my map, but when 2 markers from different layers are close together, they don't gather, they still remains separate. Is there a way to create a "general cluster" for all the layers?
Thanks

ajRiverav:
I am thinking: could you instead put the data in the same layer but use two different icons for the two subsets?

TeunvdWeij:
@ajRiverav what do you mean? How would this work? How do you still get different icons for each subgroup?

loeti82:
May I add a use-case for this feature request.
We as a volunteer fire brigade maintain a map which provides support during incidents. e.g. do we have internal incidents plan, is there a defibrillator around. 
Those things exist in separated layers, as they do have different stakeholders and maintenance- and life-cycles. 
To avoid loosing time during an incident, the location needs to be precise, and therefore it is possible that items on different layers share the same coordinates. e.g. incident plans and fire alarm control panel are at the same coordinates.
![umap overlapping layers](https://github.com/umap-project/umap/assets/8694371/9e9a5486-63c1-4515-8888-9231ec6c1d2c)
For this use case, it would be helpful to cluster among two or more layers and not only with in one layer.

Issue #115, opened 2014-09-17
Make possible to display custom computation on clusters


yohanboniface:
Originally reported by: **Mario Frasca (BitBucket: [mfrasca](http://bitbucket.org/mfrasca), GitHub: [mfrasca](http://github.com/mfrasca))**

---

I'm new to umap so I might have missed the obvious...

I need to help monitor the presence of a virus infection on an area (chikungunya in the Caribbean): selected users would insert cases at 'house' level, so in a house we could have 4 cases and I would like case-markers to have a 'count' that serves as weight for the marker.

the central point is that it would be very practical to have a heatmap layer that uses the same data as the default or clustered one.

I also suppose we want to cluster cases, but then we need take the individual weights in account and show those sums, not just the count of the markers.

I'm afraid this is several issues in one...
- couple heatmap "presentation" layer to other "data" layer
- weight property with clustering layers
- heatmap uses weight property

---
- Bitbucket: https://bitbucket.org/yohanboniface/umap/issue/115

PR #2851, opened 2025-08-04
feat: use our custom cluster layer

### LINKED ISSUE #326: Clustered layer: markers should be at center of cluster
I am using clustered layers at http://umap.openstreetmap.fr/en/map/go-clubs_76310 , but the Canadian markers look like they're in the USA, because the clustered map puts the marker at the bottom right of a cluster of markers. The map would be easier to read if they were at the center of gravity of a cluster.

yohanboniface:
Overriding MarkerCluster is fragile, and we've made quite a lot of monkeypatches.
Plus we want to prepare the possibility to cluster semantically, not only geographicaly. And when we'll have layer groups, to cluster a whole group (so combining amounts and colors).

This clustering algorithm is very very naive, but uMap is not meant to be used with millions of points, so it may be enough for our usages. Let's see and improve later if needed.

Still needed:
- [ ] playwright tests
- [x] spiderify function when reaching max zoom and some markers are still on the same cluster (eg. same exact position)
- [ ] allow to cluster on a property (eg. a region name)
- [ ] allow to define a max zoom for clustering (eg. when zooming over the "region zoom", let's uncluster)
- [ ] allow to display a custom aggregation instead of length in the cluster marker (eg. display the total amount of money spent)

fix #326
cf #115
cf #508
cf #2675


yohanboniface:
Let's do remaining tasks on dedicated PRs.

PR #2861, opened 2025-08-08
feat: adding a Route template with elevation chart

davidbgk:


umap/static/umap/js/modules/rendering/template.js
```diff
@@ -282,3 +289,53 @@ class Wikipedia extends PopupTemplate {
     return body
   }
 }
+
+class Route extends PopupTemplate {
+  async renderBody(feature) {
+    let prev
+    let dist = 0
+    const data = []
+    const latlngs = feature.ui.getLatLngs()
```
yohanboniface, line 298: We should be defensive if the feature is not a LineString here.

umap/static/umap/js/modules/rendering/template.js
```diff
@@ -282,3 +289,53 @@ class Wikipedia extends PopupTemplate {
...
+      if (prev) {
+        dist = map.distance(latlng, prev)
+      }
+      data.push([latlng.alt, dist])
+      prev = latlng
+    }
+    const [chart, { altitude }] = Utils.loadTemplateWithRefs(`
+      <elevation-chart data-elevation="${JSON.stringify(data)}">
+        <p>
+          Altitude : <span data-ref="altitude">0</span> meters
```
yohanboniface, line 313: Translate altitude, plus add length and total gain and loss.

umap/templates/umap/js.html
```diff
@@ -36,6 +36,8 @@
         defer></script>
 <script src="{% static 'umap/vendors/simple-statistics/simple-statistics.min.js' %}"
         defer></script>
+<script src="{% static 'umap/vendors/simple-elevation-chart/elevation-chart.js' %}"
```
yohanboniface, line 39: To be pushed on npm :)

yohanboniface (commented):
:tada:

umap/static/umap/js/modules/rendering/template.js
```diff
@@ -282,3 +289,66 @@ class Wikipedia extends PopupTemplate {
...
+      if (prev) {
+        dist = map.distance(latlng, prev)
+      }
+      data.push([latlng.alt, dist])
+      prev = latlng
+    }
+    const [root, { altitude, chart }] = Utils.loadTemplateWithRefs(`
+      <div>
+        <p>
+          ${translate('Distance:')} ${properties.measure} •
```
yohanboniface, line 317: Maybe we should put ":" outside of the translated string, so to better factorise we other translated strings ?

PR #2866, opened 2025-08-12
feat: add control to print the map and allow to download as image

### LINKED ISSUE #88: Add a Print function

yohanboniface:
fix #88


davidbgk:
That's great! I wonder if we should put a message when exporting images given that it can take some time to generate and there is no visual feedback that there is actually something working in the background.

It can be tricky to get the exact contours of the map where you want it but with a bit of try and adjust it ends up rendering the right part :)

PR #2875, opened 2025-08-25
fix: never trust old id in import, even from _storage

yohanboniface:
cf #2847


umap/static/umap/js/modules/umap.js
```diff
@@ -1609,8 +1609,8 @@ export default class Umap {
     for (const geojson of importedData.layers) {
       if (!geojson._umap_options && geojson._storage) {
         geojson._umap_options = geojson._storage
-        delete geojson._storage
       }
+      delete geojson._storage
```
yohanboniface, line 1613: that's the real fix (it was not deleted in case _umap_options also existed

PR #2903, opened 2025-09-03
Minimum cluster value

s-r-g-i-t-h-u-b:
We need this slider start sooner

### DIFF
--- a/umap/static/umap/js/modules/rendering/layers/cluster.js
+++ b/umap/static/umap/js/modules/rendering/layers/cluster.js
@@ -259,7 +259,7 @@ export const Cluster = FeatureGroup.extend({
       'properties.cluster.radius',
       {
         handler: 'Range',
-        min: 40,
+        min: 10,
         max: 200,
         step: 10,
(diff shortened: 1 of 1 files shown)


davidbgk:
@s-r-g-i-t-h-u-b do you have a public example map where it's pertinent? Thanks!

s-r-g-i-t-h-u-b:
With modified modified cluster.radius values in umap-import for some layers. Used values are like 15 or 20.

https://kinderaufsrad.org/aktionsbuendnis/#aktionsorte ( umap iframe is https://umap.openstreetmap.fr/de/map/aktionswochen-strassenfuralle-herbst-2025_1274455?scaleControl=false&miniMap=false&scrollWheelZoom=false&zoomControl=true&editMode=disabled&moreControl=true&searchControl=null&tilelayersControl=null&embedControl=null&datalayersControl=true&onLoadPanel=none&captionBar=false&captionMenus=true#6/51.413/9.811 )

yohanboniface:
Thanks! :+1:

PR #2921, opened 2025-09-09
fix: make the edition header responsive

davidbgk:


umap/static/umap/map.css
```diff
@@ -1205,12 +1205,8 @@ ul.photon-autocomplete {
 
 @media all and (max-width: 480px) {
 
-    .umap-main-edit-toolbox .share-status {
```
davidbgk, line 1208: These two declarations were moved to `bar.css` for consistency.

davidbgk:
Before:
Capture d’écran, le 2025-09-09 à 10 09 43


After:
Capture d’écran, le 2025-09-09 à 10 09 55

PR #2923, opened 2025-09-09
fix: more responsive sidebars on small screens

davidbgk:


davidbgk:
Before:
Capture d’écran, le 2025-09-09 à 11 16 12

After:
Capture d’écran, le 2025-09-09 à 11 16 25

PR #2962, opened 2025-09-30
feat: add export in CSV with WKT geometry

yohanboniface:
This allow to keep the complex geometries (polygon, polyline), and may also allow to reimport the data in Grist so to edit them there and add them in remote data.


davidbgk (approved):
~Missing betterknown dep?~

PR #2966, opened 2025-10-01
fix: fix star button in caption panel

yohanboniface:


yohanboniface:
Before:
Screenshot From 2025-10-01 09-47-36

After:
Screenshot From 2025-10-01 09-47-53

davidbgk (approved):

PR #2970, opened 2025-10-02
chore: move toolbox to the right

yohanboniface:
The idea is not to mix the name part and the toolbox part, so when we need to have icons on the name part it does not clash.

Also, when toolbox appears on edit mode, turning edit on/off will be less flicky.


yohanboniface:
image

image


cc @Aurelie-Jallut

PR #2977, opened 2025-10-02
feat: display image tag when present in OSM template

yohanboniface:
cf https://forum.openstreetmap.fr/t/umap-valoriser-relier-vers-les-donnees-wikidata/37615/8

### DIFF
--- a/umap/static/umap/js/modules/rendering/template.js
+++ b/umap/static/umap/js/modules/rendering/template.js
@@ -231,6 +231,11 @@ class OSM extends PopupTemplate {
         )
       )
     }
+    if (props.image) {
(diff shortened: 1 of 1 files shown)


umap/static/umap/js/modules/rendering/template.js
```diff
@@ -231,6 +231,9 @@ class OSM extends PopupTemplate {
         )
       )
     }
+    if (props.image) {
+      body.appendChild(Utils.loadTemplate(`<div><img src="${props.image}" /></div>`))
```
davidbgk, line 235: Do we have access to an `alt` props?

PR #2983, opened 2025-10-08
fix: only show Wikipedia intro, and add credit and full link in footer

### LINKED ISSUE #2888: Credentials missing for Wikipedia popup content
I just discovered the [Wikipedia popup content style](https://github.com/umap-project/umap/pull/2365), which is great. However the credentials are missing – Wikipedia contents are under the CC-by-sa licence. IMHO the credentials should be displayed on the popup itself, as well as the Credentials section of the Legend/Caption panel.

### LINKED ISSUE #2889: Wikipedia popup content : only insert the article summary if there is one
I just discovered the [Wikipedia popup content style](https://github.com/umap-project/umap/pull/2365), which is great. However integrating the full article doesn't seem appropriate when it is lengthy. 

Perhaps the article summary when there is one (the section at the top of eg. https://fr.wikipedia.org/wiki/Ch%C3%A2teau_des_ducs_de_Bretagne) together with a picture and a **link to the full article** on Wikipedia, would more be suitable.

**Additional context**
There seems to be a way to extract the article summary : https://stackoverflow.com/questions/61188921/how-to-get-table-info-and-summary-of-page-using-wikipedia-api

yohanboniface:
fix #2888
fix #2889


davidbgk (approved):

yohanboniface:
image

PR #3071, opened 2025-11-07
Update logout redirect URL to use settings

fabriziopicconi:
logout redirect url grabbed from settings

### DIFF
--- a/umap/views.py
+++ b/umap/views.py
@@ -1484,7 +1484,7 @@ def webmanifest(request):
 
 def logout(request):
     do_logout(request)
-    return HttpResponseRedirect(getattr(settings, "LOGOUT_REDIRECT_URL", "/"))
+    return HttpResponseRedirect(settings.LOGOUT_REDIRECT_URL)
 
 
 class LoginPopupEnd(TemplateView):


fabriziopicconi:
I've discussed this change by chatgpt and it agrees solution  ;-)

umap/views.py
```diff
@@ -1484,7 +1484,7 @@ def webmanifest(request):
 
 def logout(request):
     do_logout(request)
-    return HttpResponseRedirect("/")
+    return HttpResponseRedirect(getattr(settings, "LOGOUT_REDIRECT_URL", "/"))
```
davidbgk: I would be more inclined to define a default value (`"/"`) for `LOGOUT_REDIRECT_URL` [in base.py](https://github.com/umap-project/umap/blob/d147fb114b35faa74e7151c5e3e4bf7f883d4627/umap/settings/base.py#L163) and then use directly the setting here:

```suggestion
    return HttpResponseRedirect(settings.LOGOUT_REDIRECT_URL)
```

davidbgk:
Thank you @fabriziopicconi !

PR #3085, opened 2025-11-14
feat: add osm_type and osm_id properties for OSM data

### LINKED ISSUE #3072: uMap converts "{id}" to "{type}/{id}" format in popups (e.g., "node/13203221")
When using the placeholder {id} in uMap popup templates, the value is returned as a combined string of type and ID in the format "node/13203221" instead of just the numeric ID "13203221". This causes problems when constructing URLs or performing operations that require only the numeric ID.

**To reproduce:**

- Open the map at https://umap.openstreetmap.de/de/map/bug-demo_113404
- Edit a layer popup template to include the {id} placeholder, e.g. `Feature ID: {id}`
- Save and open any marker popup
- Notice the {id} value is formatted as "node/13203221" instead of just "13203221"
- This breaks direct editor links or any logic expecting a numeric-only ID

Image

**Expected behavior:**
The {id} placeholder should return the numeric OSM ID only, e.g., "13203221", without the type prefix (node/, way/, or relation/).

yohanboniface:
The data is parsed by osmtogeojson, which creates an "id" property consisting in "osm_type/osm_id", which can limit the ability to use the id itself in some situations (eg. when calling osm.org/edit?node=xxxx).

fix #3072

### DIFF
--- a/docs-users/fr/support/faq.md
+++ b/docs-users/fr/support/faq.md
@@ -94,6 +94,7 @@ Toute propriété de l'élément sera disponible, ainsi que:
 - `{layer}` → le nom du calque de l'élément
 - `{zoom}` → le zoom actuel de la carte
 - `{id}` → identifiant unique d'un élément
+- `{osm_type}`, `{osm_id}` → type (node, way, relation) et identifiant unique d'un élément, quand les données viennent d'OpenStreetMap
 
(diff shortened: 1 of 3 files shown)


new-issue:
Thanx for the quick fix!

Issue #2778, opened 2025-06-19
See which drawing tool is activated


Aurelie-Jallut:
**Is your feature request related to a problem? Please describe.**
I can't see which tool is selected. (i'm blind ;) For example, if I do something like import data and then want to select one of the imported markers, I find that it starts drawing and I can't easily figure out how to get back to point selection mode.

**Describe the solution you'd like**
- Visually identify which drawing tool is selected
- A selection button would be helpful

**Additional context**
The style of the drawing buttons should be differentiated from the others (as we can have an open panel + a selected drawing tool).

Aurelie-Jallut:
<img width="390" height="1094" alt="Image" src="https://github.com/user-attachments/assets/0e0851ee-0e9e-439b-baf6-6ce0f083de13" />

The color is : color-darkCyan (#43A39F)

We can keep the way hover works actually and the light grey for the others buttons when they are activated

PR #3100, opened 2025-11-21
feat: highlight drawing tools on click

yohanboniface:
cf #2778 

cc @Aurelie-Jallut :)

https://github.com/user-attachments/assets/b61ef771-e772-401e-bdef-4871d87b90e5


Aurelie-Jallut:
thanks for that @yohanboniface @davidbgk 

Could we:
- when hovering over the button, display the color green
- when clicking: make the button change color immediately and not when I move away from the button :)
- when the button is activated and I hover over it, it doesn't change appearance, or else it changes to a slightly lighter shade (I'll find you the reference)

- When I change drawing tools, it should keep the same behavior: I tested it on osm fr and it only works on the first tool selection for me.

PR #3124, opened 2025-12-18
fix: do not fail when classified colors are named colors

yohanboniface:
We were using named color as if they were hex color.


davidbgk (approved):
🥸

PR #3187, opened 2026-02-06
Fixed the environment block and added docker.io for podman support

DiabeticTurtle:
Ive changed a little bit in the docker-compose to fix the block errors and updated to the latest umap version also added docker.io to streamline podman.io builds

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ services:
 
   # Useful only to use the real time collaboration
   redis:
-    image: redis:latest
+    image: docker.io/redis:latest
     healthcheck:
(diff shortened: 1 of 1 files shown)


yohanboniface:
Thanks! Can you explain the docker.io prefix change ? :)

DiabeticTurtle:
Yeah! Prolly not the biggest deal, but it fixed issues i had with podman-compose finding the right packages. I have updated code, that doesnt use docker.io/pkg. It really just saves like 3 seconds so podman doesnt have to rely on the default registry if its not set. 😅

yohanboniface:
Thanks! :+1:

PR #3220, opened 2026-02-24
More yunohost integration

yohanboniface:
- integrate core settings for Yunohost
- properly deal with LOGIN_URL and LOGOUT_REDIRECT_URL


umap/settings/base.py
```diff
@@ -158,9 +158,9 @@
 ROOT_URLCONF = "umap.urls"
 WSGI_APPLICATION = "umap.wsgi.application"
 
-LOGIN_URL = "/login/"
-LOGOUT_URL = "/logout/"
+LOGIN_URL = "login"
 LOGIN_REDIRECT_URL = "login_popup_end"
```
fflorent: Maybe this is due to this config that the redirection goes wrong.

PR #3270, opened 2026-04-03
Add docker CI

tux2000:
## Summary
This PR adds a GitHub Actions workflow to automatically build and publish Docker images to Docker Hub whenever a new release is published.

## Motivation
Currently, Docker images are apparently published manually and may lag behind official releases. This automation ensures Docker images stay in sync with tagged releases.

## Details
- Trigger: release.published
- Tags Docker images using the release version
- Pushes to Docker Hub (umap/umap)
- Also publishes to GitHub Container Registry (ghcr.io) using the built-in GITHUB_TOKEN (no extra secrets required)

## Requirements
Repository secrets must be configured:
- DOCKERHUB_USERNAME
- DOCKERHUB_TOKEN

## Optional
We can also publish a `latest` tag if desired.

## Note
I could not test this thoroughly to make sure it actually works to push to docker hub. Let me know if I should revise or test in some specific way.

### DIFF
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -1,0 +1,44 @@
+name: Build and Push Docker Image
+
+on:
+  release:
+    types: [published]
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Extract version
+        id: vars
+        run: echo "version=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
+
+      - uses: docker/setup-buildx-action@v3
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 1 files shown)


yohanboniface:
Thanks a lot @tux2000 !

I've two questions:
- this will build/push a new Docker image only for a new tag, correct ?
- we are about to move to Codeberg (git, issues/PR and CI), do you by chance know is this can be adapted there ? (We'll anyway keep a mirror here, so we could also still rely on Github for the Docker image)

tux2000:
1. yes, this will build and push docker images on "releases" in Github. I think this is even a subset of tags only, but in any case it is based on a clear and manually defined version of the code
2. I dont have experience with codeberg CI so I dont know how to get equivalent functionality in place there, but I would assume that this should be possible

categulario:
@davidbgk @yohanboniface I have experience writing CI that works in codeberg and might be able to asist (and I'll happily contribute to the project in that). Is there a timeline for the migration?

davidbgk:
@categulario the future repository will be at https://codeberg.org/osm-fr/umap and there is no automated sync for now.

The timeline is as soon as we can and all hands are welcome! We might want to NOT import existing issues to start with a clean state (and archiving the Github repo).

PR #3293, opened 2026-04-10
feat: support JOSM syntax for url_template

### LINKED ISSUE #3292: Support JOSM syntax for tile URLs
The syntax used to specify TMS URLs for custom backgrounds and overlays is *almost* but not quite the same as the one used by JOSM, described at https://josm.openstreetmap.de/wiki/Maps#TileMapServicesTMS

Since the JOSM data is the most interesting and comprehensive list of imagery sources, it would be quite convenient if uMap could directly support this syntax, mainly by supporting `{switch:a,b,c}` and `{zoom}` as an alias for `{z}`.

dhdaines:
Fixes: #3292

This allows the use of `{zoom}` and its variants (e.g. `{19-zoom+2}`), `{-y}` for reversed y-coordinates, and `{switch:a,b,c}` for multiple servers.

Basically to allow people to directly use map sources from this excellent list: https://josm.openstreetmap.de/wiki/Maps

(of course, we could do something even more interesting like parsing sources directly from the XML at https://josm.openstreetmap.de/maps)

### DIFF
--- a/umap/models.py
+++ b/umap/models.py
@@ -119,7 +119,7 @@ class Licence(NamedModel):
 
 class TileLayer(NamedModel):
     url_template = models.CharField(
-        max_length=400, help_text=_("URL template using OSM tile format")
+        max_length=400, help_text=_("URL template using OSM or JOSM tile format")
     )
     minZoom = models.IntegerField(default=0)
     maxZoom = models.IntegerField(default=18)
--- a/umap/static/umap/js/modules/rendering/map.js
+++ b/umap/static/umap/js/modules/rendering/map.js
@@ -158,7 +158,10 @@ const ManageTilelayerMixin = {
     if (this._controls) this._controls.tilelayers.setLayers()
   },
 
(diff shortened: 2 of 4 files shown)


yohanboniface:
Hi @dhdaines, thanks for this enhancement suggestion!

Could you describe a bit more the use case you want to cover ? More than just copy-pasting an URL from that list and changing by hand the differences in the syntax.

Just challenging, because more code always means more bugs ;)

dhdaines:
> Hi @dhdaines, thanks for this enhancement suggestion!
> 
> Could you describe a bit more the use case you want to cover ? More than just copy-pasting an URL from that list and changing by hand the differences in the syntax.

Well, in the end it's just that use case :smile: so I understand if this isn't a sufficiently interesting change!  But it served the purpose of helping me learn more about the architecture of uMap.

I will close it for now, I think that a more interesting feature would be to have an interface for adding TMS sources similar to what's in JOSM, but I don't have time to implement that at the moment.

PR #3295, opened 2026-04-10
fix(tests): allow testing without postgres superuser

### LINKED ISSUE #3294: Unable to run tests (need to connect as `postgres` user)
I've installed uMap locally for development, and I'd really like to be able to run the tests before submitting PRs, but I have just no idea how to do this properly.

This is on Debian 13 - PostgreSQL is using peer authentication by default, so I have created the `umap` database with my login user as the owner.  Everything works fine with the `umap` CLI, but I can't run any tests at all.  `make testpy` output is unintelligible but if I run `pytest -s -vv -n 0 umap/tests/test_tilelayer.py` for example I get this output:

```
No UMAP_SETTINGS defined in env var. Trying in /etc/umap/umap.conf
File /etc/umap/umap.conf not found, now trying /home/dhd/work/umap/umap/settings/local.py
Loaded local config from /home/dhd/work/umap/umap/settings/local.py
============================= test session starts ==============================
platform linux -- Python 3.13.5, pytest-9.0.2, pluggy-1.6.0 -- /home/dhd/work/umap/.venv/bin/python3
cachedir: .pytest_cache
django: version: 5.2.12, settings: umap.tests.settings (from ini)
rootdir: /home/dhd/work/umap
configfile: pytest.ini
plugins: playwright-0.7.2, rerunfailures-16.1, anyio-4.13.0, Faker-40.13.0, base-url-2.1.0, django-4.12.0, xdist-3.8.0
collecting ... collected 1 item

umap/tests/test_tilelayer.py::test_tilelayer_json Creating test database for alias 'default' ('test_umap')...
ERROR

==================================== ERRORS ====================================
____________________ ERROR at setup of test_tilelayer_json _____________________

self = <DatabaseWrapper vendor='postgresql' alias='__no_db__'>

    @async_unsafe
    def ensure_connection(self):
        """Guarantee that a connection to the database is established."""
        if self.connection is None:
            if self.in_atomic_block and self.closed_in_transaction:
                raise ProgrammingError(
                    "Cannot open a new connection in an atomic block."
                )
            with self.wrap_database_errors:
>               self.connect()

```

And then PAGES of stack trace, but basically what it says is that it's trying to connect as the `postgres` super-user in order to create new databases for testing, and obviously, it can't do this, because I'm running as my login user.

But where is it even trying to do this?  In a secret Django fixture somewhere?

And how do you run the tests, since you obviously have to connect as `postgres` somehow?  Do you define a password for `postgres`?  If so, where do you configure this password for testing?

Do I need to use Docker for development?  (please, no...)

dhdaines:
Fixes: #3294

This makes the default behaviour to reuse the `test_umap` database, ~and to use it for all subprocesses with `pytest-xdist`~ (unfotunately that doesn't work).  The contributing documentation has also been updated to describe how to set up PostgreSQL on a Debian (or other GNU/Linux distribution using peer authentication) to allow tests to run.

Also the pytest configuration is in `pyproject.toml` where it's easy to find :smile:

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -71,7 +71,9 @@ sudo -u postgres psql test_umap -c "CREATE EXTENSION postgis"
 
 You will also need to run the tests with `pytest -n 0` to prevent
 `pytest` from running multiple processes, each of which requires its
-own database.
+own database.  Alternately, you can manually create a database for
+each process as shown above - the databases will be named
+`test_umap_gw0` through `test_umap_gw3`.
 
 The Django configuration for testing is **not** in
 `umap/settings/local.py` but in `umap/tests/settings.py`.  If you need
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -52,11 +52,35 @@ This will run JavaScript and Python unittests + Playwright integration tests
 
 #### Python unit tests
 
+Python unit tests require you to have a PostgreSQL server with either:
+
+1. Super-user access (the `postgres` user) with password
+   authentication, as you might have in a Docker container.  This is
+   how the GitHub CI tests run.
+2. A `test_umap` database already created with PostGIS enabled.
+
+If you are running PostgreSQL on a Debian installation with peer
+authentication (meaning, your PostgreSQL username is the same as your
+Unix username), you will need to create the test database before
+running tests, like this:
+
+```bash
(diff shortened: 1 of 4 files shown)


davidbgk (approved):
This is great!

PR #3297, opened 2026-04-11
docs: minimally document the administration site

dhdaines:
Fixes: #3287

This adds a page and a link in "Settings" to remind people like me who haven't used Django in a while that the excellent administration site is a thing that exists.

But more importantly it's how you add tile layers (I didn't mention using it to add pictograms since it seems that's deprecated).


docs/config/admin.md
```diff
@@ -0,0 +1,61 @@
+# Administration
+
+uMap uses the built-in [Django administration
+site](https://docs.djangoproject.com/en/6.0/ref/contrib/admin/) to
```
davidbgk: I just discovered that https://docs.djangoproject.com/en/stable/ref/contrib/admin/ resolve to the latest version.

Which raises another concern: we're actually still in 5.2 😅 

I'm not sure how to deal with that dynamically with mkdocs (or if it's even possible easily) and I think it's better to link to the latest stable rather than an outdated version anyway but if you have a better idea let's discuss!

docs/config/settings.md
```diff
@@ -3,7 +3,10 @@
 uMap runs with Django, so any Django setting should work, if you know what you
 are doing.
 
-The Django settings reference is here: https://docs.djangoproject.com/en/4.2/ref/settings/
+The Django settings reference is here: https://docs.djangoproject.com/en/6.0/ref/settings/
```
davidbgk: Same here

davidbgk (approved):
This is really good, thank you!

docs/config/admin.md
dhdaines: Oh, hmm, I was looking for that link to the latest stable version.  Ideally yes the links in the dev documentation should reflect the pinned version in the dependencies, but even though that *could* be automated, it's probably not worth the trouble - so the next best option is to use the permalink to the stable version.

docs/config/settings.md
dhdaines: ah, good catch...

PR #3309, opened 2026-04-20
fix: read OPENROUTESERVICE_HOST from env

rapenne-s:
I haven't been able to test it at the moment, but the change seemed trivial?


yohanboniface:
:+1:

rapenne-s:
there was a typo with `defaul` without a t

yohanboniface:
Oops! Can you push a fix on that same branch ? I'll merge it right away :)

(Note for my self: do not merge PRs while on a visio :p )

rapenne-s:
I'm working on it, and I'm now able to test changes using a container, it turns out OPENROUTESERVICE_HOST is an URL, so I propose that I rename it OPENROUTESERVICE_URL otherwise it's not obvious, while there I could document the variable.

yohanboniface:
The thing is that the parameter is called "host" on ORS side, so I'm tempted to follow their naming. What do you think?

rapenne-s:
> The thing is that the parameter is called "host" on ORS side, so I'm tempted to follow their naming. What do you think?

I don't really have a strong opinion about this, once the variable has documentation for deployment it will be obvious to figure that an URL is expected :+1: no need to rename everything then.

PR #3332, opened 2026-05-12
fix: never force panel mode in portrait/mobile screens

### LINKED ISSUE #3320: (Mobile) Data Browser ignores selected visualization
Hello,

First of all, here's the link to the map, as requested: https://umap.openstreetmap.fr/en/map/il-maggio-dei-libri-2026-a-patti_1395550#12/38.111286/14.998169

The problem (I don't remember if I've already reported it) is important mainly for mobile users.

If you save the map with the shrinked "data browser" view, as expected (see pic 1), this mode is ignored and the view will be anyway the expanded one (see pic 2), and there is no way to change it (or at least, I don't know if one way exists). In other words, you can save the shrinked view, but there's no way to keep it.

Image
Image

yohanboniface:
fix https://github.com/umap-project/umap/issues/3320

### DIFF
--- a/umap/static/umap/js/modules/ui/panel.js
+++ b/umap/static/umap/js/modules/ui/panel.js
@@ -20,6 +20,11 @@ export class Panel {
   }
 
   setDefaultMode(mode) {
+    // In portrait mode (usually mobile), never force the mode,
+    // and let the default CSS apply, and the user decide to force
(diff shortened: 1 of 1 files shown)


picomiles:
Tank you so much; when it will be available in production?

PR #3342, opened 2026-05-21
feat: add snap to polylines

### LINKED ISSUE #2159: Snap Interaction
Please make snapping/sticking feature, e.g. when we draw new line/point, it can be attached exactly to an existing one, just like in any CAD system or OSM data editor. Umap is very lacks this.

One of these:
1) While drawing a line/point, when the mouse is near an existing one, the latter is highlighted. When we click during this highlight, the new point is creating at the nearest position from the mouse exactly on top of the highlighted one (CADs usually use this approach). 
2) While drawing a line/point, when the mouse is near an existing one, the point instantly sticks to it and we can adjust the position and click the mouse to create (see example below).

Also the snapping should be when a line ends on itself.

None. I import some OSM data and want to draw manually on top of that, connecting points/lines, but the precise mapping with this soft is impossible since new geometry's points are created at some distance from the existing vector's axis/center and this distors the final map.

**Additional context**
You can see an example here: https://openlayers.org/en/latest/examples/snap.html
(I understand, that you use Leaflet and this may be its problem, but decided to put the issue here for more people to know about it).

int-72h:
This PR adds snap functionality when drawing and editing maps. It closes #2159. 
I've used Claude to write most of it, as a result I'm marking the PR as a draft until I'm confident it hasn't pumped out complete slop. I'll let it explain what it's done:

## Snap to existing features while editing

Adds snap-to-feature when drawing and editing maps (à la Google My Maps), using the `makinacorpus/Leaflet.Snap` library adapted to uMap's Leaflet.Editable stack.

**Behaviour**
- Drawing a line/polygon/marker: the cursor and placed vertices snap onto nearby features, with a visual indicator (incl. the first vertex).
- Dragging an existing marker/vertex: snaps via Leaflet.Snap's `MarkerSnap`.
- New `snapDistance` setting (default 15px, `0` disables; per-map and `?snapDistance=` override).

**Changes**
- Vendored `leaflet-snap` + `leaflet-geometryutil` (`package.json`, `vendorsjs.sh`, `js.html`).
- Snap adapter in `umap.controls.js`; `snapDistance` in `schema.js`/`help.js`.
- Fix: `setPropertiesFromQueryString` `asNumber` wrote to an undefined var and lacked a presence guard.
- Integration tests in `test_snap.py`.

**Notes**
- Snapping is **on by default** (15px) — flip the schema default to `0` for opt-in.
- Not run against full CI / real browser; unit + targeted integration tests pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


int-72h:
It all seems to be reasonably sane; though one may want to trim back the comments?
The biggest code hog is the wrapper required to get leaflet-snap working with Editables - the rest seems to be reasonable. Somebody more familiar with the codebase should look at it though.

Issue #3298, opened 2026-04-11
download definition ends by "<br data-mce-bogus="1">"


osmvermesser:
i have long time a map - look: <br data-mce-bogus="1">

when open i get following message:

_Laden der externen Daten fehlgeschlagen für Ebene "Hauptwegweiser OHNE Direction" mit der URL "https://overpass-api.de/api/interpreter?data=[out:json][timeout:100];node["information"="guidepost"]["bicycle"="yes"][!"direction_northeast"][!"direction_northwest"][!"direction_southeast"][!"direction_southwest"][!"direction_north"][!"direction_east"][!"direction_south"][!"direction_west"](54.12296659449703,9.196071624755861,54.17660287399418,9.360866546630861);out geom;"<br data-mce-bogus="1">_

the last statement is: <br data-mce-bogus="1">

i did not change any definition!

is "<br data-mce-bogus="1">"  correct or a mistake?

regards Jan

PR #3371, opened 2026-06-04
Remove erroneous HTML tags from German translations

mstock:
These HTML tags only seem to appear in the German translations. As HTML tags in translations get escaped since 2828cab3c8, they started to show up in the user interface when using the German version.

This apparently seems to be a somewhat recurring effect, the commit which added the escaping actually removed such tags, but they were then re-introduced in 12f23bc8d6. The first time they appear seems to be 190cac3105, apparently the first time these strings got translated.

Note: I'm not sure if this is the correct way to fix this kind of error in translations and I haven't tested this since I don't have a running uMap instance or development environment, so feel free to disregard this PR if there's a better/correct approach and just take this as a hint that there is an issue somewhere (in the tool chain or maybe in Transifex?) with the German translations.

While searching for related issues, I also came across #3298 - the error in the translations most likely didn't cause that problem, but it might have been hiding the underlying issue a bit.


yohanboniface:
Thanks!
But the thing is that I already tried to remove those (see https://github.com/mstock/umap/commit/2828cab3c894ef1b3e014f8a95818496d35a87af#diff-f73b8d3069ad405ea2b4077b9089e4e439b5d7cd68cdc47a53a5b61398177ac4), so I guess we should do the cleaning on transifex directly. Could you do that ?
Otherwise I'm afraid they'll come back again at next sync.

mstock:
Thanks for your feedback. I was actually hoping to get away without the need for a Transifex account ;). But I've created one now and seem to have sent a join request to the uMap project for German. Once this gets granted, I can take a look and check if I see something in the strings that I can actually fix.

Btw. - the Transifex link in the [contribution guidelines](https://github.com/umap-project/umap/blob/master/docs/contributing.md) should probably point to `https://explore.transifex.com/openstreetmap/umap/`, I'm just getting a 404 for the current link right now, but that might be due to me not being a member of the organization, yet.

mstock:
Should also be fixed in Transifex now. There actually was some sort of 'button' in their UI (which showed the HTML in a popup when clicking on it) that could just be deleted. I initially feared that this might not even be visible there. So feel free to just close this PR as this should get fixed during the next translation update if you prefer that.

yohanboniface:
Thanks! :+1:

PR #3403, opened 2026-07-09
chore: set global.JSDOM before actually running the js tests

yohanboniface:
Setting it in the unittests files itself could not work because schema.js will call translate() at load time, and in JS modules imports are always called before the module body (so the order of the import vs the global.JSDOM affectation does not matter).


davidbgk (approved):
Conftest FTW! 😉

PR #3432, opened 2026-07-16
fix: align popup content for RTL languages

### LINKED ISSUE #3356: Popup content direction remains LTR when language is set to RTL language (Hebrew, Arabic)
**How to report a bug**

We, developers, need to REPRODUCE a bug to be able to fix it. So if you want the bug to be fixed, help us REPRODUCE it.

We need:
- A link to a map: https://umap.openstreetmap.fr/he/map/anonymous-edit/1414424:GSpRWet_rRsGQdFtqBg1Vk00Le_NcQUQEnCXVFPCzDQ

- Click on the pin in the middle of the map. See the text is aligned to the left of the popup instead of the right.
- 
- I'm not a developer, I found that in pull https://github.com/umap-project/umap/pull/2078 umap significantly improved rtl support, but this part was probably missed?

Hopefully this info will be enough to fix it.

thanks

Image

ergs0204:
Fixes #3356

Leaflet sets popup wrappers to `text-align: left`, which overrides the direction inherited from uMap's RTL document. Override that physical alignment with the logical `start` value so popup content aligns left in LTR languages and right in RTL languages.

Add a Playwright regression test that opens a Hebrew map, verifies the document is RTL, and checks that the popup uses logical start alignment.

## Tests

- `DATABASE_URL=postgis://postgres:postgres@localhost:55432/postgres uv run pytest -n 0 -vv umap/tests/integration/test_popup.py umap/tests/integration/test_view_marker.py umap/tests/integration/test_view_polygon.py umap/tests/integration/test_view_polyline.py --reruns 1 --maxfail 3` (16 passed)
- `DATABASE_URL=postgis://postgres:postgres@localhost:55432/postgres uv run pytest -n 0 -vv umap/tests/integration/test_websocket_sync.py --reruns 1 --maxfail 3` (17 passed)
- `TZ=UTC make testjs` (161 passed)
- `uv run ruff format --check --target-version=py310 umap/tests/integration/test_popup.py`
- `uv run ruff check umap/tests/integration/test_popup.py`
- `npx eslint umap/static/umap/js/`

### DIFF
--- a/umap/static/umap/css/popup.css
+++ b/umap/static/umap/css/popup.css
@@ -65,6 +65,7 @@ span.popup-icon {
 
 .leaflet-popup-content-wrapper {
     border-radius: 4px;
+    text-align: start;
 }
 
 .umap-popup-content {
--- a/umap/tests/integration/test_popup.py
+++ b/umap/tests/integration/test_popup.py
@@ -73,3 +73,26 @@ def test_table_popup(live_server, map, page):
     expect(page.get_by_role("cell", name="with bold").locator("strong")).to_have_text(
         "bold"
     )
+
+
+def test_popup_alignment_in_rtl_language(live_server, map, page):
+    data = {
+        "type": "FeatureCollection",
+        "features": [
+            {
+                "type": "Feature",
+                "geometry": {"type": "Point", "coordinates": [2.49, 48.79]},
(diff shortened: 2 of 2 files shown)


davidbgk (approved):
Thank you!

PR #3452, opened 2026-07-26
fix: export absolute iconUrl paths in .umap files for pictograms

### LINKED ISSUE #3442: Export .umap files with absolute URLs for Osmic SVG icons instead of relative paths
**Description**
When exporting a map with Osmic SVG icons to the .umap format, the icon URLs are saved as relative paths. This causes rendering issues when the exported file is used outside of the original uMap instance (for example, as an overlay in Myosmatic).

**Steps to Reproduce**
1. Add a marker to a uMap layer and use an Osmic SVG icon (e.g., Beach).
2. Export the map to the .umap format.
3. Open the exported file and inspect the source code.

**Current Behavior**
The icon path is saved as a relative URL:

"iconUrl": "/uploads/pictogram/beach.svg"

This path works correctly only when re-importing the file back into the same uMap instance.

**Expected Behavior / Proposed Solution**
The icon paths should use absolute URLs pointing to the official repository or asset source (e.g., GitHub or OSM wiki), for example:

"iconUrl": "https://wiki.openstreetmap.org/w/images/c/cd/Beach_resort-14.svg"

or https://wiki.openstreetmap.org/wiki/Category:OSM_Carto_icons
...on Github:  https://github.com/gmgeo/osmic/ 

**Context / Impact**
If a user wants to use the exported .umap file as an overlay in third-party tools like Myosmatic, the icons fail to render because the relative path becomes invalid.

Manually editing the paths in the .umap file works as a temporary workaround, but it is unusable for datasets with many markers. Changing this to an absolute URL would greatly improve interoperability and simplify rendering across different platforms. Will be nice to set custom icons dataset URL to render them in Myosmatic correctly. 
Thanks.

MatrixNeoKozak:
## What
When exporting a map to .umap format, icon URLs for pictorial markers (e.g. Osmic SVG icons) were serialized as relative paths (e.g. `/uploads/pictogram/beach.svg`), which break when the file is used outside the original uMap instance.

This fix converts relative `iconUrl` paths to absolute URLs during export using `request.build_absolute_uri()`, ensuring icons render correctly in third-party tools like Myosmatic without manual path editing.

The change is in `Map.generate_umapjson()` in `models.py` — a new `_make_icon_urls_absolute` static method iterates over every feature in the exported GeoJSON and converts any `iconUrl` starting with `/` to a fully qualified URL.

## Why
This change resolves the target issue or improvement.

## How to test
Verify that the project builds/runs correctly and the specific bug/improvement is addressed.

Fixes #3442

### DIFF
--- a/umap/models.py
+++ b/umap/models.py
@@ -321,6 +321,7 @@ class Map(NamedModel):
             if include_data:
                 with datalayer.geojson.open("rb") as f:
                     layer = json.loads(f.read())
+                self._make_icon_urls_absolute(layer, request)
             if datalayer.settings:
                 datalayer.settings.pop("id", None)
                 datalayer.settings.pop("parent", None)
@@ -330,6 +331,18 @@ class Map(NamedModel):
             datalayers.append(layer)
         umapjson["layers"] = layers_tree(datalayers, keep_ids=False)
         return umapjson
+
+    @staticmethod
+    def _make_icon_urls_absolute(layer, request):
+        if not isinstance(layer, dict):
+            return
+        for feature in layer.get("features") or []:
+            props = feature.get("properties")
+            if not isinstance(props, dict):
+                continue
+            icon_url = props.get("iconUrl")
+            if icon_url and icon_url.startswith("/"):
+                props["iconUrl"] = request.build_absolute_uri(icon_url)
 
     def get_absolute_url(self):
         return reverse("map", kwargs={"slug": self.slug or "map", "map_id": self.pk})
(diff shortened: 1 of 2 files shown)


yohanboniface:
Thanks!
Would you mind adding some test covering that function ?
Also, this will break when exporting/reimporting in the same instance: the automatic switch from black to white depending on the shape colour relies on the URL being local or not. We could make this check also against the static_url domain. But given we are in the middle of a huge migration to OpenLayers, we'll tackle this part ourselves, don't bother about it.

MatrixNeoKozak:
Added test covering `_make_icon_urls_absolute` in `test_map.py`. Let me know if you'd like additional edge cases covered.

yohanboniface:
Why did you close ? :thinking:

MatrixNeoKozak:
why not)

PR #3463, opened 2026-08-12
fix: do not crash when tilelayer url is invalid

yohanboniface:
While trying to set the preconnect tag.

(A lot of ruff noise in the middle, dunno why…)

(Also, trying to fix the flacky tests, because grrrrrr)

### DIFF
--- a/umap/views.py
+++ b/umap/views.py
@@ -132,7 +132,7 @@ class PaginatorMixin:
         return response
 
 
-class PublicMapsMixin(object):
+class PublicMapsMixin:
     def get_public_maps(self):
         qs = Map.public
         if (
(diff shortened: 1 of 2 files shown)


umap/views.py
```diff
@@ -651,7 +650,12 @@ def set_preconnect(self, properties, context):
             if tilelayers:
                 url_template = tilelayers[0].get("url_template")
         if url_template:
-            domain = urlparse(url_template).netloc
+            # url_template is user-controlled and may be malformed (e.g.
```
yohanboniface, line 653: Real fix is here.

davidbgk (approved):

PR #3496, opened 2026-09-20
fix: hide deleted and blocked maps from team pages

### LINKED ISSUE #3494: Error 403 when trying yo delete a map associated with a team
This bug has been [reported in the Framasoft forum](https://framacolibri.org/t/framacarte-erreur-403/32847) (with captures) but I can reproduce on umap.

1. create a map with your account
2. create a team
3. change permissions for the map: associate the map with the team and save
4. delete the map
5. go to your team: the map is here but unavailable and a 403 popup shows up

bensynapse:
I run Live Tennis API.

Deleted team maps remain in the gallery for team members, where their previews return 403. Exclude deleted and blocked maps before pagination, while keeping public, unlisted, private and draft maps visible to members.

Closes #3494.

Adds regressions for both excluded states and the delete-to-gallery flow; all three fail on unchanged master. Extends the existing visibility test to cover all four accessible states.

Validated in Docker with Python 3.14 and PostGIS 14:
- `make test-unit`: 293 passed; final team suite: 19 passed.
- `make testjs`: 169 passed; four existing team browser tests passed in Chromium.
- `make lint` and `make build`: passed.


yohanboniface:
Thanks!

PR #3497, opened 2026-09-20
Add CORS headers to MapDownload

### LINKED ISSUE #3495: Add CORS headers to MapDownload
The lack of an API with CORS is what keeps me using Google MyMaps as a data backend for [Ultra](https://overpass-ultra.us) on occasion. I'd MUUCCH rather use uMap than GMM. But, GMM allows me to fetch a map as KML with CORS headers, allowing me to use all the MapLibre rendering options exposed by Ultra.

Add CORS headers to [MapDownload](https://github.com/umap-project/umap/blob/master/umap/views.py#L867C7-L867C12)

I've searched the issues and while #175 and #1910 are related, `MapDownload` unlike `DataLayerView` already is a public API as it is a data contract honored by uMap's import functionality. As such, it can provide the quick win that #1910 had hoped for, while not exposing an internal API as public and without the larger effort that #175 would require.

**Additional context**
As I'm the developer of Ultra, I will add a `umap` provider to Ultra in a heartbeat if CORS headers are added.

dschep:
As described in #3495, I'd love it if I could use uMap as a backend for ultra instead of Google My Maps. This adds those necessary CORS headers. 

### Demo

I've deployed a private(I think? There's no way for folks to make an account if I've only setup django local user auth, right?) umap instance with this patch applied and pushed a new query provider to the dev branch of Ultra. I've used it to recreate my simple emoji & MapTiler Basic map that I made when planning my trip for SOTM / family vacation.

**Google My Maps powered:** https://overpass-ultra.us/#query=gist:a22e3cea5bf3da601d441b8f39b775cf&run

**uMap powered:** https://dev.overpass-ultra.us/#query=gist:2d24f7ee6475cd16fd19e9b175a1f319&run

closes #3495


yohanboniface:
Thanks!

